PR #59, opened 2015-01-06
Implement max_executors_per_slave job conf setting

josephharrington:
We currently allow users to specify a "max_executors" setting on their
jobs. This is useful to prevent one job from "overparallelizing" and
monopolizing an entire cluster.

This change implements an additional job config setting called
"max_executors_per_slave". This is useful, for example, on jobs which
*can* be parallelized across multiple machines, but *cannot* be
parallelized with multiple processes on a single machine. (That problem
can now be solved by setting max_executors_per_slave to 1.)

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -43,14 +43,16 @@ class Build(object):
         self._slaves_allocated = []
         self._num_executors_allocated = 0
         self._num_executors_in_use = 0
+
         self._max_executors = float('inf')
-        self._build_completion_lock = Lock()
+        self._max_executors_per_slave = float('inf')
 
         self._all_subjobs_by_id = {}
         self._unstarted_subjobs = None
         self._finished_subjobs = None
         self._postbuild_tasks_are_finished = False
         self._teardowns_finished = False
+        self._timing_file_path = None
 
     def api_representation(self):
         return {
@@ -69,7 +71,7 @@ class Build(object):
         """
         :type subjobs: list[Subjob]
         :type project_type: project_type.project_type.ProjectType
-        :type job_config: master.job_config.JobConfig
+        :type job_config: JobConfig
         """
         if not self._preparation_coin.spend():
             raise RuntimeError('prepare() was called more than once on build {}.'.format(self._build_id))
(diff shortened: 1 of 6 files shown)


boxcla:

app/master/build.py
```diff
@@ -43,14 +43,16 @@ def __init__(self, build_request):
         self._slaves_allocated = []
         self._num_executors_allocated = 0
         self._num_executors_in_use = 0
+
         self._max_executors = float('inf')
-        self._build_completion_lock = Lock()
```
josephharrington: This is already defined on line 42 above.

app/master/build.py
```diff
@@ -43,14 +43,16 @@ def __init__(self, build_request):
...
+        self._max_executors_per_slave = float('inf')
 
         self._all_subjobs_by_id = {}
         self._unstarted_subjobs = None
         self._finished_subjobs = None
         self._postbuild_tasks_are_finished = False
         self._teardowns_finished = False
+        self._timing_file_path = None
```
josephharrington: Random unrelated cleanup -- defining all attributes in constructor.

examples/directory job/clusterrunner.yaml
```diff
@@ -1,4 +1,6 @@
 SampleJob:
+    max_executors: 2
+    max_executors_per_slave: 5
```
josephharrington: Just adding these here for basic reference purposes. This doesn't really do anything. :smile_cat:.

test/unit/master/test_build.py
```diff
@@ -31,27 +33,54 @@ def test_allocate_slave_calls_slave_setup(self):
...
+        expected_current_num_executors_used = 0
+        for i in range(len(mock_slaves)):
+            build.begin_subjob_executions_on_slave(mock_slaves[i])
+            expected_current_num_executors_used += max_executors_per_slave
+            self.assertEqual(
+                build.execute_next_subjob_on_slave.call_count, expected_current_num_executors_used,
+                'After allocating {} slaves, build with max_executors_per_slave set to {} should only be using {} '
+                'executors.'.format(i + 1, max_executors_per_slave, expected_current_num_executors_used))
+
+        self.assertEqual(build.execute_next_subjob_on_slave.call_count, expected_total_num_executors_used,
```
josephharrington: This last assertion is redundant, but I think it makes the test clearer. 

Just realized I need to fix the message for this assertion though. Will do.

drobertduke:
great! :8ball:

PR #60, opened 2015-01-06
Update slave allocation increment value

wjdhollow:
Number of executors for a build was being incremented by
the number of executors on a slave during allocation.

If the max executors per slave, is set to a value
significantly less than the number of executors on the slave,
a build could end up with far fewer allocated slaves then expected.

This commit makes it so that the number of executors on the build
is incrmented by max executors per slave if it is less than the
number of executors available on the slave.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -119,7 +119,8 @@ class Build(object):
         :type slave: Slave
         """
         self._slaves_allocated.append(slave)
-        self._num_executors_allocated += slave.num_executors
+        self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
+
         slave.setup(self.build_id(), project_type_params=self.build_request.build_parameters())
 
     def all_subjobs(self):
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @wjdhollow, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

wjdhollow:
CLA signed

boxcla:
Verified that @wjdhollow has just signed the CLA. Thanks, and we look forward to your contribution.

app/master/build.py
```diff
@@ -119,7 +119,11 @@ def allocate_slave(self, slave):
         :type slave: Slave
         """
         self._slaves_allocated.append(slave)
-        self._num_executors_allocated += slave.num_executors
+        if slave.num_executors < self._max_executors_per_slave:
+            self._num_executors_allocated += slave.num_executors
+        else:
+            self._num_executors_allocated += self._max_executors_per_slave
```
josephharrington: Want to just do 
```python
self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
```

josephharrington:
:+1: nice catch. :smile:

PR #64, opened 2015-01-06
Prevent clusterrunner zombie processes

josephharrington:
This change adds a method to the main thread that will start a timer
of 10 seconds right before the main thread finishes. After 10 seconds,
the application will dump some debug info to the logs and then send
itself a SIGKILL signal.

The dumped debug info includes the stack traces of all currently
running threads.

This is to address an issue where we see the application shutdown the
web server but the process continues to run. This does not affect
deployment of newer versions but just causes us to have some old
zombie clusterrunner processes that we have to kill manually. This
change should prevent those zombies.

### DIFF
--- a/main.py
+++ b/main.py
@@ -2,14 +2,18 @@
 
 import argparse
 import hashlib
+import os
+import signal
 import sys
+import threading
+import time
 
 from app.subcommands.build_subcommand import BuildSubcommand
 from app.subcommands.deploy_subcommand import DeploySubcommand
 from app.subcommands.master_subcommand import MasterSubcommand
 from app.subcommands.slave_subcommand import SlaveSubcommand
 from app.subcommands.stop_subcommand import StopSubcommand
-from app.util import autoversioning, util
+from app.util import app_info, autoversioning, log, util
 from app.util.argument_parsing import ClusterRunnerArgumentParser, ClusterRunnerHelpFormatter
 from app.util.conf.configuration import Configuration
 from app.util.conf.base_config_loader import BaseConfigLoader, BASE_CONFIG_FILE_SECTION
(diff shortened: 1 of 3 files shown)


boxcla:

main.py
```diff
@@ -221,22 +225,49 @@ def _set_secret(config_filename):
     Secret.set(secret)
 
 
+def _start_app_force_kill_countdown(seconds):
+    """
+    Start a daemon thread that will wait the specified number of seconds and then dump debug info and forcefully kill
+    the app. Note that this hard kill will only be executed if the app is unexpectedly hanging. Under normal
+    circumstances, the app will exit cleanly (all nondaemon threads will finish) before the specified delay has elapsed.
```
tjlee0909: Are you saying here that we call an explicit exit() of some sort in python in the subcommand class for the master/slave services?

main.py
```diff
@@ -221,22 +225,49 @@ def _set_secret(config_filename):
...
-        subcommand_class().run(**parsed_args)
+    try:
+        unhandled_exception_handler = UnhandledExceptionHandler.singleton()
+        with unhandled_exception_handler:
+            subcommand_class().run(**parsed_args)
+
+    finally:
+        # The force kill countdown is not an UnhandledExceptionHandler teardown callback because we want it to execute
+        # in all situations (not only when there is an unhandled exception).
+        _start_app_force_kill_countdown(seconds=10)
```
tjlee0909: If the process is not dying, then wouldn't it be the case that the execution wouldn't get here?

tjlee0909: Nm, Joey explained it to me.

wjdhollow:
:+1: lgtm.  should be nice for debugging

tjlee0909:
:+1:

PR #62, opened 2015-01-07
'clusterrunner deploy' times out on deleting build results

tjlee0909:
This command often fails the first time it is run, as one of the actions of the deploy functionality is to delete all of the build results. In the Box CI environment, this can take upwards of twenty minutes.

In order to remedy this, we should rename/move the results directory and asynchronously delete the directory in a process that won't die when the main process dies.

### DIFF
--- a/app/util/fs.py
+++ b/app/util/fs.py
@@ -23,7 +23,7 @@ def async_delete(path):
         _, new_temp_path = tempfile.mkstemp(prefix='async_delete_file')
 
     shutil.move(path, new_temp_path)
-    subprocess.Popen("rm -rf {}".format(new_temp_path), shell=True)
+    subprocess.Popen(['rm', '-rf', new_temp_path])
 
 def create_dir(dir_path, mode=None):
     """
--- a/app/util/fs.py
+++ b/app/util/fs.py
@@ -1,6 +1,29 @@
 import os
+import shutil
 import tarfile
+import tempfile
+import subprocess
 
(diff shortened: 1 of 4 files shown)


boxcla:

app/util/fs.py
```diff
@@ -1,5 +1,26 @@
...
+    Under the covers, this is implemented by first synchronously renaming the file or directory to a unique name
+    inside of a temporary directory. After that, we launch an asynchronous process that deletes that file/directory.
+    This asynchronous process does not die when the invoking process (this process) exits/dies.
+
+    :param path: the absolute path to the file or directory to asynchronously delete
+    :type path: str
+    """
+    new_temp_path = tempfile.mkdtemp(prefix='async_delete_directory')
+    shutil.move(path, new_temp_path)
+    subprocess.Popen(['rm', '-rf', new_temp_path])
```
josephharrington: Should we add `shutil.move` to the blacklisted_methods list in `BaseUnitTestCase`? In other words... the desired behavior would be that tests raise an exception if they don't explicitly patch out `shutil.move`. I was wondering if this is true or if we need to add that method to the blacklist.

Let me know if that doesn't make sense. :wink:

tjlee0909: I forgot we had a blacklist. shutil.move seems like an obvious candidate for something we might forget to mock out, and definitely should.

josephharrington:
:+1:

PR #66, opened 2015-01-07
Expose top-level API representation for services

josephharrington:
This just exposes the api_representation() methods that already exist
in the ClusterMaster and ClusterSlave classes. Before this change,
these methods were not used anywhere.


boxcla:

tjlee0909:
LGTM

PR #63, opened 2015-01-07
Executing setup_build before atomizer for request handling on the master?

tjlee0909:
I've noticed that the serial request handler takes a while to atomize some of our builds, and I'm guessing that it's because the setup_build step is run before every atomization, and setup_build can take a long time to run at times.

How does everyone feel about not running setup_build before atomization?

### DIFF
--- a/app/project_type/directory.py
+++ b/app/project_type/directory.py
@@ -35,7 +35,7 @@ class Directory(ProjectType):
         self.project_directory = os.path.abspath(project_directory)
         self._logger.debug('Project directory is {}'.format(project_directory))
 
-    def _setup_build(self):
+    def _fetch_project(self):
         check_command = 'test -d "{}"'.format(self.project_directory)
         output, exit_code = self.execute_command_in_project(check_command, cwd='/')
         if exit_code != 0:
--- a/test/unit/slave/test_cluster_slave.py
+++ b/test/unit/slave/test_cluster_slave.py
@@ -148,9 +148,9 @@ class TestClusterSlave(BaseUnitTestCase):
         self.mock_network.post().status_code = http.client.OK
         slave = self._create_cluster_slave()
         project_type_mock = self.patch('app.slave.cluster_slave.util.create_project_type').return_value
-        # This test uses setup_complete_event to detect when the async setup_build() has executed.
+        # This test uses setup_complete_event to detect when the async fetch_project() has executed.
         setup_complete_event = Event()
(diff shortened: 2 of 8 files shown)


boxcla:

wjdhollow:
:+1: lgtm

app/project_type/directory.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, project_directory, config=None, job_name=None, build_project_
         self.project_directory = os.path.abspath(project_directory)
         self._logger.debug('Project directory is {}'.format(project_directory))
 
-    def _setup_build(self):
+    def _fetch_project(self):
         check_command = 'test -d "{}"'.format(self.project_directory)
         output, exit_code = self.execute_command_in_project(check_command, cwd='/')
         if exit_code != 0:
```
josephharrington: I agree that we should change the name of this method to something besides setup_build. I'm wondering about using the verb "fetch" -- it is a bit counterintuitive that "fetch_project" for directory type just tests if the directory is there. Maybe something like "prepare_project" would be more apt?

I'm not too concerned about this actually. "fetch_project" is still better and directory project types aren't really commonly used anyway. :)

test/unit/slave/test_cluster_slave.py
```diff
@@ -182,7 +182,7 @@ def test_setup_should_send_correct_state_update_to_master(self, is_setup_success
         slave.connect_to_master(self._FAKE_MASTER_URL)
         slave._project_type = MagicMock()
         if not is_setup_successful:
-            slave._project_type.setup_build.side_effect = SetupFailureError
+            slave._project_type.fetch_project.side_effect = SetupFailureError
 
         slave._async_setup_build(executors=[], project_type_params={})
```
josephharrington: Can you add an assert for the `run_job_config_setup()` call? I'm not sure if this is the test that it would go in but it seems like it should go somewhere. :)

app/project_type/directory.py
tjlee0909: We had originally named it "setup_build" because it was the do-everything method, and it made sense for it to be broad. I knew for sure that we wanted to rename it from setup_build() now that the function doesn't do everything anymore, but I'm pretty open to renaming it; though I do like "fetch_project".

josephharrington: Okay, cool. Yeah I'm fine with fetch_project if you prefer that.

test/unit/slave/test_cluster_slave.py
tjlee0909: Done

PR #67, opened 2015-01-08
In slave allocation loop, do not take down master if slave goes down

tjlee0909:
Currently if the slave service goes down, and the master tries to allocate a subjob to that slave, the whole master service gets taken down.

Instead, catch the network exception in the slave allocation loop, and mark the slave as offline in the master's bookkeeping data structure.

### DIFF
--- a/app/master/cluster_master.py
+++ b/app/master/cluster_master.py
@@ -179,7 +179,7 @@ class ClusterMaster(object):
         """
         # Mark slave dead. We do not remove it from the list of all slaves. We also do not remove it from idle_slaves;
         # that will happen during slave allocation.
-        slave.is_alive = False
+        slave.set_is_alive(False)
         # todo: Fail any currently executing subjobs still executing on this slave.
         self._logger.info('Slave on {} was disconnected. (id: {})', slave.url, slave.id)
 
@@ -337,9 +337,11 @@ class ClusterMaster(object):
 
             while build_waiting_for_slave.needs_more_slaves():
                 claimed_slave = self._idle_slaves.get()
+
                 # Remove dead slaves from the idle queue
-                if not claimed_slave.is_alive:
+                if not claimed_slave.is_alive(use_cached=False):
                     continue
+
                 # The build may have completed while we were waiting for an idle slave, so check one more time.
                 if build_waiting_for_slave.needs_more_slaves():
                     # Potential race condition here!  If the build completes after the if statement is checked,
--- a/app/master/slave.py
+++ b/app/master/slave.py
@@ -23,7 +23,7 @@ class Slave(object):
         self._num_executors_in_use = Counter()
         self._network = Network(min_connection_poolsize=num_executors)
         self.current_build_id = None
-        self.is_alive = True
+        self._is_alive = True
         self._slave_api = UrlBuilder(slave_url, self.API_VERSION)
         self._logger = log.get_logger(__name__)
 
@@ -69,7 +69,7 @@ class Slave(object):
         """
         Tell the slave to run the build teardown
         """
-        if self.is_alive:
+        if self.is_alive():
             teardown_url = self._slave_api.url('build', self.current_build_id, 'teardown')
             self._network.post(teardown_url)
         else:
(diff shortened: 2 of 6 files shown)


boxcla:

app/master/slave.py
```diff
@@ -113,3 +113,32 @@ def free_executor(self):
...
+        try:
+            response = self._network.get(self._slave_api.url())
+            self._is_alive = response.ok
+        except ConnectionError:
+            self._logger.warning('Slave with url {} is offline.', self.url)
+            self._is_alive = False
+
+        return self._is_alive
+
+    def set_is_alive(self, value):
```
tjlee0909: I was considering using the python properties (@property annotation) functionality for the getter and setter, but because the getter in this cause has to have an optional parameter, I decided to go through the old-school explicit route for the getter and setter.

app/master/cluster_master.py
```diff
@@ -179,7 +179,7 @@ def _disconnect_slave(self, slave):
         """
         # Mark slave dead. We do not remove it from the list of all slaves. We also do not remove it from idle_slaves;
         # that will happen during slave allocation.
-        slave.is_alive = False
+        slave.set_is_alive(False)
```
tjlee0909: I am concerned here about the scenario where we want to disconnect the slave, even when the slave is still responsive. Because if that is the case, on line 342 below (in this file), the slave._is_alive property will be set to True the next time the slave.is_alive(use_cached=False) method is invoked.

I feel like that scenario shouldn't happen though (where the _disconnect_slave method gets called, and the slave API is still responsive).

wjdhollow: Crazy thought,  what if we used three states: ALIVE, ZOMBIE, DEAD.

ALIVE -> marked as alive, if not use cached then must be reachable via the API
ZOMBIE -> marked as not alive, but has a responsive API
DEAD -> marked as not alive, and if not use cached, the must not be reachable via the API.

For the time being, ZOMBIE ~= DEAD, but it helps differentiate whether we manually disconnected it (ZOMBIE), or the API is unresponsive (DEAD).

josephharrington: @wjdhollow Will, I see your point about potentially creating a third in-between state, but I feel like it might be better to make the slaves fail fast and just shut down completely when something goes wrong instead of going into zombie limbo. Or did you mean the zombie state would just be a temporary/transitional state?

@tjlee0909 I agree with your last sentence, TJ. I think the behavior we should strive to make the slave conform to is that the slave (from the master's perspective) is either: (1) responsive / alive / completely functional, or (2) unresponsive / dead / not running. 

Right now though, when the slave is shutting down, it will disconnect from the master before its web server goes down, so there **would** be a window where the slave would still have a responsive API but would not expect to receive requests from the master. (The slave is doing build teardown commands during that time, so that window could last an indefinite duration.)

That window is not a risk right now. Specifically, on the master side, your change only actively pings the slave in the slave allocation loop, and the loop only acts on slaves in the `_idle_slaves` queue. The only way a slave gets put back into the idle queue is if it sends a notification to the master that it is idle (via `ClusterMaster.handle_slave_state_update`). Essentially, a slave has to specifically tell the master that it wants to be put back into the idle queue, and that won't happen if the slave has disconnected.

Obvs, that doesn't mean the scenario you mentioned couldn't come up as an issue in the future though. One solution would be to have a specific API endpoint on the slave by which the slave could report its own dead/alive status. (And ideally the slave would always respond with "alive" except during shutdown.) That would decouple the responsiveness of the web server from the internal state of the slave. 

Ugh, sorry long response. :sleeping:

josephharrington:
:+1: on the general approach. Do you plan to deal with the responsive API scenario you mentioned in this PR?

app/master/cluster_master.py
tjlee0909: Ok, this is good info. Let me think about this for a bit. I'll hold off on writing tests and merging anything yet.

app/slave/cluster_slave.py
```diff
@@ -58,16 +59,13 @@ def api_representation(self):
...
-            'connected': self._is_connected(),
+            'is_alive': self.is_alive,
             'master_url': self._master_url,
             'current_build_id': self._current_build_id,
             'slave_id': self._slave_id,
             'executors': executors_representation,
         }
 
-    def _is_connected(self):
-        return self._master_url is not None
```
tjlee0909: I believe this is no longer used anywhere, and has a slightly overlapping functionality with the self.is_alive attribute.

josephharrington:
Update looks good to me.

wjdhollow:
lgtm once the tests are fixed :+1:

Issue #69, opened 2015-01-15
All builds fail when master process has too many open files


josephharrington:
The symptom of the issue is that all builds start failing and will continue to fail until the master service is restarted. 

The stack trace we see in the master log for the failed builds is:
```
[2015-01-12 14:00:51.914] ERROR   RequestHandlerL build           Build 564 failed: filedescriptor out of range in select()
[2015-01-12 14:00:51.917] ERROR   RequestHandlerL cluster_master  Could not handle build request for build 564
Traceback (most recent call last):
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_master.py", line 321, in _build_preparation_loop
    self._request_handler.handle_request(build)
  File "/home/jenkins/ClusterRunnerBuild/app/master/serial_request_handler.py", line 56, in handle_request
    project_type.fetch_project()
  File "/home/jenkins/ClusterRunnerBuild/app/project_type/project_type.py", line 88, in fetch_project
    self._fetch_project()
  File "/home/jenkins/ClusterRunnerBuild/app/project_type/git.py", line 150, in _fetch_project
    self._execute_git_remote_command(fetch_command, self._repo_directory)
  File "/home/jenkins/ClusterRunnerBuild/app/project_type/git.py", line 181, in _execute_git_remote_command
    ['^User.*:', '^Pass.*:', '.*Are you sure you want to continue connecting.*'], timeout=timeout)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/pexpect/__init__.py", line 1451, in expect
    timeout, searchwindowsize)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/pexpect/__init__.py", line 1466, in expect_list
    timeout, searchwindowsize)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/pexpect/__init__.py", line 1535, in expect_loop
    c = self.read_nonblocking(self.maxread, timeout)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/pexpect/__init__.py", line 958, in read_nonblocking
    r, w, e = self.__select([self.child_fd], [], [], timeout)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/pexpect/__init__.py", line 1717, in __select
    return select.select(iwtd, owtd, ewtd, timeout)
[...]

PR #70, opened 2015-01-15
Fix issue with too many open files on master

josephharrington:
This change fixes issue #69.

There are three pieces to this change:
- Separate out the logic that uses pexpect and make it execute in a
  Python subprocess. Pexpect's `child.expect()` raises an exception when
  the process it is launched from has too many open files. (See #69 for
  more details on the issue.)
- Change the version of pexpect in our requirements file to pull a
  a specific commit directly from the pexpect repo. This specific commit
  includes a bugfix for running pexpect in a Python subprocess.
- Add a functional test that tests the git project type so that the new
  multiprocess logic is covered. (Our existing functional tests only
  test the directory type.) This test clones the ClusterRunner demo
  project.

Because of the way the alternate pexpect version is pulled in, we'll
need to run `pip install --upgrade -r requirements.txt` to make sure
that the official 3.3 release is replaced with the snapshot.

### DIFF
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -1,11 +1,13 @@
+from multiprocessing import Manager, Pool  # pylint: disable=no-name-in-module
 import os
 import pexpect
-import signal
+import queue
 import shutil
+import signal
 from urllib.parse import urlparse
 
-from app.util import fs
 from app.project_type.project_type import ProjectType
+from app.util import fs, log
 from app.util.conf.configuration import Configuration
 
 
@@ -17,7 +19,6 @@ class Git(ProjectType):
         "url": "https://github.com/box/StatusWolf.git",
     }
     """
-
     CLONE_DEPTH = 50
     DIRECTORY_PERMISSIONS = 0o700
 
@@ -104,6 +105,7 @@ class Git(ProjectType):
         self._hash = hash
         self._repo_directory = self.get_full_repo_directory(self._url)
         self._timing_file_directory = self.get_timing_file_directory(self._url)
+        self._git_remote_command_executor = _GitRemoteCommandExecutor()
 
         # We explicitly set the repo directory to 700 so we don't inadvertently expose the repo to access by other users
         fs.create_dir(self._repo_directory, self.DIRECTORY_PERMISSIONS)
@@ -140,14 +142,14 @@ class Git(ProjectType):
         repo_exists = git_exit_code == 0
         if not repo_exists:  # This is not a git repo yet, we have to clone the project.
             clone_command = 'git clone {} {}'. format(self._url, self._repo_directory)
-            self._execute_git_remote_command(clone_command)
+            self._git_remote_command_executor.execute(clone_command)
 
         # Must add the --update-head-ok in the scenario that the current branch of the working directory
         # is equal to self._branch, otherwise the git fetch will exit with a non-zero exit code.
         # Must specify the colon in 'branch:branch' so that the branch will be created locally. This is
         # important because it allows the slave hosts to do a git fetch from the master for this branch.
         fetch_command = 'git fetch --update-head-ok {0} {1}:{1}'.format(self._remote, self._branch)
-        self._execute_git_remote_command(fetch_command, self._repo_directory)
+        self._git_remote_command_executor.execute(fetch_command, cwd=self._repo_directory)
 
         commit_hash = self._hash or 'FETCH_HEAD'
         reset_command = 'git reset --hard {}'.format(commit_hash)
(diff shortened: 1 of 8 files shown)


boxcla:

app/project_type/git.py
```diff
@@ -1,11 +1,13 @@
...
 import pexpect
-import signal
+import queue
 import shutil
+import signal
 from urllib.parse import urlparse
 
-from app.util import fs
 from app.project_type.project_type import ProjectType
+from app.util import fs, log
```
wjdhollow: Good catch.  +1 for alphabetized imports

app/project_type/git.py
```diff
@@ -158,7 +160,71 @@ def _fetch_project(self):
     def _execute_in_repo_and_raise_on_failure(self, command, message):
         self._execute_and_raise_on_failure(command, message, self._repo_directory)
 
-    def _execute_git_remote_command(self, command, cwd=None, timeout=10):
+    def execute_command_in_project(self, *args, **kwargs):
+        """
+        Execute a command inside the repo. See superclass for parameter documentation.
+        """
+        # There is a scenario where self.project_directory doesn't exist yet (when a certain repo has never been
+        # fetched before on this particular machine). In order to avoid having python barf during this scenario,
```
wjdhollow: [minor]  "have" in "we have only pass in" seems out of place.

app/project_type/git.py
```diff
@@ -158,7 +160,71 @@ def _fetch_project(self):
...
+        # we have only pass in the cwd if it exists.
```
wjdhollow: This validation of 'cwd' seems a bit out of place to me.  What happens when 'cwd' set to None?  Does it default to setting up the project_directory in the super class?

josephharrington: Oh, I meant to comment -- these two methods (`execute_command_in_project` and `timing_file_path`) have not changed at all. They just moved up in the file. The change here was moving the method `_execute_git_remote_command` into the new class `_GitRemoteCommandExecutor`.

app/project_type/git.py
```diff
@@ -169,24 +235,34 @@ def _execute_git_remote_command(self, command, cwd=None, timeout=10):
         :param timeout: the number of seconds to wait for expected prompts before assuming there will be no prompt
         :type timeout: int
         """
-        child = pexpect.spawn(command, cwd=cwd)
+        # todo: Reenable the functionality around listening for a kill_event to abort execution of this method. This
+        # todo: has been temporarily disabled due to complexity around doing this between multiple processes.
```
wjdhollow: I like how you separated the regex patterns. It improves the readability of the method.

app/project_type/git.py
```diff
@@ -194,20 +270,24 @@ def _execute_git_remote_command(self, command, cwd=None, timeout=10):
...
         # Dump out the output stream from pexpect just in case there was an unexpected prompt that wasn't caught.
-        self._logger.debug("Output from command [{}] after {} seconds: {}".format(command, timeout, child.before))
+        log_msg_queue.put(
+            ('DEBUG', 'Output from command [{}] after {} seconds: {}', (command, timeout, child.before)))
 
         # Now we assume we are past any prompts and wait for the command to end.  We need to keep checking
         # if the kill event has been set in case the build is canceled during setup.
         finished = None
-        while not self._kill_event.is_set() and finished is None:
+        while finished is None:  # and not self._kill_event.is_set()  # todo: kill_event temporarily disabled
```
wjdhollow: Why was kill_event disabled and what are the effects?

josephharrington: The kill event was added here with the feature to enable canceling builds. I disabled it while I figure out a clean way to support this across multiple processes (the existing approach would not work). I'm planning on reenabling it very soon. Disabling this shouldn't have much effect in the short term.

app/subcommands/service_subcommand.py
```diff
@@ -16,6 +17,12 @@ class ServiceSubcommand(Subcommand):
     _THREAD_NAME = None
 
     def run(self, *args, **kwargs):
+        # Set the multiprocessing method to 'spawn'. (The default on unix is 'fork'.) According to documentation, this
```
wjdhollow: Wow.  Good fix.  How are we ensuring that the set_start_method is only called once in the program?  What happens if it is called twice?  Is this testable?

wjdhollow:
:+1:  The overall approach seems sound, but it's hard to validate the correctness from my standpoint without manually testing this change.

I made some comments about some things I was unclear about, but aside from that it looks pretty good.

app/project_type/git.py
wjdhollow: Makes sense.  Can you make a note of that in the discussion and/or the commit message?

app/project_type/git.py
wjdhollow: Never mind, I see now that was not your change.

app/project_type/git.py
josephharrington: Will do.

app/subcommands/service_subcommand.py
josephharrington: This path will definitely only execute once -- this is pretty early on in the application lifecycle. Our `main.py` calls `run()` directly.

I believe this will raise an exception if called more than once, so functional tests would definitely catch it, and also some unit tests would fail if this wasn't mocked out.

tjlee0909:
Looks great. Excellent commenting in there. I vaguely understood the details of the change when we talked verbally, but when I read through the code and comments it was pretty clear to me how it would work.

josephharrington:
Alright, I disabled the functional test that had been failing on Travis due to environment issues. I'll come back to that. That test passes for me in our environment so I'm confident that the Travis failure is not signaling a real issue.

PR #73, opened 2015-01-16
Revert "Fix issue with too many open files on master"

josephharrington:
Reverts box/ClusterRunner#70. This implementation is not compatible with cx_Freeze. :scream:


boxcla:

PR #74, opened 2015-01-20
More detailed versioning scheme

josephharrington:
This change takes our current versioning scheme and adds more
info to the version string for branches and uncommitted changes. The
intention here is to have the version string give us a clear
distinction between trunk, feature branch, and development builds.

The version scheme stays the same for trunk commits (e.g., `0.5.155`).
Trunk commits are defined as being in the first-parent git log of
`origin/master`, which can be viewed with:
   `git log --first-parent --oneline origin/master`

For non-trunk commits, the git short hash is appended to the version
number, for example: `0.5.156-cac0414`.

When there are modified files in the repo that have not yet been
committed, `-mod` is appended to the end of the version string. For
example: `0.5.156-cac0414-mod`.

This scheme maintains the expected version ordering using setuptools'
versioning guidelines. More info on this can be found at
https://pythonhosted.org/setuptools/setuptools.html#specifying-your-project-s-version

### DIFF
--- a/test/framework/base_unit_test_case.py
+++ b/test/framework/base_unit_test_case.py
@@ -19,14 +19,13 @@ class BaseUnitTestCase(TestCase):
     _base_setup_called = False
     # This allows test classes (e.g., TestNetwork) to disable network-related patches for testing the patched code.
     _do_network_mocks = True
-    _repatchable_items = {}
     _fake_hostname = 'fake_hostname'
 
     def setUp(self):
         super().setUp()
         self.addCleanup(patch.stopall)
 
-        self._repatchable_items = {}  # Reset this at the start of each test.
+        self._patched_items = {}
         self._blacklist_methods_not_allowed_in_unit_tests()
 
         # Stub out a few library dependencies that launch subprocesses.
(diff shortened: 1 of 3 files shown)


boxcla:

test/framework/base_unit_test_case.py
```diff
@@ -19,14 +19,13 @@ class BaseUnitTestCase(TestCase):
...
     _do_network_mocks = True
-    _repatchable_items = {}
     _fake_hostname = 'fake_hostname'
 
     def setUp(self):
         super().setUp()
         self.addCleanup(patch.stopall)
 
-        self._repatchable_items = {}  # Reset this at the start of each test.
+        self._patched_items = {}
```
josephharrington: Functionality is now:
 - (unchanged) still only allow a patched item to be repatched if it was patched with `allow_repatch=True`
 - (changed) now anything that has been `patch()`ed can be `unpatch()`ed

This change buys us the ability to patch something for *every* test by patching in `BaseUnitTestCase` and then unpatch it for a specific test. This usually makes more sense than needing to patch the offending item in a bunch of unrelated test classes. (Basically this allows us to whitelist instead of blacklist -- items can be patched by default instead of unpatched by default.)

PR #75, opened 2015-01-20
Add clusterrunner.yaml to repo

josephharrington:
This enables running the ClusterRunner unit tests via ClusterRunner! On
my laptop running a slave with 8 executors cuts down the amount of time
it takes to run the unit tests by about half.

Other changes in this commit to support this:
- Add \_\_init\_\_.py to a couple test directories where the file was
missing (meaning those tests weren't even being run)
- Fix or remove tests in app/util/shell that have been silently broken
because up until now they weren't being run
- Add a test that will ensure an \_\_init\_\_.py file is added to all test
directories in the future
- Add a command to .travis.yml to run unit tests via ClusterRunner.
This works as both a way to run our unit tests and as a real-world
functional test.


boxcla:

.travis.yml
```diff
@@ -7,17 +7,17 @@ env:
...
-# command to run tests
 script:
-  - export DO_PYLINT_CMD="pylint app"
-  - export TEST_UNIT_CMD="nosetests -v test/unit"
-  - export TEST_FUNC_CMD="nosetests -s -v test/functional"
-  - export CR_VERBOSE=1
+  - DO_PYLINT_CMD="pylint app"
+  - TEST_UNIT_CMD="nosetests -v test/unit"
+  - TEST_FUNC_CMD="nosetests -s -v test/functional"
+  - TEST_BUILD_UNIT_CMD="./main.py build"
```
tjlee0909: That's so sweet that this works.

tjlee0909:
:+1:

tjlee0909:
I don't know how TravisCS works, but one concern I might have is about the master/slave daemon continuing to run and persisting across builds. Hopefully those all get killed off, otherwise we're going to have to do it ourselves.

.travis.yml
```diff
@@ -7,17 +7,17 @@ env:
   - DO_LINT=true DO_UNIT=true DO_FUNCTIONAL=false
   - DO_LINT=false DO_UNIT=false DO_FUNCTIONAL=true
```
tjlee0909: Are we going to double-run the unit tests (serially and via clusterrunner) for a while and let it back in?

josephharrington: Yeah that's something I was going back and forth on. I kind of just want to run both forever -- then we always get that extra sanity check that ClusterRunner fails when it should and passes when it should.

josephharrington:
@TJ, yeah I think Travis just uses fresh VMs for each build so I don't think we'd have to worry about cleaning up. :smile: http://docs.travis-ci.com/user/ci-environment/

wjdhollow:
:+1:

PR #78, opened 2015-01-20
Update tornado requirement to 4.0.2

josephharrington:
Tornado 4.0 was released last July, so this should be a pretty safe
update. None of the 3.2.2 to 4.0.2 changes documented in the changelog
should affect us.

### DIFF
--- a/test/unit/web_framework/test_cluster_base_handler.py
+++ b/test/unit/web_framework/test_cluster_base_handler.py
@@ -18,7 +18,7 @@ class TestClusterBaseHandler(BaseUnitTestCase):
     )
     def test_set_default_headers_should_not_set_origin_header(self, request_origin, cors_conf_value):
         mock_application = MagicMock(spec_set=ClusterApplication())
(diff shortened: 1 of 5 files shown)


boxcla:

test/unit/web_framework/test_cluster_base_handler.py
```diff
@@ -37,7 +37,7 @@ def test_set_default_headers_should_not_set_origin_header(self, request_origin,
     )
     def test_set_default_headers_should_set_origin_header(self, request_origin, cors_conf_value):
         mock_application = MagicMock(spec_set=ClusterApplication())
-        mock_request = MagicMock(spec=tornado.httpserver.HTTPRequest, headers={})
+        mock_request = MagicMock(headers={})
```
josephharrington: These had to be changed because of an incompatible change in the implementation of `tornado.web.RequestHandler` that made it not work with this mock in the same way. We don't do much with this mock instance anyway, so removing the spec is not a big deal.

PR #79, opened 2015-01-21
Point stdout and stderr to dev/null for async Popen calls

josephharrington:
This fixes an issue where multiprocessing would raise an exception due
to the stdout stream being closed. The exception would only occur when
ClusterRunner had been deployed via 'clusterrunner deploy'.

This is a cleaner way to launch async subprocesses since we explicitly
will not connect to the spawned process's stdout or stderr streams.


boxcla:

josephharrington:
:+1: yay me.

PR #80, opened 2015-01-26
Fix ambigious revision and filename bug

wjdhollow:
This bug occurs when a git ref has the same name
as a file.  Git will throw an error that says that the
revision and filename is ambiguous.

The way to get around this is to specify the '--' files
option.

### DIFF
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -154,7 +154,9 @@ class Git(ProjectType):
         self._git_remote_command_executor.execute(fetch_command, cwd=self._repo_directory)
 
         commit_hash = self._hash or 'FETCH_HEAD'
-        reset_command = 'git reset --hard {}'.format(commit_hash)
+
+        # The '--' option acts as a delimiter to differentiate values that can be "tree-ish" or a "path"
+        reset_command = 'git reset --hard {} --'.format(commit_hash)
         self._execute_in_repo_and_raise_on_failure(reset_command, 'Could not reset Git repo.')
 
         self._execute_in_repo_and_raise_on_failure('git clean -dfx', 'Could not clean Git repo.')


boxcla:

drobertduke:
:+1:

app/project_type/git.py
```diff
@@ -154,7 +154,7 @@ def _fetch_project(self):
         self._git_remote_command_executor.execute(fetch_command, cwd=self._repo_directory)
 
         commit_hash = self._hash or 'FETCH_HEAD'
-        reset_command = 'git reset --hard {}'.format(commit_hash)
+        reset_command = 'git reset --hard {} --'.format(commit_hash)
```
tjlee0909: In a comment above this line, could you post a link to some documentation or stackoverflow page that indicates what the trailing '--' does? I can't find it on the official git docs.

tjlee0909:
:+1:

app/project_type/git.py
```diff
@@ -154,7 +154,14 @@ def _fetch_project(self):
...
         commit_hash = self._hash or 'FETCH_HEAD'
-        reset_command = 'git reset --hard {}'.format(commit_hash)
+
+        # git reset (--patch | -p) [<tree-ish>] [--] [<paths>...]
+        # For our use case, the '--' option acts as a delimiter to
+        # differentiate values that can be "tree-ish" or a "path"
+        # For example, the case where there is a file that has the same name as a
+        # branch.
+        # The <paths> value is optional and can be safely omitted.
+        reset_command = 'git reset --hard {} --'.format(commit_hash)
```
josephharrington: Looks good. You could maybe abbreviate the comment; I think the "--" is a pretty common syntax for git commands for exactly this reason, so it probably doesn't need a huge explanation.

Should we also add the "--" to our git checkout command? Is there any other git command we do that we could add this to?

PR #81, opened 2015-01-27
Add request params to build api response

josephharrington:
- Including the request params in the build response will make it
easier to collect data about patterns in build behavior.
- Also cleaned up a few old todos in a related file.


boxcla:

wjdhollow:
:+1: Nice change.  This will be useful when tracking down which job it originated from.

drobertduke:
:+1:

PR #82, opened 2015-01-28
Add more logging around build requests

josephharrington:
This change adds more logs in both the human-readable app log and the
(machine-readable) event log. Added logs are all related to collecting
timing data for build requests and preparation/atomization.


boxcla:

wjdhollow:
:+1:

PR #83, opened 2015-01-28
Parallelize build preparation across separate projects

drobertduke:
Currently all build preparation (fetching, atomization) is
serial.  This introduces the concept of distinct projects
which will not conflict with each other during build prep.
Each distinct project gets its own serial queue, but a build
for project-A is not blocked by the prep for a project-B build.


boxcla:

drobertduke:
This is a WIP.  Starting the PR early so you can validate/invalidate the approach.

josephharrington:
Just sent you an email with a potential alternative approach. :love_letter:

app/master/build.py
```diff
@@ -87,7 +101,7 @@ def prepare(self, subjobs, project_type, job_config):
 
         self._max_executors = job_config.max_executors
         self._max_executors_per_slave = job_config.max_executors_per_slave
-        self._timing_file_path = project_type.timing_file_path(job_config.name)
+        self._timing_file_path = self.project_type.timing_file_path(job_config.name)
```
wjdhollow: The project_type could be None at this point.  
Previously, this possibly was avoided by explicitly requiring the project_type be passed in as a parameter.

Should there be additional error handling logic here?

app/master/cluster_master.py
```diff
@@ -228,7 +224,22 @@ def handle_request_for_new_build(self, build_params):
         if build_request.is_valid():
             build = Build(build_request)
             self._all_builds_by_id[build.build_id()] = build
-            self._request_queue.put(build)
+
```
wjdhollow: It appears to me that the project_type can be initialized during construction or initialized lazily since we have all the necessary information when we instantiate it.

app/master/serial_request_handler.py
```diff
@@ -64,10 +48,7 @@ def _compute_subjobs_for_build(self, build_id, job_config, project_type):
         :type project_type: project_type.project_type.ProjectType
         :rtype: list[Subjob]
         """
-        try:
-            atoms_list = job_config.atomizer.atomize_in_project(project_type)
-        except AtomizerError as ex:
-            raise BuildProjectError('Build failed during atomization.') from ex
+        atoms_list = job_config.atomizer.atomize_in_project(project_type)
```
wjdhollow: Why was the try/catch removed?

app/master/cluster_master.py
```diff
@@ -228,7 +224,22 @@ def handle_request_for_new_build(self, build_params):
...
+            build.generate_project_type()
+            project_id = build.project_type.project_id()
+            if project_id in self._request_queues.keys():
+                self._request_queues[project_id].put(build)
+            else:
+                self._request_queues[project_id] = Queue()
+                self._request_queues[project_id].put(build)
```
tjlee0909: I know this is just a WIP, but I believe handle_request_for_new_build can be called concurrently, meaning that there is a chance that you will have multiple build requests for the same project coming in, and consequently we would overwrite the same project_id in the self._request_queue queue.

app/project_type/docker.py
```diff
@@ -151,3 +151,8 @@ def _remove_file_system_unfriendly_characters(self, unescaped_path):
         """
         valid_chars = "-_.()%s%s" % (string.ascii_letters, string.digits)
         return ''.join(c for c in unescaped_path if c in valid_chars)
+
+    def project_id(self):
+        # TODO: Ideally each docker container could have a unique project_id since we can fetch/atomize any docker
+        # containers in parallel without conflicts
+        return self._image
```
tjlee0909: For Docker, I don't think we can fetch multiple images concurrently for the single docker daemon that may be running.

I'm thinking the project_id for Docker in this design should return a static 'docker' or something with a comment indicating the reason for the static-ness.

app/master/build.py
drobertduke: True.  I may need to rethink this.  There are also tests that depend on being able to create a mock project_type and pass it in via this function

app/master/cluster_master.py
drobertduke: Good catch :baseball:

app/master/serial_request_handler.py
drobertduke: I moved the definition of BuildProjectError to Build.  I thought about defining a new type of error to be raised here, or importing BuildProjectError, but then I wasn't sure if it's necessary to transform an AtomizerError to a BuildProjectError.  Maybe I'm missing something though.

app/project_type/docker.py
drobertduke: Gotcha.

PR #84, opened 2015-01-30
Remove managed queue from multiprocessing usage

josephharrington:
We ran into an issue where instantiating the multiprocessing.Manager
was raising a BrokenPipeException. We didn't find a good root cause
but just decided to remove the Manager entirely since we were only
using it as a nice-to-have to make sure our logs were properly
formatted.


boxcla:

tjlee0909:
:+1:

PR #86, opened 2015-01-30
Builds should never hang if all subjobs complete

### LINKED ISSUE #65: Build will hang forever if an allocated slave dies
Right now if a command in setup_build hangs or fails on a single slave, then the entire build will hang even if all subjobs complete on other slaves.

The reason is that we have some code that blocks a build from finishing until no slaves report as working on that build.

josephharrington:
Currently, if a slave fails during setup_build or teardown_build, the
build will be stuck in BUILDING state forever, even if all subjobs
complete successfully on other slaves. (Issue #65)

The reason is that we have code that prevents a build from finishing
until all slaves ever assigned to that build have successfully
completed teardown_build commands. This is actually unnecessary - a
build should be considered complete as soon as its last subjob
completes.

This change removes the code that blocks a build from finishing until
teardown_build is complete. It also clears a slave's `current_build_id`
field when a slave disconnects (since a disconnected slave is not
working on any build.)

Fixes #65.

### DIFF
--- a/app/client/cluster_api_client.py
+++ b/app/client/cluster_api_client.py
@@ -105,6 +105,30 @@ class ClusterSlaveAPIClient(ClusterAPIClient):
     This is a light wrapper client around the ClusterSlave REST API.
     """
     # TODO: Move the API call logic from slave.py into this class.
+    def block_until_idle(self, timeout=None):
+        """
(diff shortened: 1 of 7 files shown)


boxcla:

PR #88, opened 2015-02-02
Add tests for testing num tries for executing git remote commands

tjlee0909:
I had to push the previous PR without tests due to it being an emergency. Following it up with this PR.

### DIFF
--- a/test/unit/project_type/test_git.py
+++ b/test/unit/project_type/test_git.py
@@ -178,3 +178,12 @@ class TestGit(BaseUnitTestCase):
         git_executor._execute_git_remote_command('fake command', cwd=None, timeout=0, log_msg_queue=MagicMock())
 
         self.assertTrue(matched_prompt, "The password prompt was not matched by pexpect")
+
+    def test_execute_raises_broken_pipe_error_on_last_try(self):
+        git_executor = _GitRemoteCommandExecutor()
+        self.patch('app.project_type.git.sleep')
+        self.patch('app.project_type.git.Manager').side_effect = \
+            [BrokenPipeError(''), BrokenPipeError(''), BrokenPipeError('')]
+
+        with self.assertRaises(BrokenPipeError):
+            git_executor.execute('fake command', cwd=None, timeout=0, num_tries=3)


boxcla:

test/unit/project_type/test_git.py
```diff
@@ -178,3 +178,15 @@ def expect_side_effect(patterns, *args, **kwargs):
...
 
         self.assertTrue(matched_prompt, "The password prompt was not matched by pexpect")
+
+    def test_execute_raises_broken_pipe_error_on_last_try(self):
+        git_executor = _GitRemoteCommandExecutor()
+        self.patch('app.project_type.git.sleep')
+        mock_manager = self.patch('app.project_type.git.Manager')
+
+        def manager_init_side_effect(*args):
+            raise BrokenPipeError('')
```
wjdhollow: I feel like there is a more concise way of telling a MagicMock to raise an exception

wjdhollow:
:+1: with small comment

test/unit/project_type/test_git.py
wjdhollow: Does it work if you just do a direct assignment?

mock_manager.side_effect = BrokenPipeError('')

tjlee0909: No =(. It felt a little verbose to me too, and some googling had suggested to me that what you typed should have worked.

josephharrington: This works in console:
```python
from unittest.mock import MagicMock
m = MagicMock()
m.test.side_effect = BrokenPipeError
m.test()
```
```
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python3/3.4.2/Frameworks/Python.framework/Versions/3.4/lib/python3.4/unittest/mock.py", line 896, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/usr/local/Cellar/python3/3.4.2/Frameworks/Python.framework/Versions/3.4/lib/python3.4/unittest/mock.py", line 952, in _mock_call
    raise effect
BrokenPipeError
```

We also do exception side effects in test_project_type.py and test_autoversioning.py:
```python
self.check_output_mock.side_effect = [subprocess.CalledProcessError(1, 'fake')] 
```

PR #89, opened 2015-02-05
Perform ProjectType generation in the Build class

tjlee0909:
...rather than in the serial request handler.

This is going to be one of a multi-PR change that will be rolled out onto ClusterRunner. I figured it'd be easier to rollout if we do this in multiple deployments. This is just a subset of what Duke had in this previous PR that he had up.

I will add tests soon. I just wanted some eyes on it for now (again, it's basically just a subset of Duke's changes).


boxcla:

app/master/cluster_master.py
```diff
@@ -213,6 +213,7 @@ def handle_request_for_new_build(self, build_params):
         if build_request.is_valid():
             build = Build(build_request)
             self._all_builds_by_id[build.build_id()] = build
+            build.generate_project_type()
```
wjdhollow: I mentioned this on Duke's PR as well, but having to call generate_project_type explicitly  opens the risk that project_type is never initialized and is None.  

I see we do some explicit None/ or isinstance checks in the code, but I am wondering if it would be better for consumers if the generate_project_type was called in the constructor.  

I also think separating the instantiation of the build object, and the call to generate_project_type makes the code look separated.  If we choose to keep the call to generate_project_type, can we put the call directly under the object creation?

wjdhollow:
The overall approach looks good to me.  I see that all you did really was move project type generation into the build class.  I made a small comment, but as long as this is tested functionally before pushing :+1:

app/master/cluster_master.py
tjlee0909: I had made this same comment on Joey's PR when he had an initialize() method in addition to his constructor. He pointed out that it's bad practice to do any significant work in the constructor other than variable assignment for a few reasons.

The biggest one is that it's not expected. When you instantiate an object, you don't expect the constructor to throw an exception or to take a lot of computing resources to do--you expect it to always return you an instance quickly.

The other reason was for testability--as it is harder to test the constructor of objects.

wjdhollow: Oh, I see your point, particularly in this case where there are some system level heavy lifting going on, like generating symlinks.  :+1:  on leaving it out

PR #90, opened 2015-02-05
Add unique project_id() method to the project_type classes.

tjlee0909:
Again, essentially another subset of Duke's build. This one is really benign though, and is essentially a no-op.

For every unique project_id(), the builds will have their own request handler.


boxcla:

drobertduke:
:+1:

PR #91, opened 2015-02-06
Pull out functionality of new build request handling into its own class.

tjlee0909:
The ClusterMaster class is starting to bloat a little bit, and before we
add any more complexity to the class, it'd be good to separate the logic
out.

- Created a BuildRequestHandler class
- Created a SlaveAllocator class
- The serial_request_handler class is now a part of the
  BuildRequestHandler class

### DIFF
--- a/app/master/build_request_handler.py
+++ b/app/master/build_request_handler.py
@@ -1,20 +1,91 @@
 import json
 import os
+from queue import Queue
 
 from app.master.atom_grouper import AtomGrouper
 from app.master.build_request import BuildRequest
 from app.master.subjob import Subjob
 from app.master.time_based_atom_grouper import TimeBasedAtomGrouper
-from app.util.log import get_logger
 from app.project_type.project_type import ProjectType
+from app.util import analytics
+from app.util.log import get_logger
+from app.util.safe_thread import SafeThread
 
 
-class SerialRequestHandler(object):
+class BuildRequestHandler(object):
+    """
+    The BuildRequestHandler class is responsible for preparing a non-prepared build.
+
+    Implementation notes:
+
+    This class manages two critical Queue's in ClusterRunner: request_queue and builds_waiting_for_slaves.
+
+    The request_queue is the queue of non-prepared Build instances that the BuildRequestHandler has
+    yet to prepare. This queue is populated by the ClusterMaster instance.
+
+    The builds_waiting_for_slaves queue is the queue of prepared Build instances that the
+    BuildRequestHandler has completed build preparation for, and is waiting for the SlaveAllocator (a separate
+    entity) to pull Builds from.
+
+    All of the input of builds come through self.handle_build_request() calls, and all of the output
+    of builds go through self.next_prepared_build() calls.
+    """
 
     def __init__(self):
         self._logger = get_logger(__name__)
+        self._builds_waiting_for_slaves = Queue()
+        self._request_queue = Queue()
+        self._request_queue_worker_thread = SafeThread(
+            target=self._build_preparation_loop, name='RequestHandlerLoop', daemon=True)
 
-    def handle_request(self, build):
+    def start(self):
+        """
+        Start the infinite loop that will accept unprepared builds and put them through build preparation.
+        """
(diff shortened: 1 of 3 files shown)


boxcla:

app/master/build_preparer.py
```diff
@@ -9,7 +9,7 @@
 from app.project_type.project_type import ProjectType
 
 
-class SerialRequestHandler(object):
+class BuildPreparer(object):
```
tjlee0909: I'm not married to this name. I generally don't like verby nouns as class names.

josephharrington: We could just move all this logic entirely into the `Build` class. It seems like most of the changes to this class since the beginning have been to move stuff out of here into `Build`, so it feels like the direction we've been moving for awhile.

I totally agree with you on the "verby" class names -- I feel like it's kind of a signal that the code is structured more imperatively than object-oriented.

app/master/cluster_master.py
```diff
@@ -25,26 +22,13 @@ class ClusterMaster(object):
...
-        self._request_queue_worker_thread.start()
-
-        self._slave_allocation_worker_thread = SafeThread(
-            target=self._slave_allocation_loop, name='SlaveAllocationLoop', daemon=True)
-        self._slave_allocation_worker_thread.start()
-
         self._master_results_path = Configuration['results_directory']
-
+        self._all_slaves_by_url = {}
+        # This is an OrderedDict so we can more easily implement get_queue()
```
josephharrington: I think you can remove this comment -- we don't even have a `get_queue()` anymore. :cookie:

app/master/build_request_handler.py
```diff
@@ -1,20 +1,84 @@
...
+                    analytics.record_event(analytics.BUILD_PREPARE_FINISH, build_id=build.build_id(),
+                                           log_msg='Build {build_id} successfully prepared and waiting for slaves.')
+                    self._builds_waiting_for_slaves.put(build)
+            except Exception as ex:  # pylint: disable=broad-except
+                build.mark_failed(str(ex))
+                self._logger.exception('Could not handle build request for build {}.'.format(build.build_id()))
+
+    def _prepare_build(self, build):
         """
         Prepare a Build to be distributed across slaves.
```
josephharrington: Nice comments!

josephharrington:
:+1: Noice.

app/master/slave_allocator.py
```diff
@@ -0,0 +1,57 @@
...
+class SlaveAllocator(object):
+    """
+    The SlaveAllocator class is responsible for allocating slaves to prepared builds.
+    """
+
+    def __init__(self, build_request_handler):
+        """
+        :type build_request_handler: BuildRequestHandler
+        """
+        self._logger = get_logger(__name__)
```
wjdhollow: Hmm.  I know we've had the 'slave_allocation' prefix from previous code, but now that the functions are in a SlaveAllocator class, it seems a bit overly verbose to keep it. I think _slave_allocation_worker_thread can just be changed to _worker_thread, and _slave_allocation_loop can just be something like _process_builds or something.

app/master/slave_allocator.py
```diff
@@ -0,0 +1,57 @@
...
+        self._build_request_handler = build_request_handler
+        self._idle_slaves = OrderedSetQueue()
+        self._slave_allocation_worker_thread = SafeThread(
+            target=self._slave_allocation_loop, name='SlaveAllocationLoop', daemon=True)
+        self._slave_allocation_worker_thread.start()
```
josephharrington: Do you think we should start the thread in a separate start() method instead of in the constructor? I know it doesn't make too much difference since we were (previously and currently) doing it in the ClusterMaster constructor anyway.

app/master/slave_allocator.py
tjlee0909: Good point, I'll rename them now that the context is already described by the enclosing class.

app/master/slave_allocator.py
tjlee0909: Yeah, I was thinking that, I'll do that.

PR #92, opened 2015-02-08
Add pep8 and makefile

josephharrington:
This adds the pep8 linter which catches a few things that pylint does
not. Also, to simplify running lint and tests both locally and on
Travis CI, I added a makefile.

This simplifies the act of running the various tests and linters we
have without needing to add a bunch of bash scripts to the repo:
 - `make` to run lint and all tests (probably the most common thing)
 - `make lint` or `make test` to just run either lint or tests
 - `make freeze` to run the cx_Freeze build process

The makefile approach is roughly based on what [HTTPie](https://github.com/jakubroztocil/httpie/blob/master/Makefile) is doing with
their makefile and .travis.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,20 +4,12 @@ python:
 
 # each line defined in env will create a new parallel test job on Travis
 env:
-  - DO_LINT=true DO_UNIT=true DO_FUNCTIONAL=false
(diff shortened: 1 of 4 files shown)


boxcla:

tjlee0909:
:+1:

PR #93, opened 2015-02-11
Parallelize ClusterRunner builds for unique project ids.

tjlee0909:
Now each unique ClusterRunner project has its own preparation lock, meaning that we can have build preparation going on in parallel as long as these builds are for different projects.

Haven't tested it yet, wanted to get eyes on this PR as soon as possible.

### DIFF
--- a/app/master/build_request_handler.py
+++ b/app/master/build_request_handler.py
@@ -1,6 +1,7 @@
 import json
 import os
 from queue import Queue
+from threading import Lock
 
 from app.master.atom_grouper import AtomGrouper
 from app.master.build_request import BuildRequest
@@ -37,6 +38,7 @@ class BuildRequestHandler(object):
         self._request_queue = Queue()
         self._request_queue_worker_thread = SafeThread(
             target=self._build_preparation_loop, name='RequestHandlerLoop', daemon=True)
+        self._project_preparation_locks = {}
 
     def start(self):
         """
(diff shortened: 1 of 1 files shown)


boxcla:

app/master/build_request_handler.py
```diff
@@ -73,6 +75,22 @@ def _build_preparation_loop(self):
         """
         while True:
             build = self._request_queue.get()
+            project_id = build.project_type.project_id()
+
+            if project_id not in self._project_preparation_locks:
+                self._project_preparation_locks[project_id] = Lock()
+
+            project_lock = self._project_preparation_locks[project_id]
```
tjlee0909: Because this method (_build_preparation_loop()) cannot be called concurrently, it's safe to perform operations such as this without worrying about concurrency.

app/master/build_request_handler.py
```diff
@@ -73,6 +75,22 @@ def _build_preparation_loop(self):
...
+            async_preparation_thread = SafeThread(target=self._prepare_build_async, name='BuildPreparationThread',
+                                                  args=[build, project_lock])
+            async_preparation_thread.start()
+
+    def _prepare_build_async(self, build, project_lock):
+        """
+        :type build: Build
+        :type project_lock: Lock
+        """
```
wjdhollow: Can you explain what is happening in the below line?

Is the lock acquired and released via the `with` keyword?  What happens if multiple threads call `acquire()`?  Do they acquire it in the order that they called it, or do we even care?

tjlee0909: That's correct, the 'with project_lock:' is a short-hand for calling a project_lock.acquire() at the beginning and calling project_lock.release() at the end when the block exits.

When multiple threads call acquire, then I don't think there is any contractual guarantee on the order in which the threads receive the locks, no. It is an issue that we were aware of when designing it, but we decided that it was minor enough to not matter.

wjdhollow: Nice.  The approach looks good to me.

drobertduke:
so simple :+1:

josephharrington:
:+1:

PR #94, opened 2015-02-14
Cache N event logs to improve API performance

wjdhollow:
Cache N event logs to improve the API performance.  Currently, we load from file every time, and that file can grow very large.  Now the default is to serve events from the cache, and to only load from disk if the timestamps are outside of the since_timestamp or since_id range.

### DIFF
--- a/app/util/analytics.py
+++ b/app/util/analytics.py
@@ -19,10 +19,19 @@ MASTER_TRIGGERED_SUBJOB = 'MASTER_TRIGGERED_SUBJOB'
 SERVICE_STARTED = 'SERVICE_STARTED'
 SUBJOB_EXECUTION_FINISH = 'SUBJOB_EXECUTION_FINISH'
 SUBJOB_EXECUTION_START = 'SUBJOB_EXECUTION_START'
+LOG_CACHE_MAX_SIZE = 10000
 
 _analytics_logger = None
 _eventlog_file = None
 _event_id_generator = Counter()
+_log_cache = collections.deque(maxlen=LOG_CACHE_MAX_SIZE)
+
+
+def write_to_log_cache(event):
+    """
+    :type event: dict
+    """
+    _log_cache.append(event)
 
 
 def initialize(eventlog_file=None):
@@ -83,7 +92,9 @@ def record_event(tag, log_msg=None, **event_data):
         event_data['__id__'] = _event_id_generator.increment()
         event_data['__tag__'] = tag
         event_data['__timestamp__'] = time.time()
-        _analytics_logger.event(json.dumps(event_data, sort_keys=True))  # pylint: disable=no-member
+        json_dumps = json.dumps(event_data, sort_keys=True)
+        _analytics_logger.event(json_dumps)  # pylint: disable=no-member
+        write_to_log_cache(event_data)
         # todo(joey): cache most recent N events so get_events() doesn't always have to load file
 
     if log_msg:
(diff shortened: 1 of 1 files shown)


boxcla:

app/util/package_version.py
```diff
@@ -1 +1 @@
-version = '0.0.0'  # This should not be set to a real version. This file is modified during freeze.
+version = "0.5.197-b3b126e-mod"  # DO NOT COMMIT
```
josephharrington: WHAT HAVE YOU DONE?!

Just kidding. I should probably add a lint rule for the phrase "DO NOT COMMIT".

app/util/analytics.py
```diff
@@ -19,10 +19,28 @@
 SERVICE_STARTED = 'SERVICE_STARTED'
 SUBJOB_EXECUTION_FINISH = 'SUBJOB_EXECUTION_FINISH'
 SUBJOB_EXECUTION_START = 'SUBJOB_EXECUTION_START'
+LOG_CACHE_MAX_SIZE = 10000
 
 _analytics_logger = None
 _eventlog_file = None
 _event_id_generator = Counter()
+_log_cache = collections.deque([], LOG_CACHE_MAX_SIZE)
```
josephharrington: How about `collections.deque(maxlen=LOG_CACHE_MAX_SIZE)`?

Keyword args are one of my favorite things about Python. :neckbeard:

app/util/analytics.py
```diff
@@ -83,7 +101,9 @@ def record_event(tag, log_msg=None, **event_data):
         event_data['__id__'] = _event_id_generator.increment()
         event_data['__tag__'] = tag
         event_data['__timestamp__'] = time.time()
-        _analytics_logger.event(json.dumps(event_data, sort_keys=True))  # pylint: disable=no-member
+        json_dumps = json.dumps(event_data, sort_keys=True)
+        _analytics_logger.event(json_dumps)  # pylint: disable=no-member
+        write_to_log_cache(json_dumps)
```
josephharrington: Should we cache the un-jsonified event so we don't need to `json.loads` it again?

app/util/analytics.py
```diff
@@ -114,13 +134,27 @@ def get_events(since_timestamp=None, since_id=None):
     since_timestamp = float(since_timestamp) if since_timestamp else 0.0
     since_id = int(since_id) if since_id else 0
 
+    filtered_events_from_cache = get_events_from_cache(since_timestamp, since_id)
+    if len(filtered_events_from_cache) > 0:
+        return filtered_events_from_cache
```
josephharrington: What if the user has requested more events than are in the cache? I know it's probably not common but do you think we should support loading more than `LOG_CACHE_MAX_SIZE` events?

I think this is still better than what we have now. :smile:

app/util/analytics.py
```diff
@@ -114,24 +125,63 @@ def get_events(since_timestamp=None, since_id=None):
     since_timestamp = float(since_timestamp) if since_timestamp else 0.0
     since_id = int(since_id) if since_id else 0
 
-    with open(_eventlog_file, 'r') as f:
-        # todo(joey): This is inefficient since it reads the whole log file into memory (can be several megabytes).
-        # We probably want something like http://code.activestate.com/recipes/120686/
-        reversed_log_lines = reversed(f.readlines())
+    filtered_events_from_cache = get_events_from_cache(since_timestamp, since_id)
+    if len(filtered_events_from_cache) > 0:
```
wjdhollow: @josephharrington re: "What if the user requests more than LOG_CACHE_MAX_SIZE?"

It doesn't look like we currently have a way of a user request expressing that they want more than a certain number of entries.  Currently, we support the since_timestamp and since_id query parameters which will return any logs that happened after that date.  Since our logs roll over, there is no guarantee that the response will contain all the logs that occurred after a certain timestamp.  

That is why I do a check if any logs have been returned before falling back to check from the file.

josephharrington: Oh yeah that's what I meant -- the user might specify a timestamp from a long time ago, or a since_id of 0 (or just specify nothing at all) which may push the number of events we'd like to return over that `LOG_CACHE_MAX_SIZE` limit. 

It seems like in that case we'll just return the entire cache and then stop, even though it may be appropriate to go load even more events from the file.

Actually it seems like currently we would never load any events from the file at all since the cache would always win.

wjdhollow: Makes sense.  I'll update my PR and change event log so that it goes to disk if all of the logs since_timestamp were not found.  Currently, we default since_timestamp to 0 when it is not specified.  I think I will change it to something like a 5 hour look back by default.  

I will default since_id to None and omit that check unless it was explicitly requested.

josephharrington: Sounds good. Yeah we could just make it load the file any time the number of events returned from the cache equals `LOG_CACHE_MAX_SIZE` -- the annoying part is de-duping cached events and events from file, and also dealing with those file log rotations.

If you decide to make these improvements in pieces / multiple PRs, that's cool too. What you have now is still an improvement over my dumb implementation. :)

wjdhollow:
closing this PR.  opening another one with the refactor

PR #95, opened 2015-02-14
Add an executor_index for the entire build

drobertduke:
There is a use case for functional tests where a set
of logins are generated once, then each test executor
in a build needs to use a login that is not used by
any other executor in the build.

This adds a BUILD_EXEUCTOR_INDEX environment param
which stores a sequential executor id, unique across
the whole build. This can be used to index into a list
of logins created by the user.

Also adds a MACHINE_EXECUTOR_INDEX which is the same as
EXECUTOR_INDEX.

Creating this pull request early (no tests yet) so we can critique the approach.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -136,9 +136,14 @@ class Build(object):
         :type slave: Slave
         """
         self._slaves_allocated.append(slave)
-        self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
 
-        slave.setup(self.build_id(), project_type_params=self.build_request.build_parameters())
+        slave.setup(
+            self.build_id(),
+            project_type_params=self.build_request.build_parameters(),
+            num_executors_already_allocated=self._num_executors_allocated
+        )
+
+        self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
 
     def all_subjobs(self):
         """
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -140,7 +140,7 @@ class Build(object):
         slave.setup(
             self.build_id(),
             project_type_params=self.build_request.build_parameters(),
-            num_executors_already_allocated=self._num_executors_allocated
+            build_executor_start_index=self._num_executors_allocated
         )
 
         self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
--- a/app/project_type/project_type.py
+++ b/app/project_type/project_type.py
@@ -71,29 +71,18 @@ class ProjectType(object):
             config_contents = f.read()
         return config_contents
 
-    def fetch_project(self, executors=None, project_type_params=None):
+    def fetch_project(self):
         """
         Fetch the project onto the local machine.
 
         Runs once per machine per build. Runs the project_type's retrieve command (fetch, reset, etc) and produces
         a list of per-executor project_types.
-
-        :type executors: list [SubjobExecutor]
-        :type project_type_params: dict [str, str]
         """
-        if (executors is None) != (project_type_params is None):
-            raise RuntimeError('setup_build called with invalid params, either both executors and project_type_params '
-                               'should be set, or neither')
-
         self._fetch_project()
         self._logger.info('Build setup complete.')
 
         if self._remote_files:
             self._run_remote_file_setup()
-
-        # If executors were passed in, run configure_environment to do per-executor setup.
-        if executors and project_type_params:
-            self._setup_executors(executors, project_type_params)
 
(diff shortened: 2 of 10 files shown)


boxcla:

app/slave/cluster_slave.py
```diff
@@ -101,21 +104,23 @@ def setup_build(self, build_id, project_type_params):
...
+        :type num_executors_already_allocated: int
         """
-        # todo(joey): It's strange that the project_type is setting up the executors, which in turn set up projects.
-        # todo(joey): I think this can be untangled a bit -- we should call executor.configure_project_type() here.
+        self._base_executor_index = num_executors_already_allocated
         try:
-            self._project_type.fetch_project(executors, project_type_params)
+            for executor in executors:
+                executor.configure_project_type(project_type_params)
+            self._project_type.fetch_project()
```
josephharrington: :moneybag::moneybag::moneybag:

josephharrington:
I like it. Let's chat a spell about this in person this week.

app/project_type/project_type.py
```diff
@@ -71,30 +71,19 @@ def _get_config_contents(self):
...
         return config_contents
 
-    def fetch_project(self, executors=None, project_type_params=None):
+    def fetch_project(self):
         """
         Fetch the project onto the local machine.
 
         Runs once per machine per build. Runs the project_type's retrieve command (fetch, reset, etc) and produces
         a list of per-executor project_types.
-
```
tjlee0909: Interesting that you removed this stuff now. Though we have no use for this right now, we will have to add this kind of functionality once we add the per-executor setup and teardown.

app/slave/cluster_slave.py
tjlee0909: I believe the executor configuration should occur after the project has been fetched.

app/master/build.py
```diff
@@ -136,9 +136,14 @@ def allocate_slave(self, slave):
         :type slave: Slave
         """
         self._slaves_allocated.append(slave)
-        self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
 
-        slave.setup(self.build_id(), project_type_params=self.build_request.build_parameters())
+        slave.setup(
+            self.build_id(),
+            project_type_params=self.build_request.build_parameters(),
+            num_executors_already_allocated=self._num_executors_allocated
```
josephharrington: How do you feel about naming the argument that we send to the slave a more assertive name like `build_executor_start_index` -- I think that name explains what the slave will use it for vs. `num_executors_already_allocated` which is kind of a passive fact about existing state but does not indicate why it's being given to the slave.

app/slave/subjob_executor.py
```diff
@@ -82,6 +84,8 @@ def execute_subjob(self, build_id, subjob_id, subjob_artifact_dir, atomic_comman
                 'ARTIFACT_DIR': atom_artifact_dir,
                 'ATOM_ID': atom_id,
                 'EXECUTOR_INDEX': self.id,
```
josephharrington: Will EXECUTOR_INDEX be deprecated now?

test/unit/slave/test_cluster_slave.py
```diff
@@ -195,11 +196,28 @@ def test_async_setup_happy_path_invokes_correct_methods(self):
...
+        slave = self._create_cluster_slave()
+        slave._base_executor_index = 12
+        slave._master_api = Mock()
+        executor = Mock()
+        slave._idle_executors = Mock()
+
+        with patch.object(builtins, 'open', mock_open(read_data='asdf')):
+            slave._execute_subjob(1, 2, executor, '', [])
+
+        executor.execute_subjob.assert_called_with(1, 2, '', [], 12)
```
josephharrington: Should the tests use keyword args instead? I have no idea what this means.

josephharrington:
:+1: nice! A few minor comments.

tjlee0909:
:+1:

app/master/build.py
drobertduke: good point

app/slave/subjob_executor.py
drobertduke: yes.  I'll make a comment in the code, and then i should change any documentation examples.

test/unit/slave/test_cluster_slave.py
drobertduke: Adding keyword args on this line causes the test to fail, assert_called_with doesn't correctly compare a call using keywords with a call not using keywords I guess?  I added keywords to all the other test calls I could though

PR #96, opened 2015-02-15
Add lint rule for DO NOT COMMIT comments

josephharrington:
This is the comment we insert into package_version.py during freeze.
The file containing the DO NOT COMMIT comment is a temporary file, but
if something goes wrong during freeze then the file might persist. This
is just an extra check to help us notice when that happens.


boxcla:

test/framework/pylint/__init__.py
```diff
@@ -0,0 +1,8 @@
+from test.framework.pylint.clusterrunner_raw_checker import ClusterRunnerRawChecker
+
+
+def register(linter):
+    """
+    Register custom lint checkers with pylint. Any new checkers should also be registered here.
+    """
+    linter.register_checker(ClusterRunnerRawChecker(linter))
```
josephharrington: The reason for putting this register call in \_\_init.py\_\_ is so that if we add more checker classes, we'll only need to add another `register_checker` call in this file and won't need to change Makefile.

drobertduke:
cool :+1:

PR #97, opened 2015-02-23
Cache N event logs to improve API performance

wjdhollow:
- Writes event logs to cache up to N, or up to a maximum look back time (LOG_CACHE_EXPIRE_TIME_IN_HOURS). Whichever is greater
- Moves most of the logic in analytics to event_log for testability
- Items in the cache are examined for expiration only once the cache size exceeds LOG_CACHE_THRESHOLD


boxcla:

PR #100, opened 2015-02-26
Implement retry logic with binary exponential backoff.

tjlee0909:


boxcla:

app/util/decorators.py
```diff
@@ -27,6 +27,47 @@ def function_with_retries(*args, **kwargs):
...
+        @wraps(function)
+        def function_with_retries(*args, **kwargs):
+            # If initial_delay is negative, then exponentiation would go infinitely.
+            if initial_delay <= 0:
+                raise RuntimeError('initial_delay must be greater than 0, was set to {}'.format(str(initial_delay)))
+
+            # initial_delay must be smaller than max_delay.
+            if initial_delay > max_delay:
+                raise RuntimeError('initial_delay, {}, must be smaller than max_delay, {}'.format(
+                    str(initial_delay), str(max_delay)))
```
josephharrington: When using `.format` you shouldn't need to cast to str -- that's one of the main features of it. :)

app/util/decorators.py
```diff
@@ -27,6 +27,47 @@ def function_with_retries(*args, **kwargs):
     return method_decorator
 
 
+def retry_on_exception_exponential_backoff(exceptions, initial_delay=0.1, max_delay=10):
```
josephharrington: Another option would be to allow specification of the _total_ amount of delay (the sum of all the delays), and optionally the exponential scale factor, then we could express the linear retry decorator as a special case of the exponential one. I'm fine with this as you've written it though too.

app/util/decorators.py
```diff
@@ -27,6 +27,47 @@ def function_with_retries(*args, **kwargs):
...
+                    return_value = function(*args, **kwargs)
+                except exceptions as ex:
+                    if delay > max_delay:
+                        raise  # final attempt failed
+                    log.get_logger(__name__).warning('Call to {} raised {}("{}"). Retrying in {} seconds.',
+                                                     function.__qualname__, type(ex).__name__, ex, delay)
+                    time.sleep(delay)
+                    delay *= 2
+                else:
+                    return return_value
```
josephharrington: I can't remember why we put this in the `else` block in the original decorator, but it seems like we could just do `return function(*args, **kwargs)` in the try block and this should work the same?

app/util/network.py
```diff
@@ -39,7 +39,7 @@ def get(self, *args, **kwargs):
         return self._request('GET', *args, **kwargs)
 
     # todo: may be a bad idea to retry -- what if post was successful but just had a response error?
-    @retry_on_exception(exceptions=(requests.ConnectionError,))
+    @retry_on_exception_exponential_backoff(exceptions=(requests.ConnectionError,))
```
josephharrington: Should we add the retries to get requests as well?

josephharrington:
:+1:

app/util/decorators.py
tjlee0909: Done

app/util/decorators.py
tjlee0909: Yeah, I was worried about adding too many arguments, but four isn't too bad, and it's likely not going to increase in the future.

app/util/network.py
tjlee0909: Might as well.

app/util/decorators.py
tjlee0909: Actually that'd only be three arguments, as the max_delay argument would go away if we introduced the total amount of delay.

app/util/decorators.py
tjlee0909: Done

app/util/decorators.py
```diff
@@ -6,7 +6,7 @@
 from app.util.secret import Secret
 
 
-def retry_on_exception(exceptions, num_attempts=10, retry_delay=0.1):
+def retry_on_exception_linear(exceptions, num_attempts=10, retry_delay=0.1):
```
josephharrington: Now I think this could just be:
```python
def retry_on_exception_linear(exceptions, num_attempts=10, retry_delay=0.1):
    return retry_on_exception_exponential_backoff(exceptions, initial_delay=0.1, total_delay=num_attempts * retry_delay, exponential_factor=1):
```

Or I guess we could just delete it!

josephharrington:
:+1: after test fixes.

PR #102, opened 2015-03-05
Kill child procs on stop

wjdhollow:
This fixes an issue where child processes are not properly stopped by the stop subcommand.

### DIFF
--- a/app/subcommands/stop_subcommand.py
+++ b/app/subcommands/stop_subcommand.py
@@ -13,7 +13,7 @@ class StopSubcommand(Subcommand):
     # The commands that can be used to identify a clusterrunner process (for accurate killing).
     _command_whitelist_keywords = ['clusterrunner', 'main.py master', 'main.py slave']
     # The number of seconds to wait between performing a SIGTERM and a SIGKILL
-    _sigterm_sigkill_grace_period_sec = 2
+    SIGTERM_SIGKILL_GRACE_PERIOD_SEC = 2
 
     def run(self, log_level):
         """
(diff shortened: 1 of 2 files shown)


boxcla:

app/subcommands/stop_subcommand.py
```diff
@@ -62,23 +62,37 @@ def _kill_pid_in_file_if_exists(self, pid_file_path):
...
         # Try killing gracefully with SIGTERM first. Then give process some time to gracefully shutdown. If it
         # doesn't, perform a SIGKILL.
         # @TODO: use util.timeout functionality once it gets merged
-        os.kill(int(pid), signal.SIGTERM)
+        self.kill_proc_and_children(proc, signal.SIGTERM)
         sigterm_start = time.time()
 
         while (time.time()-sigterm_start) <= self._sigterm_sigkill_grace_period_sec:
-            if not psutil.pid_exists(int(pid)):
+            if not proc.is_running():
```
josephharrington: What if the parent process dies in response to SIGTERM but the child processes don't? It seems like this logic would not send a SIGKILL to the children in that case.

app/subcommands/stop_subcommand.py
```diff
@@ -62,23 +62,33 @@ def _kill_pid_in_file_if_exists(self, pid_file_path):
...
         # @TODO: use util.timeout functionality once it gets merged
-        os.kill(int(pid), signal.SIGTERM)
+        procs_to_kill = proc.get_children() + [proc]
+        self.kill_running_procs(procs_to_kill, signal.SIGTERM)
         sigterm_start = time.time()
 
-        while (time.time()-sigterm_start) <= self._sigterm_sigkill_grace_period_sec:
-            if not psutil.pid_exists(int(pid)):
+        while (time.time()-sigterm_start) <= self.SIGTERM_SIGKILL_GRACE_PERIOD_SEC:
+            if all([not ptk.is_running() for ptk in procs_to_kill]):
```
josephharrington: `proc_to_kill` instead of `ptk`? If there's a perfect level of verbosity in naming, I'd rather be on the more verbose side. I think a little goes a long way in readability. :)

josephharrington: Oh also, you could also go with `if not any([proc_to_kill.is_running()...` but I am having an internal debate right now about which is more intuitive.

app/subcommands/stop_subcommand.py
```diff
@@ -62,23 +62,33 @@ def _kill_pid_in_file_if_exists(self, pid_file_path):
...
 
         self._logger.info("Killed process with PID {0} with SIGTERM".format(pid))
+
+    def kill_running_procs(self, procs_to_kill, sig):
+        """
+        :type procs_to_kill: list[psutil.Process]
+        :type sig: int
+        """
+        running_procs = [proc for proc in procs_to_kill if proc.is_running()]
+        for proc in running_procs:
```
josephharrington: Should we move the above logger.info message into this method?

app/subcommands/stop_subcommand.py
wjdhollow: Done and Done.  Doing a not within the `all` smelled funny to me.  I prefer it on the outside the way you proposed.

app/subcommands/stop_subcommand.py
```diff
@@ -62,23 +62,33 @@ def _kill_pid_in_file_if_exists(self, pid_file_path):
...
             self._logger.info(
-                "PID {0} is running, but command '{1}' is not a clusterrunner command".format(pid, proc_command))
+                "PID {0} is running, but command '{1}' is not a clusterrunner command".format(proc.pid, proc_command))
             return
 
         # Try killing gracefully with SIGTERM first. Then give process some time to gracefully shutdown. If it
         # doesn't, perform a SIGKILL.
         # @TODO: use util.timeout functionality once it gets merged
-        os.kill(int(pid), signal.SIGTERM)
+        procs_to_kill = proc.get_children() + [proc]
```
tjlee0909: get_children() is deprecated. We should use proc.children(recursive=True)

josephharrington: Ooh, nice catch.

wjdhollow: Done.

app/subcommands/stop_subcommand.py
wjdhollow: Done.

tjlee0909:
:+1:

PR #105, opened 2015-03-14
Fix git fetching

### LINKED ISSUE #103: Git builds on non-hash specific refs cause ClusterRunner slaves to go offline
When launching builds with branch set to a re-used branch (ie: v5-dev, HEAD, dev, etc) ClusterRunner will break on fetching on the slave machines on subsequent runs (not the first runs) for git builds with that branch, causing slaves to go offline.

josephharrington:
There is a race condition right now where if a git branch is specified
in a build request, slaves will pull from the master using that branch
name (whose head may change when the master handles future builds).

This change creates a local ref on the master whenever a git build is
requested. This ref is then passed to the slaves instead of the user-
specified branch name so that the slaves are guaranteed to always fetch
the same commit.

As part of this change, I also refactored how the slave build parameter
overriding works.

This will fix #103.

### DIFF
--- a/app/master/slave.py
+++ b/app/master/slave.py
@@ -4,7 +4,6 @@ from app.util.network import Network
 from app.util.safe_thread import SafeThread
 from app.util.secret import Secret
 from app.util.url_builder import UrlBuilder
-from app.util import util
 import requests.exceptions
 
 
@@ -48,25 +47,24 @@ class Slave(object):
 
         self.current_build_id = None
 
-    def setup(self, build_id, project_type_params, build_executor_start_index):
+    def setup(self, build):
         """
-        Execute a setup command on the slave for the specified build. The command is executed asynchronously from the
-        perspective of this method, but any subjobs will block until the slave finishes executing the setup command.
+        Execute a setup command on the slave for the specified build. The setup process executes asynchronously on the
+        slave and the slave will alert the master when setup is complete and it is ready to start working on subjobs.
 
-        :param build_id: The build id that this setup command is for.
-        :type build_id: int
-
-        :param project_type_params: The parameters that define the project type this build will execute in
-        :type project_type_params: dict
+        :param build: The build to set up this slave to work on
+        :type build: Build
         """
-        setup_url = self._slave_api.url('build', build_id, 'setup')
-        slave_project_type_params = util.project_type_params_for_slave(project_type_params)
+        slave_project_type_params = build.build_request.build_parameters().copy()
+        slave_project_type_params.update(build.project_type.slave_param_overrides())
+
+        setup_url = self._slave_api.url('build', build.build_id(), 'setup')
         post_data = {
             'project_type_params': slave_project_type_params,
-            'build_executor_start_index': build_executor_start_index,
+            'build_executor_start_index': build.num_executors_allocated,
         }
         self._network.post_with_digest(setup_url, post_data, Secret.get())
-        self.current_build_id = build_id
+        self.current_build_id = build.build_id()
 
     def teardown(self):
         """
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -57,6 +57,7 @@ class Git(ProjectType):
 
     # pylint: disable=redefined-builtin
     # Disable "redefined-builtin" because renaming the "hash" parameter would be a breaking change.
+    # todo: Deprecate the "branch" parameter and create a new one named "ref" to replace it.
     def __init__(self, url, build_project_directory='', project_directory='', remote='origin', branch='master',
                  hash='FETCH_HEAD', config=None, job_name=None, remote_files=None):
         """
@@ -90,6 +91,7 @@ class Git(ProjectType):
         self._repo_directory = self.get_full_repo_directory(self._url)
         self._timing_file_directory = self.get_timing_file_directory(self._url)
         self._git_remote_command_executor = _GitRemoteCommandExecutor()
+        self._local_ref = None
 
         # We explicitly set the repo directory to 700 so we don't inadvertently expose the repo to access by other users
         fs.create_dir(self._repo_directory, self.DIRECTORY_PERMISSIONS)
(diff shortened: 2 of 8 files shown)


boxcla:

app/project_type/git.py
```diff
@@ -149,25 +150,38 @@ def _fetch_project(self):
 
         # Must add the --update-head-ok in the scenario that the current branch of the working directory
         # is equal to self._branch, otherwise the git fetch will exit with a non-zero exit code.
-        # Must specify the colon in 'branch:branch' so that the branch will be created locally. This is
-        # important because it allows the slave hosts to do a git fetch from the master for this branch.
-        fetch_command = 'git fetch --update-head-ok {0} {1}:{1}'.format(self._remote, self._branch)
+        fetch_command = 'git fetch --update-head-ok {} {}'.format(self._remote, self._branch)
```
tjlee0909: Yeah, we really don't need to create the branch locally as long as we have the FETCH_HEAD. +1.

I'm guessing this line also gets rid of the possibility of throwing that fast-forward error when you do a "git fetch scm REMOTE_BRANCH:LOCAL_BRANCH"  in the scenario where the remote branch has somehow gotten behind the local branch.

app/master/slave.py
```diff
@@ -48,25 +47,24 @@ def mark_as_idle(self):
 
         self.current_build_id = None
 
-    def setup(self, build_id, project_type_params, build_executor_start_index):
+    def setup(self, build):
```
tjlee0909: Yay, so much nicer. I think originally we (I?) were too worried about making law of demeter violations, but this is better and makes sense.

app/project_type/git.py
```diff
@@ -128,6 +109,26 @@ def __init__(self, url, build_project_directory='', project_directory='', remote
...
+        param_overrides = super().slave_param_overrides()
+
+        # We modify the repo url so the slave clones or fetches from the master directly. This should be faster than
+        # cloning/fetching from the original git remote.
+        master_repo_url = 'ssh://{}{}'.format(Configuration['hostname'], self._repo_directory)
+        param_overrides['url'] = master_repo_url  # This causes the slave to clone directly from the master.
+
+        # The user-specified branch is overwritten with a locally created ref so that slaves working on a job can
+        # continue to fetch the same HEAD, even if the master resets the user-specified branch for another build.
+        param_overrides['branch'] = self._branch
```
tjlee0909: It took a little while wrapping my head around the order in which the local_ref gets created in the first place (on the master upon build preparation) and when that gets propagated to the slave.

One of the worries I had about this line is that it may not be clear whether this._branch has already been properly populated. There is nothing we can do about getting the local ref value, however I think it'd be slightly easier to understand and not screw-up the ordering of (in future PR's from other developers) if you had a separate local variable called self._local_ref.

Maybe in line 163 you can set self._local_ref instead of a temporary variable, and not overwrite self._branch at all. And then in this line (line 128) you could set param_overrides['branch'] = self._local_ref. This way it's really obvious that the slave_param_overrides is taking the 'branch' argument from self._local_ref of the master and it's really obvious where that gets set (self._fetch_project).

app/project_type/git.py
```diff
@@ -149,25 +150,38 @@ def _fetch_project(self):
...
+        # Validate and convert the user-specified hash/refspec to a full git hash
+        self._hash = self._execute_in_repo_and_raise_on_failure(
+            'git rev-parse {}'.format(self._hash),
+            'Could not rev-parse "{}" to a commit hash.'.format(self._hash)
+        ).strip()
+
+        # Save this hash as a local ref. Named local refs are necessary for slaves to fetch correctly from the master.
+        local_ref = 'refs/clusterrunner/' + self._hash
+        update_ref_command = 'git update-ref {} {}'.format(local_ref, self._hash)
+        self._execute_in_repo_and_raise_on_failure(update_ref_command, 'Could not update local ref.')
```
tjlee0909: Maybe include the local_ref and or self._hash in the error message? I'm sorry, you may have to go over 100 lines =P

app/project_type/git.py
```diff
@@ -77,7 +58,7 @@ def get_timing_file_directory(url):
     # pylint: disable=redefined-builtin
     # Disable "redefined-builtin" because renaming the "hash" parameter would be a breaking change.
     def __init__(self, url, build_project_directory='', project_directory='', remote='origin', branch='master',
```
tjlee0909: This might be out of scope for the PR, or maybe not--I think it's time we rename the 'branch' argument to 'ref'.

tjlee0909:
Some minorish comments, but mostly LGTM.

app/project_type/git.py
josephharrington: Nice. I like that.

app/project_type/git.py
josephharrington: Ah true, I'll add that as a todo.

app/project_type/git.py
josephharrington: Good point! I double checked and the `_execute_in_repo_and_raise_on_failure` method does include the attempted command in the exception message, so I think it should provide enough info if we hit this error.

tjlee0909:
:+1:

app/project_type/git.py
```diff
@@ -128,6 +111,26 @@ def __init__(self, url, build_project_directory='', project_directory='', remote
...
+
+        :return: A set of values to override original project type params
+        :rtype: dict[str, str]
+        """
+        param_overrides = super().slave_param_overrides()
+
+        # We modify the repo url so the slave clones or fetches from the master directly. This should be faster than
+        # cloning/fetching from the original git remote.
+        master_repo_url = 'ssh://{}{}'.format(Configuration['hostname'], self._repo_directory)
+        param_overrides['url'] = master_repo_url  # This causes the slave to clone directly from the master.
```
drobertduke: Cool!

drobertduke:
good stuff :+1:

PR #110, opened 2015-03-18
Fix slave stuck on subjob completion error

### LINKED ISSUE #98: If a slave fails to post results back to master, there's no retry or error logic to handle it, slave hangs indefinitely
I'm running into a situation while running tests where we will see sporadic failures after a slave finished executing a subjob and calls Network.post to send the results back to master. If the POST call fails, the slave does not check the return code and will wait for new jobs to be assigned indefinitely. This causes the build to hang because the master hasn't received results back for all the subjobs. I'm trying to add some additional logging to some of the slave classes and/or retry logic on the post call, but I'm not too familiar with the code. I've attached some of the logs from the slave below, let me know if you see something.

```
[2015-02-19 11:06:18.634] DEBUG   Bld63-Sub15     project_type    Command completed with exit code 0.
[2015-02-19 11:06:25.852] WARNING Bld63-Sub15     decorators      Call to Network.post raised ConnectionError("HTTPConnectionPool(host='ip-10-110-90-25.us-west-2.compute.internal', port=43000): Max retries exceeded with url: /v1/build/63/subjob/15/result (Caused by <class 'ConnectionResetError'>: [Errno 104] Connection reset by peer)"). Retrying in 0.1 seconds.
[2015-02-19 11:06:25.970] INFO    Bld63-Sub15     cluster_slave   Build 63, Subjob 15 completed and sent results to master.
```

josephharrington:
This addresses an issue in which any error raised during subjob
completion (e.g., results tar file creation fails) would cause the
slave and build to hang indefinitely.

This change combined with #100 should fix #98.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -176,12 +176,12 @@ class Build(object):
                 break
             slave.claim_executor()
             self._num_executors_in_use += 1
-            self.execute_next_subjob_on_slave(slave)
+            self.execute_next_subjob_or_teardown_slave(slave)
 
-    def execute_next_subjob_on_slave(self, slave):
+    def execute_next_subjob_or_teardown_slave(self, slave):
         """
         Grabs an unstarted subjob off the queue and sends it to the specified slave to be executed. If the unstarted
-        subjob queue is empty, we mark the slave as idle.
+        subjob queue is empty, we teardown the slave to free it up for other builds.
 
         :type slave: Slave
         """
@@ -207,8 +207,14 @@ class Build(object):
         :type subjob_id: int
         :type payload: dict
         """
-        self._handle_subjob_payload(subjob_id, payload)
-        self._mark_subjob_complete(subjob_id)
+        try:
+            self._handle_subjob_payload(subjob_id, payload)
+            self._mark_subjob_complete(subjob_id)
+
+        except Exception:
+            self._logger.exception('Error while completing subjob; marking build as failed.')
+            self.mark_failed('Error occurred while completing subjob {}.'.format(subjob_id))
+            raise
 
     def _handle_subjob_payload(self, subjob_id, payload):
         if not payload:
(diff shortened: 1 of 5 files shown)


boxcla:

app/master/build.py
```diff
@@ -176,12 +176,12 @@ def begin_subjob_executions_on_slave(self, slave):
                 break
             slave.claim_executor()
             self._num_executors_in_use += 1
-            self.execute_next_subjob_on_slave(slave)
+            self.execute_next_subjob_or_teardown_slave(slave)
 
-    def execute_next_subjob_on_slave(self, slave):
+    def execute_next_subjob_or_teardown_slave(self, slave):
```
josephharrington: This rename is just to make it more descriptive as to the actual behavior.

test/unit/master/test_cluster_master.py
```diff
@@ -128,17 +132,32 @@ def test_handle_result_reported_from_slave_does_nothing_when_build_is_canceled(s
         build_id = 1
         slave_url = "url"
         build = Build(BuildRequest({}))
-        build.handle_subjob_payload = Mock()
```
josephharrington: Fun fact: because of the way this test was written it was still passing even though `handle_subjob_payload` was no longer a real method.

app/master/cluster_master.py
```diff
@@ -234,8 +234,10 @@ def handle_result_reported_from_slave(self, slave_url, build_id, subjob_id, payl
         slave = self._all_slaves_by_url[slave_url]
         # If the build has been canceled, don't work on the next subjob.
         if not build.is_finished:
-            build.complete_subjob(subjob_id, payload)
-            build.execute_next_subjob_on_slave(slave)
+            try:
+                build.complete_subjob(subjob_id, payload)
+            finally:
```
tjlee0909: Ok, so these 2 two additional lines are the fix.

tjlee0909:
:+1:

PR #112, opened 2015-03-24
Fix test setup loophole

josephharrington:
Currently we attempt to enforce that super().setUp() is called from all
our test classes that override setUp(). We do this by making an
assertion in tearDown(). The loophole: if the test class overrides both
setUp() and tearDown(), the assertion will never be made.

This change closes the loophole, making it much harder to accidentally
subvert this check.


boxcla:

wjdhollow:
:+1: nice

PR #115, opened 2015-03-27
Fix event log caching

wjdhollow:
This change modifies the default behavior of how the event_log query works. Previously, if you specified a timestamp or id and they were not encountered, an empty list of events was returned. 

Instead, we return all the events that are in the log file.


boxcla:

app/util/analytics.py
```diff
@@ -22,9 +11,14 @@
...
-_eventlog_file = None
-_event_id_generator = Counter()
+_event_log = None
+
+
+def write_to_log_cache(event):
+    """
+    :type event: dict
+    """
+    # _log_cache.append(event)
```
josephharrington: What is this for?

app/util/analytics.py
```diff
@@ -105,35 +65,9 @@ def get_events(since_timestamp=None, since_id=None):
     :return: The list of events in the given range
     :rtype: list[dict] | None
     """
-    # eventlogs to STDOUT should only be used in debug situations, so don't worry about making events available.
-    if _eventlog_file is None or _eventlog_file.upper() == 'STDOUT':
+    if isinstance(_event_log, EventLog):
```
josephharrington: What do you think of doing `if _event_log:` here just like up above on line 52? Were you worried about something specific with `_event_log` being some other object?

app/util/analytics.py
```diff
@@ -105,35 +65,9 @@ def get_events(since_timestamp=None, since_id=None):
...
+        since_timestamp = float(since_timestamp) if since_timestamp else since_timestamp
+        since_id = int(since_id) if since_id else since_id
+        return _event_log.get_events(since_timestamp=since_timestamp, since_id=since_id)
+    else:
```
josephharrington: Should we raise an exception if a user calls `get_events` without initializing first?

app/util/analytics.py
```diff
@@ -37,35 +31,9 @@ def initialize(eventlog_file=None):
...
 
-    event_handler.format_string = '{record.message}'  # only output raw log message -- no timestamp or log level
-    handler = TaggingHandler(
-        {'event': event_handler},  # enable logging to the event_handler with the event() method
-        bubble=True,
-    )
-    handler.push_application()
-
-    _analytics_logger = TaggingLogger('analytics', ['event'])
+    _event_log = EventLog(filename=eventlog_file, disable_logging=eventlog_file is None)
```
josephharrington: Should the EventLog class just decide for itself if logging is disabled since it already has the value for eventlog_file?

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+        # only output raw log message -- no timestamp or log level
+        event_handler = self._get_event_handler()
+        event_handler.format_string = '{record.message}'
+        handler = TaggingHandler(
+            {'event': event_handler},
+            bubble=True,
+        )
+        handler.push_application()
+
+    def _initalize_analytics_logger(self):
```
josephharrington: Typo: "initalize" -> "initialize"

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+        The time is considered stale if it is is older than start_time + LOG_CACHE_EXPIRE_TIME_IN_HOURS
+
+        :rtype: bool
+        """
+        if len(self._log_cache) == 0:
+            return False
+        else:
+            latest = self._latest_timestamp_in_cache()
+            oldest = self._oldest_timestamp_in_cache()
+            return oldest + LOG_CACHE_EXPIRE_TIME_IN_HOURS * 60 * 60 < latest
```
josephharrington: This is cool! Do you think it would make sense to simplify this and not look at the latest timestamp at all? Presumably we'd only be calling `_expire_stale_items_in_cache` when we add a new event, so the latest timestamp would pretty much always be the current timestamp. Does that sound right? I was thinking that change would simplify this logic significantly but if not, then I think it's fine.

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+    def record_event(self, tag, log_msg=None, **event_data):
+        """
+        Record an event containing the specified data. Currently this just json-ifies the event and outputs it to
+        the configured analytics logger (see analytics.initialize()).
+
+        :param tag: A string identifier that describes the event being logged (e.g., "REQUEST_SENT")
+        :type tag: str
+        :param log_msg: A message that will also be logged to the human-readable log (not the event log). It will be
+            string formatted with tthe event_data dict.  This is a convenience for logging to both human- and
+            machin-readable logs.
```
josephharrington: typozzz: "tthe", "machin-readable" :neckbeard:

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+        significantly older than the most recent item. This expiration time is determined by the constant,
+        LOG_CACHE_EXPIRE_TIME_IN_HOURS.
+
+        Since we do not want to artificially impose limits on the log cache, we will not expire items
+        if the total length of the log cache is less than LOG_CACHE_THRESHOLD.
+        """
+        while len(self._log_cache) > LOG_CACHE_THRESHOLD and self.oldest_cache_event_is_stale():
+            self._log_cache.popleft()
+
+    def oldest_cache_event_is_stale(self):
```
josephharrington: Any reason this is public?

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+            latest = self._latest_timestamp_in_cache()
+            oldest = self._oldest_timestamp_in_cache()
+            return oldest + LOG_CACHE_EXPIRE_TIME_IN_HOURS * 60 * 60 < latest
+
+    def _latest_timestamp_in_cache(self):
+        """
+        :rtype: float | None
+        """
+        if len(self._log_cache) == 0:
+            return None
```
josephharrington: It looks like returning None from this and the next method might cause a `TypeError` exception on line 155 when we try to use the None in a calculation?

josephharrington: Ah, I see that currently hitting that path would be impossible since you do the length == 0 check in all three places. Do you think it might be slightly more robust if instead of doing the length == 0 check on line 150 if we just caught the TypeError and returned False? (To make it more explicit what we are guarding against?)

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+                since_id=since_id
+            )
+
+        return events
+
+    def _should_try_get_event_from_log_cache(self, since_id=None,
+                                             since_timestamp=None):
+        if since_id is None and since_timestamp is None:
+            return True
+        elif since_timestamp and since_timestamp <= self._oldest_timestamp_in_cache():
```
josephharrington: Seems like we should have a similar check for since_id? Alternatively we could remove the option to provide since_id completely -- it seems enormously rare that since_id would behave differently from since_timestamp. 

(The one example I can think of is that two consecutive events have the exact same timestamp but a client somehow only received one of those events in an API request -- then the next request it made would be missing the second event. That seems extremely unlikely.)

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+import sys
+import time
+
+from app.util import fs, log
+from app.util.conf.configuration import Configuration
+from app.util.counter import Counter
+
+
+LOG_CACHE_EXPIRE_TIME_IN_HOURS = 5
+LOG_CACHE_THRESHOLD = 10000
```
josephharrington: What do you think of cranking this up to like 50k or 100k? I just loaded our timeline dashboard and it is not even super tightly packed but still loaded 12k events. (Each of our PHPUnit jobs has ~1k subjobs, with each subjob having 2 events each, so it would only take 5 phpunit jobs within half an hour to surpass a cache of 10k.)

app/util/event_log.py
```diff
@@ -0,0 +1,265 @@
...
+        :return: The list of events in the given range
+        :rtype: list[dict] | None
+        """
+        if self.logging_disabled or self.filename.upper() == 'STDOUT':
+            return None
+
+        if since_timestamp is not None and since_id is not None:
+            raise ValueError('since_timestamp and since_id can not be used at the same time')
+        events = []
+        condition_met = False
```
josephharrington: As soon as I read the name `condition_met`, my mind immediately goes "What condition?!" :bowtie: 

What do you think of something like `is_query_complete`?

---

As I look at this more, it seems like the `condition_met` flag might not be doing much? It looks like the flow is:


Alternatively, we could just do the condition check in this method after querying the cache instead of having the `_get_events_from_reversed_generator` method do it. Something like:
```
if not self._query_extends_beyond_cache(since_timestamp):
    events = self._get_event_from_cache(since_timestamp)


```

app/util/analytics.py
wjdhollow: It was an easter egg I left for reviewers. Congratulations, you win a prize :sparkles: :gift_heart: 

... actually it was me being sloppy. I left in code I abandoned during my change :blush: 
good catch :smile:

app/util/analytics.py
wjdhollow: I'll remove the extra parameter

app/util/analytics.py
wjdhollow: Done

app/util/analytics.py
wjdhollow: Done.

app/util/event_log.py
wjdhollow: Sure, why not. I think that would just be a cache on the order of 10s MB at most. We can always drop it later if we see problems.

app/util/event_log.py
wjdhollow: Done. :blush:

app/util/event_log.py
wjdhollow: Done. :blush:

app/util/event_log.py
wjdhollow: Done. I'll change it to do the TypeError check and I'll update to use the latest timestamp.

app/util/event_log.py
wjdhollow: +1 for error checking using exceptions. http://www.jeffknupp.com/blog/2013/02/06/write-cleaner-python-use-exceptions/

I can see now why Python leans towards the EAFP camp

app/util/event_log.py
wjdhollow: Good point. I think this can be simplified to just do one boolean check.

app/util/event_log.py
josephharrington: Just to make sure we're on the same page -- I think that if we keep the option for the user to specify either id or timestamp, then right here we need to check both the id **and** the timestamp to determine if the cache has all the data we want since we'll only ever have at most one of id or timestamp values specified.

josephharrington:
:+1: I think we should go ahead and merge this before we forget what it does. :stuck_out_tongue:

PR #118, opened 2015-04-09
Add shutdown mode for slaves

drobertduke:
- Slaves now have a shutdown mode where they will not accept
additional subjobs.  The slave process will terminate when all
executors are free.
- The master now provides an API for shutting down a named set of
slaves or shutting down all slaves.
- Adds shutdown functionality to the client and the command line.
- This commit also adds missing test coverage to slave_allocator
and slave.


boxcla:

test/functional/job_configs.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, config, expected_to_fail, expected_num_subjobs, expected_num_
     teardown_build:
         - echo "Doing build teardown."
         - sleep 1
-        - ALL_SUBJOB_FILES=$(ls $PROJECT_DIR/subjob_file_*.txt)
+        - ALL_SUBJOB_FILES=$(ls $PROJECT_DIR/subjob_file_*.txt || mktemp -t clusterrunner)
```
drobertduke: Had to add this when I found out that putting a slave in shutdown mode may cause it to run teardown before it has run any subjobs.

app/master/build.py
```diff
@@ -182,20 +184,32 @@ def execute_next_subjob_or_teardown_slave(self, slave):
...
+            # build would be stuck.
+            with self._subjob_assignment_lock:
+                subjob = self._unstarted_subjobs.get(block=False)
+                self._logger.debug('Sending subjob {} (build {}) to slave {}.',
+                                   subjob.subjob_id(), subjob.build_id(), slave.url)
+                try:
+                    slave.start_subjob(subjob)
+                except ShutdownSlaveError:
+                    self._unstarted_subjobs.put(subjob)
+                    self._free_slave_executor(slave)
```
josephharrington: Want to add a comment here about why we free the executor? It's because the slave is not really shut down, correct? Like at this point we're just waiting for all the executors to free up so we can call teardown. I wonder if we should rename the `ShutdownSlaveError` to something that signifies that it's an ongoing process like `SlaveShuttingDownError` or `SlaveMarkedForShutdownError`?

app/client/cluster_api_client.py
```diff
@@ -76,12 +87,52 @@ def get_build_status(self, build_id):
                                             'URL: {}, Content:{}'.format(build_status_url, response_data))
         return response_data
 
+    def block_until_build_started(self, build_id, timeout=None, build_in_progress_callback=None):
+        """
+        Poll the build status endpoint until the build has started or finished.
```
wjdhollow: This comment is confusing. Should it be updated or maybe the method should be renamed?

app/client/cluster_api_client.py
```diff
@@ -99,6 +150,38 @@ def is_build_finished():
 
         poll.wait_for(is_build_finished, timeout_seconds=timeout)
 
+    def get_slaves(self):
+        slave_url = self._api.url('slave')
+        response = self._network.get(slave_url)
+        return response.json()
+
```
wjdhollow: These methods are so beautiful I want to cry :+1:

app/client/cluster_api_client.py
```diff
@@ -99,6 +150,38 @@ def is_build_finished():
 
         poll.wait_for(is_build_finished, timeout_seconds=timeout)
 
+    def get_slaves(self):
+        slave_url = self._api.url('slave')
```
wjdhollow: Is this method only for testing purposes? Method docs?

app/master/slave.py
```diff
@@ -156,3 +167,45 @@ def set_is_alive(self, value):
...
+    def is_shutdown(self):
+        """
+        Whether the slave is in shutdown mode.
+        """
+        return self._is_shutdown
+
+    def kill(self):
+        """
+        Instructs the slave process to kill itself.
+        """
```
wjdhollow: Just curious, why did you name this the `setup_url`?

app/master/slave.py
```diff
@@ -47,6 +50,10 @@ def mark_as_idle(self):
 
         self.current_build_id = None
 
+        if self._is_shutdown:
```
wjdhollow: I noticed in ClusterSlave.kill() there is still a todo comment to shutdown more gracefully. Correct me if I am wrong, but I feel this change addresses that todo. Maybe we should remove it, or open an issue for it.

app/master/build.py
```diff
@@ -182,20 +184,32 @@ def execute_next_subjob_or_teardown_slave(self, slave):
...
         """
         try:
-            subjob = self._unstarted_subjobs.get(block=False)
-            self._logger.debug('Sending subjob {} (build {}) to slave {}.',
-                               subjob.subjob_id(), subjob.build_id(), slave.url)
-            slave.start_subjob(subjob)
+            # This lock prevents the scenario where a subjob is pulled from the queue but cannot be assigned to this
+            # slave because it is shutdown, so we put it back on the queue but in the meantime another slave enters
+            # this method, finds the subjob queue empty, and is torn down.  If that was the last 'living' slave, the
+            # build would be stuck.
```
josephharrington: Awesome.

app/master/build.py
```diff
@@ -182,20 +184,32 @@ def execute_next_subjob_or_teardown_slave(self, slave):
...
-            if num_executors_in_use == 0:
-                try:
-                    self._slaves_allocated.remove(slave)
-                except ValueError:
-                    pass  # We have already deallocated this slave, no need to teardown
-                else:
-                    slave.teardown()
+            self._free_slave_executor(slave)
+
+    def _free_slave_executor(self, slave):
```
josephharrington: What do you think of "`_free_slave_executor_or_teardown`"?

app/master/cluster_master.py
```diff
@@ -143,6 +144,15 @@ def handle_slave_state_update(self, slave, new_slave_state):
         do_transition = slave_transition_functions.get(new_slave_state)
         do_transition(slave)
 
+    def _shutdown_slave(self, slave):
```
josephharrington: I think we should be careful about making a clear distinction between "kill" and this new shutdown functionality. Like maybe we want to call it "graceful shutdown" or something like that?

app/master/cluster_master.py
```diff
@@ -143,6 +144,15 @@ def handle_slave_state_update(self, slave, new_slave_state):
...
+        """
+        Puts slave in shutdown mode so it cannot receive new builds. The slave will be killed when finished with any
+        running builds.
+        :type slave: Slave
+        """
+        slave.set_shutdown_mode()
+        self._logger.info('Slave on {} was put in shutdown mode. (id: {})', slave.url, slave.id)
```
josephharrington: Do we also want to remove the slave from the `_all_slaves_by_url` dict? This is what the API uses to display a list of slaves, so I feel like we might just want to remove it from the API's list of slaves when we've explicitly told a slave to shut down. (E.g., it should just disappear from the slave monitor dashboard, not show up as red?)

app/master/slave.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, slave_url, num_executors):
         self._network = Network(min_connection_poolsize=num_executors)
         self.current_build_id = None
         self._is_alive = True
+        self._is_shutdown = False
```
josephharrington: Extension from earlier comment: maybe something like "is_shutting_down"?

app/web_framework/cluster_master_application.py
```diff
@@ -253,3 +255,15 @@ def get(self):
...
+
+class _SlaveShutdownHandler(_ClusterMasterBaseHandler):
+    @authenticated
+    def post(self):
+        shutdown_all = self.decoded_body.get('shutdown_all')
+        slaves_to_shutdown = self._cluster_master.all_slaves_by_id().keys() if shutdown_all else\
+            [int(slave_id) for slave_id in self.decoded_body.get('slaves')]
+
+        for slave_id in slaves_to_shutdown:
+            slave = self._cluster_master.get_slave(slave_id)
```
josephharrington: Should we assemble the entire list of slaves first before doing shutdown on any of them? If the user has a single invalid slave id halfway through the list, only the first half will be shutdown, and then this will return a 404 error.

Also/alternatively, should we return a response with the list of successfully shutdown slaves?

app/web_framework/cluster_slave_application.py
```diff
@@ -145,8 +145,10 @@ def get(self):
 
 
 class _KillHandler(_ClusterSlaveBaseHandler):
+    @authenticated
     def post(self):
         self._write_status()
+        # self._cluster_slave.kill()
```
josephharrington: Did you mean to remove this?

test/functional/job_configs.py
josephharrington: Would it make more sense to add an if check to the next line instead of making a temp file? If we didn't run any subjobs then there is not really anything to verify.

app/web_framework/cluster_master_application.py
```diff
@@ -41,7 +42,8 @@ def __init__(self, cluster_master):
                     ]),
                     RouteNode(r'queue', _QueueHandler),
                     RouteNode(r'slave', _SlavesHandler, 'slaves').add_children([
-                        RouteNode(r'(\d+)', _SlaveHandler, 'slave')
+                        RouteNode(r'(\d+)', _SlaveHandler, 'slave'),
```
tjlee0909: You probably have already considered this, but why don't we have a shutdown endpoint nested within the slave id (e.g. v1/slave/23/shutdown) with no body to shut down an individual slave?

app/web_framework/cluster_master_application.py
tjlee0909: 404? Why wouldn't it 500?

app/slave/cluster_slave.py
```diff
@@ -173,18 +175,20 @@ def _do_build_teardown_and_reset(self, timeout=None):
         for executor in self.executors_by_id.values():
             executor.kill()
 
-        if not self._project_type:
-            return  # There is no build to tear down! (Slave is idle.)
+        # This lock keeps multiple teardown attempts from racing each other and throwing exceptions.
+        with self._build_teardown_lock:
```
tjlee0909: This feels somehow redundant with the self._build_teardown_coin. @josephharrington ?

app/slave/cluster_slave.py
```diff
@@ -195,7 +199,7 @@ def _send_master_idle_notification(self):
         self._logger.info('Notifying master that this slave is ready for new builds.')
         self._notify_master_of_state_change(SlaveState.IDLE)
 
-    def _disconnect_from_master(self):
+    def disconnect_from_master(self):
```
tjlee0909: I don't see this being publicly used anywhere, why the private->public change?

app/master/cluster_master.py
```diff
@@ -151,8 +161,7 @@ def _disconnect_slave(self, slave):
         """
         # Mark slave dead. We do not remove it from the list of all slaves. We also do not remove it from idle_slaves;
         # that will happen during slave allocation.
-        slave.set_is_alive(False)
-        slave.current_build_id = None
+        slave.disconnect()
```
tjlee0909: Do you think we should maybe do a slave.kill() instead of just a slave.disconnect()?

app/master/build.py
tjlee0909: Yeah, this is kind of confusing, because the slave executor is claimed (incremented) in the calling function, begin_subjob_executions_on_slave().

app/web_framework/cluster_master_application.py
josephharrington: @tjlee0909 are you suggesting that it *should* return a 500? My point was that currently as it is written, the code would return a 404 since the `self._cluster_master.get_slave(slave_id)` call would raise an `ItemNotFoundError` which will be translated to a 404. My point was that when that exception is raised, we'll already have shut down some slaves, so it's a bit surprising.

The alternative would be to build up the list of slaves first (before calling `handle_slave_state_update(slave, SlaveState.SHUTDOWN)` on any of them). While we're checking this, maybe we'd also want to convert that 404 to something else (maybe just a 400?). Then once we're sure all the provided IDs are valid slaves, that's when iterate through them and tell each to shut down.

tjlee0909: @josephharrington nope, I was saying that I thought the current behavior was to 500, but I didn't know that an ItemNotFoundError exception translated into a 404.

tjlee0909: I just read "ItemNotFoundError" outloud, duh, that's a 404.

app/slave/cluster_slave.py
josephharrington: @tjlee0909 I don't think it's completely redundant -- I think the coin was meant as more of an assertion on behavior than as a concurrency control mechanism. But I think you're right that if we are adding a stronger control like a lock, we probably won't need the weaker check provided by the coin.

@drobertduke, under what conditions were you encountering the exception, and was it the `BuildTeardownError` from the coin or was it something else?

At some point (when we added the coin) there was clearly an assumption that we were not going to call this more than once for a single build. The coin was essentially an assertion for that assumption. It doesn't seem like that assumption has changed, but it seems like we now need extra concurrency control around this logic and so maybe the coin will be obsolete.

I don't disagree that a lock might be necessary (depending on which exceptions you were seeing) but one alternate approach to think about is only locking around setting `self._project_type` to `None`. That way other threads would not have to wait for teardown to complete:
```python
with self._build_teardown_lock:
    if not self._project_type:
        return
    project_type_to_teardown = self._project_type
    self._project_type = None

project_type_to_teardown.teardown_build(timeout=timeout)
...
```

app/master/cluster_master.py
josephharrington: I don't think so -- disconnect is initiated by the slave so the slave should already be shutting itself down.

josephharrington: Although we might want to rename the method on the Slave object so it's clear that this is not actually sending the slave any requests or telling the slave to do anything. Maybe something like `mark_disconnected`?

tjlee0909: Ah, yes, that confused me. So the _disconnect_slave() method in this class simply updates the master's bookeaping on that particular slave.

app/client/cluster_api_client.py
drobertduke: Good point, I updated the comment.

app/client/cluster_api_client.py
drobertduke: Added docs.  This is currently only used for testing but I'd like to add a client command to list slaves later.

app/client/cluster_api_client.py
drobertduke: :smile:

app/master/build.py
drobertduke: If frees the slave executor each time, it may optionally teardown.  Not sure if I should include that in the function name.

app/master/build.py
drobertduke: Renamed the Error and added a comment.

app/master/cluster_master.py
drobertduke: I renamed all the uses of 'shutdown' as a verb to 'graceful_shutdown'

app/master/cluster_master.py
drobertduke: I like the idea of removing it from _all_slaves_by_url, but I'm not sure where to do the removal.  If we do it here, the slave can still be executing a build but the master will no longer know about it.

slave.mark_as_idle() is where we check if the slave is in shutdown mode and 'kill' it if so, but from the slave object there's no way to remove it from the master's _all_slaves_by_url.  Still thinking about this.

app/master/cluster_master.py
drobertduke: renamed it to mark_dead

app/master/slave.py
drobertduke: renamed to is_in_shutdown_mode

app/master/slave.py
drobertduke: done

app/master/slave.py
drobertduke: whoops good catch, copy/paste mistake

app/slave/cluster_slave.py
drobertduke: In my functional tests, I start a build and then tell slaves to shutdown.  There's a scenario where the build completes and a slave starts running teardown(), then it receives the shutdown command, tries running teardown() again, and throws a big error.  In this scenario I don't want an error: as long as teardown only runs once, then everything is working correctly.  The lock allows the second teardown() call to exit 'properly' because `if not self._project_type` is True.  

Some alternatives: 
1) Remove the lock and don't raise an error if the coin is already spent.
```
if not self._build_teardown_count.spend() or not self._project_type: #  Order matters!!
  return
```
2) Remove the lock and wrap every _do_build_teardown_and_reset with a try/catch and ignore the BuildTeardownError.

3) Keep the lock and remove the build_teardown_coin.

I'm leaning toward 1) I think

app/slave/cluster_slave.py
drobertduke: it's used in test_cluster_slave, I guess that's why I changed it.  I'll change it back

app/web_framework/cluster_master_application.py
drobertduke: Good thinking.  I'm adding this route and moving some of the common shutdown logic to ClusterMaster

app/slave/cluster_slave.py
tjlee0909: I think the original intent of the build_teardown_coin was to signal that this code has no legitimate scenario where it can be executed multiple times for the same build. But now we have just introduced a legitimate scenario for it.

I'd say go with 1). as well. We definitely don't need both though.

app/web_framework/cluster_master_application.py
drobertduke: Good catch, fixed.

app/web_framework/cluster_slave_application.py
drobertduke: thanks!

test/functional/job_configs.py
drobertduke: If there are no subjob_file_*.txt files, the ls command will exit non-zero and the whole build fails immediately.  Not what I expected from ls.

app/master/cluster_master.py
josephharrington: I'm fine with punting on this for now. You're right -- it sounds trickier than I thought originally and is really just a nice to have that we can add later.

PR #119, opened 2015-04-12
Fix command line help for frozen app

josephharrington:
Our dynamically generated help strings for the build subcommand were
not being correctly displayed in the frozen application. The reason is
that the help strings are generated from docstrings which are removed
during freeze.


boxcla:

tjlee0909:
K, I'm assuming you know what you're doing here. :+1:

PR #120, opened 2015-04-12
Add atom timings to API response

josephharrington:
- Create a new Atom class that takes the place of our existing atoms
which were just strings
- Include both expected and actual time as fields in the Atom object,
and include these in the subjob API response.

### DIFF
--- a/app/master/time_based_atom_grouper.py
+++ b/app/master/time_based_atom_grouper.py
@@ -1,5 +1,4 @@
 from collections import OrderedDict
-from operator import itemgetter
 
 from app.master.atom_grouper import AtomGrouper
 
@@ -57,7 +56,7 @@ class TimeBasedAtomGrouper(object):
     def __init__(self, atoms, max_executors, atom_time_map, project_directory):
         """
         :param atoms: the list of atoms for this build
-        :type atoms: list[str]
+        :type atoms: list[app.master.atom.Atom]
         :param max_executors: the maximum number of executors for this build
         :type max_executors: int
         :param atom_time_map: a dictionary containing the historic times for atoms for this particular job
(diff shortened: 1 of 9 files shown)


boxcla:

app/master/time_based_atom_grouper.py
```diff
@@ -103,28 +103,27 @@ def groupings(self):
     def _coalesce_new_atoms_with_historic_times(self, new_atoms, old_atoms_with_times, project_directory):
```
tjlee0909: I remember naming this method. I think it wasn't a great name then, and it's an even worse name now with this change.

Maybe we can rename it to _set_expected_times()?

app/master/time_based_atom_grouper.py
```diff
@@ -103,28 +103,27 @@ def groupings(self):
...
         """
         Combine the historic timing data from previous atoms (for the same job) with the new atoms generated for this
-        job.
+        job. This will update the expected_time fields for the atom objects in new_atoms.
```
tjlee0909: "Set the expected run time (new_atom.expected_time)  of each atom in new_atoms using historic timing data."

tjlee0909:
Wow, I didn't think this change would involve any significant work. Good thing you nipped it in the bud before it got out of hand.

:+1:

tjlee0909:
Btw, what do you plan on doing now with this added functionality of being able to view expected vs actual atom timing data. More dashboard stuff?

josephharrington:
@tjlee0909 Yeah, maybe some dashboard stuff eventually. We could also just run one-off reports to get a sense of how much variability we are seeing in atom times.

app/master/time_based_atom_grouper.py
josephharrington: Will do.

app/master/time_based_atom_grouper.py
josephharrington: Yeah, nice. Will do.

app/master/time_based_atom_grouper.py
```diff
@@ -74,18 +73,19 @@ def groupings(self):
...
-                = self._coalesce_new_atoms_with_historic_times(self._atoms, self._atom_time_map, self._project_directory)
+            total_estimated_runtime = self._coalesce_new_atoms_with_historic_times(
+                self._atoms, self._atom_time_map, self._project_directory)
         except _AtomTimingDataError:
             grouper = AtomGrouper(self._atoms, self._max_executors)
             return grouper.groupings()
 
         # 2). Sort them by time, and add them to an OrderedDict
-        sorted_atom_times_left = OrderedDict(sorted(new_atoms_with_times_list, key=itemgetter(1), reverse=True))
+        atoms_sorted_by_desc_time = sorted(self._atoms, key=lambda atom: atom.expected_time, reverse=True)
```
wjdhollow: desc_time?

app/master/time_based_atom_grouper.py
wjdhollow: Ha. I was actually going to make the same comment.

wjdhollow:
:+1: w/ minor comment

app/master/time_based_atom_grouper.py
josephharrington: Yeah "descending" -- I'll clarify the name.

PR #129, opened 2015-04-29
Support empty commands in YAML (fixes #122)

### LINKED ISSUE #122: Lines with only comments in clusterrunner.yaml cause build setup errors
Florian ran into an issue where the clusterrunner master failed to prepare a build because it couldn't parse this:

setup_build: 	
        - ln -s etherpad-lite/src/ asset-pipeline/node_modules/ep_etherpad-lite
        - sudo npm install -g node-inspector	
        - #bin/installDependencies.sh 

Here is the stack trace from the clusterrunner_master.log:

[2015-04-13 18:23:53.171] ERROR   Bld928-Preparat build           Build 928 failed: 'NoneType' object has no attribute 'strip'
[2015-04-13 18:23:53.172] ERROR   Bld928-Preparat build_request_h Could not handle build request for build 928.
  File "/home/jenkins/ClusterRunnerBuild/app/master/build_request_handler.py", line 103, in _prepare_build_async
    self._prepare_build(build)
  File "/home/jenkins/ClusterRunnerBuild/app/master/build_request_handler.py", line 133, in _prepare_build
    job_config = project_type.job_config()
  File "/home/jenkins/ClusterRunnerBuild/app/project_type/project_type.py", line 50, in job_config
    return config.get_job_config(self._job_name)
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_runner_config.py", line 39, in get_job_config
    self._parse_raw_config()
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_runner_config.py", line 67, in _parse_raw_config
    self._set_config(config)
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_runner_config.py", line 122, in _set_config
    for job_name, job_values in config.items()}
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_runner_config.py", line 122, in <dictcomp>
    for job_name, job_values in config.items()}
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_runner_config.py", line 138, in _construct_job_config
    setup_build = self._shell_command_list_to_single_command(job_values.get(SETUP_BUILD))
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_runner_config.py", line 164, in _shell_command_list_to_single_command
    stripped_command = command.strip().rstrip(';')
AttributeError: 'NoneType' object has no attribute 'strip'

Once Florian removed the last line, it worked fine.

aptxkid:


boxcla:

josephharrington:
Minor thing: want to move the "fixes 122" in the commit title into the commit message? Github doesn't seem to linkify it if it's in the title.

app/master/cluster_runner_config.py
```diff
@@ -160,7 +160,7 @@ def _shell_command_list_to_single_command(self, commands):
         # We should join the commands with double ampersands UNLESS the command already ends with a single ampersand.
         # A semicolon (or a double ampersand) is invalid syntax after a single ampersand.
         sanitized_commands = []
-        for command in commands:
+        for command in [c for c in commands if c is not None]:
```
josephharrington: This is just kind of a style preference issue, but I wonder if this would be a bit more readable if we did the `None` check inside the existing `for` loop? I do like the list comprehension syntax but since we've already got a normal `for` loop here, it feels a bit weird to combine it with a comprehension.

test/unit/master/test_cluster_runner_config.py
```diff
@@ -77,22 +94,30 @@ class TestClusterRunnerConfig(BaseUnitTestCase):
...
+        }),
+        valid_config_with_empty_command=(_VALID_CONFIG_WITH_EMPTY_COMMAND, {
+            'command': 'echo "go" ',
+            'atomizer': [{'ENV_VAR': 'echo "atom"'}],
+        }),
     )
-    def test_all_conf_properties_are_correctly_parsed(self, conf_method_name, expected_value):
-        config = ClusterRunnerConfig(self._COMPLETE_VALID_CONFIG)
+    def test_valid_conf_properties_are_correctly_parsed(self, config_stream, expected_loaded_config):
+        config = ClusterRunnerConfig(config_stream)
```
josephharrington: Minor thing: "string" might be a more accurate name than "stream".

test/unit/master/test_cluster_runner_config.py
```diff
@@ -77,22 +94,30 @@ class TestClusterRunnerConfig(BaseUnitTestCase):
...
+            'name': 'Best Job Ever',
+            'max_executors': 21,
+            'setup_build': 'echo "This is setup! Woo!" && sleep 1 ',
+            'command': 'echo "Now I\'m doing $THE_THING!" && echo "Semicolons are fun." > /tmp/my_hard_work.txt ',
+            'atomizer': [{'THE_THING': 'printf \'something with a number %d\\n\' {1..50}'}],
+        }),
+        valid_config_with_empty_command=(_VALID_CONFIG_WITH_EMPTY_COMMAND, {
+            'command': 'echo "go" ',
+            'atomizer': [{'ENV_VAR': 'echo "atom"'}],
+        }),
```
josephharrington: I like this test modification!

app/master/cluster_runner_config.py
aptxkid: Absolutely. Will do.

test/unit/master/test_cluster_runner_config.py
```diff
@@ -55,6 +55,15 @@ class TestClusterRunnerConfig(BaseUnitTestCase):
     PHPUnit:
     """
 
+    _VALID_CONFIG_WITH_EMPTY_COMMAND = """
+    PHPUnit:
+        commands:
+            - echo "go"
+            -
```
josephharrington: Should we add a comment to the empty line (or add an additional line with a comment?)

An example of what I mean is:
```
commands:
    - echo "go"
    - # some dumb comment
    - 
```

The reason I suggest this is related to the specific use case listed in the issue ticket #122. It looks like that syntax for a comment gets interpreted as a yaml comment and ignored instead of being interpreted as a bash comment and included in the final command. (This is a good thing since I think a bash comment would break something! :smile:) But I feel like we'd want to make sure we keep that behavior if we ever change yaml libraries or something like that.

josephharrington:
One more minor comment but overall looks great! :+1:

test/unit/master/test_cluster_runner_config.py
aptxkid: You're absolutely right. I left it out earlier because I thought it's not necessary to (sort of) test PyYAML here. But you're totally right that we want to keep the behavior consistent even we choose not to use PyYAML in the future. Will add.

aptxkid:
@josephharrington I believe I have addressed all the comments.

PR #130, opened 2015-04-30
Fix logging same error twice in build_runner

### LINKED ISSUE #108: Client repeats some output when build fails
Example output (note error and cancel messages are printed twice):
```
Build is running. (Build id: 1)
Polling build status url: http://jharrington.local:43000/v1/build/1
0 of 4 subjobs are complete (0.0%).
1 of 4 subjobs are complete (25.0%).
Build aborted due to error: Error occurred while handling results of subjob 1.
Build aborted due to error: Error occurred while handling results of subjob 1.
Script aborted due to error!
Cancelling build 1
Executing teardown callback: <bound method BuildRunner._cancel_build of <app.client.build_runner.BuildRunner object at 0x109918710>>
Cancelling build 1
```

Also we probably don't need to print out the teardown callback info.

aptxkid:
This fixes #108

I added some unit tests in test_build_runner.py to prevent regression.
We usually don't write unit tests for logging messages, but logging
messages here are user facing so we want to keep them consistent.

### DIFF
--- a/test/unit/client/test_build_runner.py
+++ b/test/unit/client/test_build_runner.py
@@ -1,5 +1,5 @@
 from requests.models import Response
-from unittest.mock import MagicMock, Mock
+from unittest.mock import MagicMock, Mock, call
 
 from app.client.build_runner import BuildRunner
 from test.framework.base_unit_test_case import BaseUnitTestCase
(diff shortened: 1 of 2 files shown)


boxcla:

test/unit/client/test_build_runner.py
```diff
@@ -1,42 +1,94 @@
...
-        start_response = Response()
-        start_response.json = MagicMock(return_value={"build_id": "1"})
-        runner._network.post = MagicMock(return_value=start_response)
-        status_response = Response()
-        status_response.json = MagicMock(return_value=response)
-        runner._network.get = MagicMock(return_value=status_response)
-        runner._download_and_extract_results = MagicMock()
-        return runner
+    def _create_build_runner_instance(self, response):
+        with patch(build_runner.__name__ + '.get_logger') as mock_get_logger:
```
josephharrington: We have a patch helper in `BaseUnitTestCase` that I feel creates a cleaner syntax. So you can do:
```python
mock_get_logger = self.patch('app.client.build_runner.get_logger')
...
```

test/unit/client/test_build_runner.py
```diff
@@ -7,36 +7,87 @@
 
 class TestBuildRunner(BaseUnitTestCase):
 
-    def mock_runner_with_status_response(self, response):
+    def _create_build_runner_instance(self, response):
+        mock_get_logger = self.patch('app.client.build_runner.get_logger')
+        logger = MagicMock()
+        mock_get_logger.return_value = logger
```
josephharrington: I was curious about a very slight potential simplification, replacing lines 12 and 13 with:
```python
logger = mock_get_logger.return_value
```
I think accessing the return_value directly should return a Mock -- and depending on how it's used, you may or may not to replace it explicitly with a MagicMock. I think it's fine either way -- it was just something that I was thinking about. :wink:

josephharrington:
:+1: :+1: with a small optional comment.

PR #132, opened 2015-05-06
Return 202 when try to download artifact if it's not ready

### LINKED ISSUE #104: When artifact is not ready, trying to download causes a server error
```
[2015-03-12 11:56:34.507] ERROR   MasterTornadoTh tornado.applica Uncaught exception GET /v1/build/3958/artifacts.tar.gz (10.8.189.14)
HTTPRequest(protocol='http', host='pod4101-automation1024.pod.box.net:43000', method='GET', uri='/v1/build/3958/artifacts.tar.gz', version='HTTP/1.1', remote_ip='10.8.189.14', headers={'Accept-Encoding': 'gzip, deflate, sdch', 'Host': 'pod4101-automation1024.pod.box.net:43000', 'Connection': 'keep-alive', 'Cookie': 'box_visitor_id=5494751403d2b1.65401525; box_marketing=1; external_referrer=https%3A%2F%2Fwww.docusign.net%2FSigning%2F%3Fti%3D811315b7c87646e8ba94a3656fdc838a; _ga=GA1.2.1627628535.1413828457', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8', 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.76 Safari/537.36', 'Accept-Language': 'en-US,en;q=0.8'})
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/tornado/web.py", line 1346, in _when_complete
    callback()
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/site-packages/tornado/web.py", line 1367, in _execute_method
    self._when_complete(method(*self.path_args, **self.path_kwargs),
  File "/home/jenkins/ClusterRunnerBuild/app/web_framework/cluster_master_application.py", line 209, in get
    artifact_file_path = self._cluster_master.get_path_for_build_results_archive(int(build_id))
  File "/home/jenkins/ClusterRunnerBuild/app/master/cluster_master.py", line 268, in get_path_for_build_results_archive
    raise ItemNotReadyError('Build artifact file is not yet ready. Try again later.')
app.util.exceptions.ItemNotReadyError: Build artifact file is not yet ready. Try again later.
[2015-03-12 11:56:34.509] ERROR   MasterTornadoTh tornado.access  500 GET /v1/build/3958/artifacts.tar.gz (10.8.189.14) 2.25ms
```

aptxkid:
This fixes #104

### DIFF
--- a/app/web_framework/cluster_base_handler.py
+++ b/app/web_framework/cluster_base_handler.py
@@ -10,9 +10,13 @@ from app.util.network import ENCODED_BODY
 
 
 class ClusterBaseHandler(tornado.web.RequestHandler):
+    """
+    ClusterBaseHandler is the base handler for all request handlers of ClusterRunner services.
+    """
 
-    SUCCESS_STATUS = 'SUCCESS'
-    FAILURE_STATUS = 'FAILURE'
+    def __init__(self, *args, **kwargs):
+        self._logger = log.get_logger(__name__)
+        super().__init__(*args, **kwargs)
 
     _exception_status_codes = {
         ItemNotReadyError: http.client.ACCEPTED,
@@ -21,9 +25,28 @@ class ClusterBaseHandler(tornado.web.RequestHandler):
         ItemNotFoundError: http.client.NOT_FOUND,
     }
 
-    def __init__(self, *args, **kwargs):
-        self._logger = log.get_logger(__name__)
-        super().__init__(*args, **kwargs)
+    def _handle_request_exception(self, ex):
+        """
+        This is the "catch-all" exception handler that Tornado uses to map some exception types to appropriate HTTP
+        status codes. The default status code is 500 (generic server error).
+
+        :param ex: The exception that was caught
+        :type ex: Exception
+        """
+        # _handle_request_exception() is called in the exception handler, so we can still use logger.exception.
+        self._logger.exception('Exception occurred during request to {}.', self.request.uri)
+        status_code = self._exception_status_codes.get(type(ex), http.client.INTERNAL_SERVER_ERROR)
+        self.set_status(status_code)
+        self.finish()
+
+
+class ClusterBaseAPIHandler(ClusterBaseHandler):
+    """
+    ClusterBaseAPIHandler is the base handler for all API endpoints that are supposed to return json responses.
+    """
+
+    SUCCESS_STATUS = 'SUCCESS'
+    FAILURE_STATUS = 'FAILURE'
 
     def initialize(self, route_node=None):
         """
@@ -31,6 +54,16 @@ class ClusterBaseHandler(tornado.web.RequestHandler):
         :type route_node: RouteNode
         """
         self._route_node = route_node
+
+    def _handle_request_exception(self, ex):
+        """
+        Ensure exceptions are caught and a JSON response is always returned to the client.
+
+        :param ex: The exception that was caught
+        :type ex: Exception
+        """
+        self.write({'error': str(ex)})
+        super()._handle_request_exception(ex)
 
     def prepare(self):
         """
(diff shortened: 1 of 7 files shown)


boxcla:

app/web_framework/cluster_master_application.py
```diff
@@ -206,10 +207,14 @@ def initialize(self, route_node=None, cluster_master=None):
...
-        self.set_header('Content-Type', 'application/octet-stream')  # this should be downloaded as a binary file
-        return super().get(path=artifact_filename)
+        try:
+            artifact_file_path = self._cluster_master.get_path_for_build_results_archive(int(build_id))
+            self.root, artifact_filename = os.path.split(artifact_file_path)
+            self.set_header('Content-Type', 'application/octet-stream')  # this should be downloaded as a binary file
+            return super().get(path=artifact_filename)
+        except ItemNotReadyError as ex:
+            self.set_status(status_code=http.client.ACCEPTED, reason=str(ex))
+            return
```
josephharrington: Did you see the `_exception_status_codes` dict [in cluster_base_handler.py](https://github.com/box/ClusterRunner/blob/master/app/web_framework/cluster_base_handler.py#L17-L22). We could probably just add an entry to that?

josephharrington: Actually, I just noticed that `ItemNotReadyError` is already in that dict! It looks like that's not working as expected. Ah, it's because `_BuildResultHandler` does not inherit from `ClusterBaseHandler`.

Hmm, this is probably fine then -- although I wonder if we should add a comment explaining why the exception needs to be caught here. (I think I've been confused by this before. :wink:)

aptxkid: We probably should not catch exceptions for requests by overriding the private method `_handle_request_exception`. I am thinking about a better way to do it. And we can make the handler cover StaticFileHandler as well.

app/web_framework/cluster_base_handler.py
```diff
@@ -21,9 +22,29 @@ class ClusterBaseHandler(tornado.web.RequestHandler):
         ItemNotFoundError: http.client.NOT_FOUND,
     }
 
-    def __init__(self, *args, **kwargs):
-        self._logger = log.get_logger(__name__)
-        super().__init__(*args, **kwargs)
+    def _handle_request_exception(self, ex):
```
aptxkid: I plan to make another PR to move this logic to either write_error or send_error.

app/web_framework/cluster_base_handler.py
```diff
@@ -21,9 +22,29 @@ class ClusterBaseHandler(tornado.web.RequestHandler):
...
+        self._logger.exception('Exception occurred during request to {}.', self.request.uri)
+        status_code = self._exception_status_codes.get(type(ex), http.client.INTERNAL_SERVER_ERROR)
+        response = {'error': str(ex)}
+
+        self.set_status(status_code)
+        self.write(response)
+        self.finish()
+
+
+class ClusterBaseRequestHandler(ClusterBaseHandler):
```
josephharrington: What is the difference between `ClusterBaseHandler` and `ClusterBaseRequestHandler`? I wonder if we can find better naming (or at least add class docs) to more clearly signify what should go in which class. They both handle requests, so it feels like those names are conceptually identical.

It looks like `ClusterBaseRequestHandler` is just for JSON API requests and `ClusterBaseHandler` is for any request, so what do you think of renaming `ClusterBaseRequestHandler` to maybe something like `ClusterBaseAPIHandler` or `ClusterBaseJsonHandler`?

app/web_framework/cluster_master_application.py
```diff
@@ -60,11 +60,27 @@ def initialize(self, route_node=None, cluster_master=None):
         super().initialize(route_node)
 
 
-class _RootHandler(_ClusterMasterBaseHandler):
+class _ClusterMasterBaseStaticFileHandler(ClusterBaseHandler, tornado.web.StaticFileHandler):
```
josephharrington: I like the multiple inheritance approach so that all the handlers can share exception handling. How do you feel about keeping this a part of `_BuildResultHandler` for now?

I suggest that just because that single handler is the only route uses this pattern, so it's a bit hard to guess what abstraction would be useful if we ever had another static file handler. And also I think the inheritance chain of all these handlers has gotten complicated enough that it might be easier to comprehend (for now) if we just kept all this inside `_BuildResultHandler` like 
```python
class _BuildResultHandler(ClusterBaseHandler, tornado.web.StaticFileHandler):
...
```

What do you think?

josephharrington: My original comment was maybe not very clear -- I was suggesting getting rid of `_ClusterMasterBaseStaticFileHandler` entirely and just making `_BuildResultHandler` do the multiple inheritance? What do you think of that?

aptxkid: Your original comment was very clear. I have no idea why I missed it. :disappointed:

aptxkid: And also, it makes perfect sense to do what you suggested.

app/web_framework/cluster_base_handler.py
```diff
@@ -32,6 +55,16 @@ def initialize(self, route_node=None):
...
 
+    def _handle_request_exception(self, ex):
+        """
+        Ensure exceptions are caught and a JSON response is always returned to the client.
+
+        :param ex: The exception that was caught
+        :type ex: Exception
+        """
+        self.write({'error': str(ex)})
+        super()._handle_request_exception(ex)
```
josephharrington: Nice.

josephharrington:
:+1: Looks great!

PR #138, opened 2015-05-11
Validate deployment by checking each slave is connected

### LINKED ISSUE #137: Validate successful deployment by checking each slave is connected
Right now, validation simply check the number of slaves connected.

``` python
def all_slaves_registered():
    return len(self._registered_slave_hostnames(slave_api_url, network)) == len(slaves_to_validate)
```

Of course, this can cause problems. When I was trying to deploy clusterrunner on two Mac VMs, the deployment actually failed but the validation passed due to exactly this problem.

aptxkid:
Right now, validation simply check the number of slaves connected.

``` python
def all_slaves_registered():
      return len(self._registered_slave_hostnames(slave_api_url,
            network)) == len(slaves_to_validate)
```

Of course, this can cause problems. When I was trying to deploy
clusterrunner on two Mac VMs, the deployment actually failed but
the validation passed due to exactly this problem.

This fixes #137

### DIFF
--- a/app/subcommands/deploy_subcommand.py
+++ b/app/subcommands/deploy_subcommand.py
@@ -201,7 +201,13 @@ class DeploySubcommand(Subcommand):
         network = Network()
 
         def all_slaves_registered():
-            return len(self._registered_slave_hostnames(slave_api_url, network)) == len(slaves_to_validate)
+            registered_slave_uids = set(
+                [Network.rsa_key(x) for x in self._registered_slave_hostnames(slave_api_url, network)]
+            )
+            slaves_to_validate_uids = set(
+                [Network.rsa_key(x) for x in slaves_to_validate]
+            )
+            return registered_slave_uids == slaves_to_validate_uids
 
         if not wait_for(
                 boolean_predicate=all_slaves_registered,
--- a/test/unit/subcommands/test_deploy_subcommand.py
+++ b/test/unit/subcommands/test_deploy_subcommand.py
@@ -1,8 +1,15 @@
+from functools import partial
+from unittest.mock import MagicMock
+import uuid
+
+from genty import genty, genty_dataset, genty_args
+
 from app.subcommands.deploy_subcommand import DeploySubcommand
 from app.util.network import Network
 from test.framework.base_unit_test_case import BaseUnitTestCase
 
 
+@genty
 class TestDeploySubcommand(BaseUnitTestCase):
     def setUp(self):
         super().setUp()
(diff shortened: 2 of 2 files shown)


boxcla:

app/subcommands/deploy_subcommand.py
```diff
@@ -201,7 +201,9 @@ def _validate_successful_deployment(self, master_service_url, slaves_to_validate
         network = Network()
 
         def all_slaves_registered():
-            return len(self._registered_slave_hostnames(slave_api_url, network)) == len(slaves_to_validate)
+            registered_slave_uids = set(map(Network.rsa_key, self._registered_slave_hostnames(slave_api_url, network)))
```
aptxkid: Using public rsa key to identify a host is not reliable since there's nothing preventing two different hosts to have the same rsa key pair. I am trying to use ip or MAC address here. But it's not relevant to this change. I will make another PR to replace `Network.rsa_key`.

pylintrc
```diff
@@ -40,7 +40,7 @@ load-plugins=
 
 disable=R, missing-docstring, fixme, invalid-name, bad-continuation, attribute-defined-outside-init, arguments-differ,
     attribute-defined-outside-init, star-args, locally-disabled, unused-argument, unused-variable, no-init,
-    global-statement,
+    global-statement, bad-builtin
```
wjdhollow: It looks like the use of `map` is discouraged in Python 3. Is there a reason you are not using a list comprehension instead?

http://www.artima.com/weblogs/viewpost.jsp?thread=98196
https://docs.python.org/3.1/whatsnew/3.0.html#views-and-iterators-instead-of-lists

wjdhollow:
:ok_hand: + small comment

test/unit/subcommands/test_deploy_subcommand.py
```diff
@@ -140,3 +147,66 @@ def rsa_key(*args, **kwargs):
         deploy_subcommand = DeploySubcommand()
         non_registered = deploy_subcommand._non_registered_slaves(registered_hosts, slaves_to_validate)
         self.assertEquals(0, len(non_registered))
+
+    @genty_dataset(
+        (
```
josephharrington: What do you think about using the named datasets here (with keyword args to genty_dataset)? I feel like it definitely makes the test names more readable during the test run, and usually also makes it easier to comprehend what each set of test data is doing.

josephharrington:
Minor suggestion on the tests, but looks good overall :+1:.

PR #142, opened 2015-05-12
Use IP addr to identify hosts

### LINKED ISSUE #139: Identify host using IP address
Right now we use rsa public key to identify each host, which is not reliable since there's nothing preventing two hosts to have the same rsa key pair. We should use the IP address to identify hosts.

aptxkid:
This fixes #139

### DIFF
--- a/app/subcommands/deploy_subcommand.py
+++ b/app/subcommands/deploy_subcommand.py
@@ -202,10 +202,10 @@ class DeploySubcommand(Subcommand):
 
         def all_slaves_registered():
             registered_slave_uids = set(
-                [Network.rsa_key(x) for x in self._registered_slave_hostnames(slave_api_url, network)]
+                [Network.get_host_id(x) for x in self._registered_slave_hostnames(slave_api_url, network)]
             )
             slaves_to_validate_uids = set(
-                [Network.rsa_key(x) for x in slaves_to_validate]
+                [Network.get_host_id(x) for x in slaves_to_validate]
             )
             return registered_slave_uids == slaves_to_validate_uids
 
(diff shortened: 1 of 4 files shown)


boxcla:

aptxkid:
Just realized that IP `localhost` will actually not match the IP of the hostname. So if someone is running `clusterrunner deploy -m bla.com` on bla.com, it's going to have a problem.

aptxkid:
I solved the `localhost` problem by mapping it to `socket.gethostname()`

app/subcommands/deploy_subcommand.py
```diff
@@ -267,21 +267,21 @@ def _non_registered_slaves(self, registered_slaves, slaves_to_validate):
         :return: list of slave hostnames that haven't registered with the master service yet
         :rtype: list[str]
         """
-        registered_rsa_keys = []
+        registered_host_ids = []
 
         for registered_slave in registered_slaves:
-            registered_rsa_keys.append(Network.rsa_key(registered_slave))
+            registered_host_ids.append(Network.get_host_id(registered_slave))
```
wjdhollow: Just curious, why for loops instead of list comprehensions? Is it a matter of line length?

app/subcommands/deploy_subcommand.py
```diff
@@ -267,21 +267,21 @@ def _non_registered_slaves(self, registered_slaves, slaves_to_validate):
...
 
-        slaves_to_validate_rsa_key_host_pairs = {}
+        slaves_to_validate_host_id_pairs = {}
 
         for slave_to_validate in slaves_to_validate:
-            slaves_to_validate_rsa_key_host_pairs[Network.rsa_key(slave_to_validate)] = slave_to_validate
+            slaves_to_validate_host_id_pairs[Network.get_host_id(slave_to_validate)] = slave_to_validate
```
wjdhollow: Alternatively, you can use the dictionary list comprehensions here if you want :smile:

app/subcommands/deploy_subcommand.py
aptxkid: No. You're totally right. It was doing a little more. But now it makes perfect sense to use list comprehensions. I will update it. :smiley:

app/subcommands/deploy_subcommand.py
```diff
@@ -267,21 +267,21 @@ def _non_registered_slaves(self, registered_slaves, slaves_to_validate):
...
 
         non_registered_slave_hosts = []
 
-        for rsa_key in slaves_to_validate_rsa_key_host_pairs:
-            if rsa_key not in registered_rsa_keys:
-                non_registered_slave_hosts.append(slaves_to_validate_rsa_key_host_pairs[rsa_key])
+        for host_id in slaves_to_validate_host_id_pairs:
```
wjdhollow: I'm kind of beating a dead horse, but I think this sort of condition checking is more readable with a list comprehension. Maybe change the variables names so it can fit on one line.

app/util/network.py
```diff
@@ -138,45 +136,41 @@ def are_hosts_same(host_a, host_b):
...
-        :return: the rsa key string, without the 'ssh-rsa' prefix. Returns None if failed ssh-keyscan fails.
+        :return: the id of the host
         :rtype: str|None
         """
-        proc = subprocess.Popen('ssh-keyscan -t rsa {}'.format(host), shell=True, stdout=PIPE, stderr=PIPE)
-        output, error = proc.communicate()
-
-        if proc.returncode != 0:
-            log.get_logger(__name__).error('Failed to get rsa string with output: {}, error: {}'.format(output, error))
+        try:
```
wjdhollow: I'd prefer throwing an error rather than returning None. That way you can handle the error case with a simple try block. Maybe just annotate that it throws a socket.gaierror and handle it in are_host_same

wjdhollow:
:ok_hand: small comments (borderline nitpicky), but nothing blocking.

app/subcommands/deploy_subcommand.py
aptxkid: These are just renaming variables to get rid of references to 'rsa_key'. But you are totally right.

app/util/network.py
aptxkid: I try to avoid doing too much in one PR. The contract was returning None for failures, which might be improved but not in this PR, IMO.

PR #146, opened 2015-05-13
Translate 'localhost' to hostname when asking the slave to connect to master

### LINKED ISSUE #144: slave trying to connect to `localhost:43000`
Deploy by running:
```bash
clusterrunner deploy -v -s bar.com
```
on foo.com

Cluster slave log on bar.com:
requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=43000) ... [Errno 61]

aptxkid:
This fixes #144

### DIFF
--- a/app/deployment/remote_slave_service.py
+++ b/app/deployment/remote_slave_service.py
@@ -1,3 +1,5 @@
+import socket
+
 from app.deployment.remote_service import RemoteService
 
 
@@ -20,6 +22,8 @@ class RemoteSlaveService(RemoteService):
         :param num_executors: the number of concurrent executors that will run in this slave service
         :type num_executors: int
         """
+        if master_host == 'localhost':
+            master_host = socket.gethostname()
         slave_args = '--master-url {}:{}'.format(master_host, str(master_port))
         slave_args += ' --port {}'.format(str(slave_port))
         slave_args += ' --num-executors {}'.format(str(num_executors))
--- a/test/unit/deployment/test_remote_slave_service.py
+++ b/test/unit/deployment/test_remote_slave_service.py
@@ -1,0 +1,44 @@
+import socket
+from unittest.mock import Mock
+
+from genty import genty, genty_dataset
+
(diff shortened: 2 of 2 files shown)


boxcla:

app/deployment/remote_slave_service.py
```diff
@@ -20,6 +22,8 @@ def start(self, master_host, master_port, slave_port, num_executors):
         :param num_executors: the number of concurrent executors that will run in this slave service
         :type num_executors: int
         """
+        if master_host == 'localhost':
+            master_host = socket.gethostname()
```
josephharrington: Ah, nice. I wonder if the 'localhost' ambiguity could bite us anywhere else / if it would be useful to do this translation at a higher level in the app. (I wouldn't expect that in this PR but I am just thinking out loud. :smile:)

josephharrington:
Looks pretty good, Lu. Can you add more detail to the commit message and/or issue about what the root cause of the bug was and how this change fixes the issue? (I think I understand what was happening, but  we might want to note the specific use case that caused the issue -- it seems like this wouldn't have been mac-specific after all?)

app/deployment/remote_slave_service.py
tjlee0909: Totes. Totes.

tjlee0909:
:+1:

app/deployment/remote_slave_service.py
aptxkid: @josephharrington I thought about this as well. My original thought was to get rid of `localhost` completely, which is probably not going to happen since you want to have the capability to have `localhost` in the config yaml file to reference to the local host. And I find deployment the only place that this translation is only needed (I could totally be wrong). But anyway, if we find that there are some places else need to do this translation, we can think about how to do it better to keep it DRY. Does it sound good?

aptxkid:
@josephharrington I've removed the `mac` label of the issue. I will add more details to the commit message.

josephharrington:
:+1:

Issue #141, opened 2015-05-12
Logic in BaseConfigLoader for detecting bool conf values is incorrect


josephharrington:
If a config setting has a boolean value and is overridden in a config file, it will cause an exception on app startup:

```
Traceback (most recent call last):
  File "/Users/jharrington/code/ClusterRunner/main.py", line 274, in <module>
    main(sys.argv[1:])
  File "/Users/jharrington/code/ClusterRunner/main.py", line 259, in main
    _initialize_configuration(parsed_args.pop('subcommand'), parsed_args.pop('config_file'))
  File "/Users/jharrington/code/ClusterRunner/main.py", line 212, in _initialize_configuration
    conf_loader.load_from_config_file(config, config_filename)
  File "/Users/jharrington/code/ClusterRunner/app/util/conf/base_config_loader.py", line 107, in load_from_config_file
    self._load_section_from_config_file(config, config_filename, self.CONFIG_FILE_SECTION)
  File "/Users/jharrington/code/ClusterRunner/app/util/conf/base_config_loader.py", line 154, in _load_section_from_config_file
    self._cast_and_set(key, value, config)
  File "/Users/jharrington/code/ClusterRunner/app/util/conf/base_config_loader.py", line 164, in _cast_and_set
    config.set(key, int(value))
ValueError: invalid literal for int() with base 10: 'False'
```

The bug was introduced with [this change](https://github.com/box/ClusterRunner/commit/55fef394df0053a9bc5e3dfc910914699807ab56#diff-6e4b0b5265e2332751166695c401ae3aR159) that replaced `type(x) is y` checks with `isinstance(x, y)`. The problem here is that `bool` is actually a subclass of `int`, so adding the `isinstance` checks actually changed the behavior of the code.

josephharrington:
Fixed in #149.

PR #149, opened 2015-05-13
Fix bug with conf parsing

josephharrington:
Fixes issue #141 which would cause an error if a config file tried to
set a boolean config value.

### DIFF
--- a/app/util/conf/base_config_loader.py
+++ b/app/util/conf/base_config_loader.py
@@ -103,16 +103,14 @@ class BaseConfigLoader(object):
         if self.CONFIG_FILE_SECTION:
             self._load_section_from_config_file(config, config_filename, self.CONFIG_FILE_SECTION)
 
-    def _load_section_from_config_file(self, config, config_filename, section):
+    def _get_config_file_whitelisted_keys(self):
         """
-        Load a config file and copy all the values in a particular section to the Configuration singleton
-        :type config: Configuration
-        :type config_filename: str
-        :type section: str
+        Return the list of keys that we allow to be specified in a config file. Subclasses can override this method but
+        should in general append values to the list returned by their superclass.
+
+        :rtype: list[str]
         """
-        # Only keys from this list will be loaded from a conf file.  If the conf file contains other keys we will
-        # error to alert the user.
-        config_key_validation = [
+        return [
             'secret',
             'base_directory',
             'log_level',
@@ -129,6 +127,14 @@ class BaseConfigLoader(object):
             'git_strict_host_key_checking',
             'cors_allowed_origins_regex',
         ]
+
+    def _load_section_from_config_file(self, config, config_filename, section):
+        """
+        Load a config file and copy all the values in a particular section to the Configuration singleton
+        :type config: Configuration
+        :type config_filename: str
+        :type section: str
+        """
         try:
             config_parsed = ConfigFile(config_filename).read_config_from_disk()
         except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


boxcla:

app/util/conf/base_config_loader.py
```diff
@@ -156,19 +164,23 @@ def _cast_and_set(self, key, value, config):
...
         default_value = config.get(key)
-        if isinstance(default_value, int):
+
+        if isinstance(default_value, bool):
+            value_mapping = {'true': True, 'false': False}
+            if not isinstance(value, str) and not value.lower() in value_mapping.keys():
+                raise InvalidConfigError('The value for {} should be True or False, but it is "{}"'.format(key, value))
+            config.set(key, value_mapping[value.lower()])
+
+        elif isinstance(default_value, int):
```
josephharrington: This is the actual fix -- just switched the order of the type checks so that `bool` is checked before `int`.

test/unit/util/conf/test_base_config_loader.py
```diff
@@ -1,10 +1,62 @@
...
+        conf.set('some_list', ['red', 'white', 'blue'])
+        conf.set('some_str', 'America!')
+        conf.set('some_nonwhitelisted_key', 1492)
+
+
+@genty
 class TestBaseConfigLoader(BaseUnitTestCase):
 
+    @genty_dataset(
+        bool_type=('some_bool', 'False', False),
```
josephharrington: The dataset for bool_type is the test that would have failed before the above fix.

app/util/conf/base_config_loader.py
```diff
@@ -129,6 +127,14 @@ def _load_section_from_config_file(self, config, config_filename, section):
             'git_strict_host_key_checking',
             'cors_allowed_origins_regex',
         ]
+
```
tjlee0909: Stupid git diff'ing.

app/util/conf/base_config_loader.py
```diff
@@ -139,12 +145,14 @@ def _load_section_from_config_file(self, config, config_filename, section):
...
+            raise InvalidConfigError('The config file {} does not contain a [{}] section'
+                                     .format(config_filename, section))
+
         clusterrunner_config = config_parsed[section]
+        whitelisted_file_keys = self._get_config_file_whitelisted_keys()
         for key in clusterrunner_config:
-            if key not in config_key_validation:
-                raise _InvalidConfigError('The config file contains an invalid key: {}'.format(key))
+            if key not in whitelisted_file_keys:
+                raise InvalidConfigError('The config file contains an invalid key: {}'.format(key))
```
tjlee0909: I don't want to bike-shed, as you haven't made anything worse--but we should probably gather ALL of the non-whitelisted keys and raise an error with them instead of just crapping out on the first one.

app/util/conf/base_config_loader.py
tjlee0909: Could you also add a comment to the top of the 'if' statement indicating that this particular clause HAS to be evaluated first?

tjlee0909:
:+1: LGTM with minor comments.

app/util/conf/base_config_loader.py
josephharrington: I thought about adding a comment, but decided against it in favor of adding tests that will fail now if the logic is wrong.

app/util/conf/base_config_loader.py
josephharrington: I probably agree with that. Feel free to put it in a new enhancement ticket if you wanna. :wink:

app/util/conf/base_config_loader.py
tjlee0909: Why not both? If someone were to perform a refactoring that somehow shifts the order of operations (or doesn't guarantee a certain linear ordering for example--I know, a stretch), then it'd be near their first iteration of the development cycle that they would discover the problem. Where-as a one-line comment would let them know immediately.

josephharrington: Done.

app/util/conf/base_config_loader.py
tjlee0909: Will do.

PR #153, opened 2015-06-05
Only register signal handler for SIGINFO if the platform supports it

### LINKED ISSUE #152: Don't register SIGINFO signal handlers on platforms do not support it
According to Python documentation, https://docs.python.org/3.4/library/signal.html#signal.signal, on Windows, register signal handler for SIGINFO will raise ValueError.

aptxkid:
According to Python documentation, https://docs.python.org/3.4/library/signal.html#signal.signal, on Windows, register signal handler for SIGINFO will raise ValueError. Instead of crashing, we should just swallow the exception and log a warning.

This fixes #152


boxcla:

josephharrington:
:+1:

PR #156, opened 2015-06-08
Suppress `AttributeError` while blacklisting methods in base_unit_tes…

### LINKED ISSUE #155: Certain methods in the `blacklisted_methods` are not available on all the platforms
Certain methods in the `blacklisted_methods` in `base_unit_test_case` are not available on all the platforms. E.g. `os.chown` is not available on Windows. This can cause all the unit test to fail on Windows as patching `os.chown` will raise `AttributeError`.

aptxkid:
…t_case

Some of the methods being blacklisted are not available on all the platforms. E.g. `os.chown` is not available on Windows. Patching it will raise `AttributeError` and cause all unit tests to fail on Windows.

This fixes #155

### DIFF
--- a/test/framework/base_unit_test_case.py
+++ b/test/framework/base_unit_test_case.py
@@ -224,7 +224,9 @@ class BaseUnitTestCase(TestCase):
         for disabled_reason, patch_targets in blacklisted_methods.items():
             for patch_target in patch_targets:
                 # Suppress UnitTestPatchError, which happens if target has already been patched (no safeguard needed).
-                with suppress(UnitTestPatchError):
+                # Suppress AttributeError, which happens if trying to patch a target that is not available. (e.g.
+                # os.chown on Windows)
+                with suppress(UnitTestPatchError, AttributeError):
                     self._blackist_target(patch_target, disabled_reason)
 
     def _blackist_target(self, patch_target, disabled_reason):


boxcla:

aptxkid:
I am not a crazy about suppressing `AttributeError` here as we might have a typo in the blacklist and accidentally miss patching the right method. The other option, I can think of, would be do some platform specific logic. e.g. If `unix` then patch certain methods. This looks pretty bad but it is more explicit.

aptxkid:
So I want to discuss about what people think is the best way of dealing with this. Then I will add a test or two.

josephharrington:
I totally get your point about the downside of this approach causing blacklist typos to fail silently. On the other hand, I think that the only time we add things to the blacklist are when we notice or want to avoid specific side effects, so I'd think usually the person adding to the blacklist would do some manual checks to make sure the thing they added was actually blacklisted.

Like you said, it's a tradeoff with readability -- I probably lean more towards the way you've done it here rather than trying to essentially tag or group everything in the blacklist with what systems we expect them to exist on.

aptxkid:
@josephharrington Cool. That's what I slightly lean towards as well. I will add some tests.

josephharrington:
:+1:

PR #159, opened 2015-06-09
Hard kill the app via os._exit

### LINKED ISSUE #158: Kill the app via _exit instead of sending SIGKILL
In `main.py` we handle force kill by sending SIGKILL to the process itself. This doesn't work on Windows as SIGKILL is not defined in Python for Windows. IMO, we should use `os._exit` here.

aptxkid:
Sending SIGKILL works fine except it doesn't work on Windows.

This fixes #158

### DIFF
--- a/main.py
+++ b/main.py
@@ -3,7 +3,6 @@
 import argparse
 import hashlib
 import os
-import signal
 import sys
 import threading
 import time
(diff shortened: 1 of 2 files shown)


boxcla:

josephharrington:
:+1:

PR #161, opened 2015-06-09
Use `os.path.join` instead of hard coding temp dir

### LINKED ISSUE #160: Use `tempfile` instead of hard coding temp directory sys path in tests
`test_complete_subjob_writes_and_extracts_payload_to_correct_directory` and `test_exception_is_raised_if_problem_occurs_writing_subjobs` hard coded '/tmp/results' as the temp directory. We should change it to use `tempfile` which provides a cross-platform solution for this.

aptxkid:
Using `os.path.join` to create legic platform specific sys path for tests instead of hard coding sys paths. I tried to use tempfile.gettempdir() but it calls `os.open` at some point and `os.open` is blacklisted in the base unit test class.

This fixes #160

### DIFF
--- a/test/unit/master/test_build.py
+++ b/test/unit/master/test_build.py
@@ -1,4 +1,5 @@
 from queue import Queue
+from os.path import abspath, join
 import sys
 from threading import Event
 from unittest.mock import MagicMock, Mock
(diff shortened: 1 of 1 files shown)


boxcla:

tjlee0909:
:+1:

PR #162, opened 2015-06-09
Fix hard coded path in test_directory

aptxkid:
In test_directory, tests have hard coded paths (e.g. /var/besttimingserver), which are not cross-platform. This commit changes all the hard coded paths into cross-platform paths by using methods in os.path module and os.getcwd().


boxcla:

tjlee0909:
:+1:

Could you add a description to the PR please?

You could use your pydoc comment for the timing file:

'+        Construct the sys path of the directory where the timing file should reside based on the project_directory.
+        project_directory is the sys path of the project which contains the clusterrunner.yaml file.'

PR #163, opened 2015-06-09
Remove `Docker` project type

aptxkid:
`Docker` project type is broken and it's not being used anywhere.


boxcla:

tjlee0909:
**sigh**

Maybe one day we'll bring this back, but I don't see that in the near future.

:+1:

PR #164, opened 2015-06-09
Fall back to os.kill if os.killpg is not available

aptxkid:
os.killpg is not available on Windows, fall back to os.kill, in that case.


boxcla:

tjlee0909:
:+1:

PR #166, opened 2015-06-10
Refactor test_stop_subcommand

aptxkid:
The old test code has number of calls, hacks (e.g. NUM_OF_IS_RUNNING_CHECKS_TILL_SIGKILL) hard coded. Refactored the test code to remove the hacks and try to test the public interface of stop_subcommand in a better way.

### DIFF
--- a/test/unit/subcommands/test_stop_subcommand.py
+++ b/test/unit/subcommands/test_stop_subcommand.py
@@ -1,167 +1,181 @@
-import psutil
+from io import StringIO
+import os
+from os.path import join
 import signal
-from unittest.mock import mock_open, Mock
+from unittest.mock import call, mock_open, Mock
 
-from app.subcommands.stop_subcommand import StopSubcommand
+import psutil
+
+from app.subcommands.stop_subcommand import Configuration, StopSubcommand
 from test.framework.base_unit_test_case import BaseUnitTestCase
 
-NUM_OF_IS_RUNNING_CHECKS_TILL_SIGKILL = 4
 
 class TestStopSubcommand(BaseUnitTestCase):
 
-
     def setUp(self):
         super().setUp()
-        self.patch('os.remove')
-        self.os_kill_patch = self.patch('os.kill')
-        self.patch('time.sleep')
-        self.os_path_exists_patch = self.patch('os.path.exists')
-        self.psutil_pid_exists_patch = self.patch('psutil.pid_exists')
+        self._mock_os_path_exists = self.patch('os.path.exists')
+        self._mock_os_kill = self.patch('os.kill')
+        self._mock_psutil_pid_exists = self.patch('psutil.pid_exists')
+        self._mock_psutil_process = self.patch('psutil.Process')
 
-    def _mock_open(self, read_data):
-        open_mock = mock_open(read_data=read_data)
-        self.patch('app.subcommands.stop_subcommand.open', new=open_mock, create=True)
+        self._fake_slave_pid_file_sys_path = join(os.getcwd(), 'slave_pid_file')
+        self._fake_master_pid_file_sys_path = join(os.getcwd(), 'master_pid_file')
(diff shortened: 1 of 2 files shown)


boxcla:

test/unit/subcommands/test_stop_subcommand.py
```diff
@@ -1,167 +1,175 @@
...
-            children=[]
+        self._master_process = self._create_mock_process(
+            self._fake_master_pid,
+            cmdline=['python', 'main.py', 'master'],
         )
 
-        stop_subcommand = StopSubcommand()
-        stop_subcommand._kill_pid_in_file_if_exists('/tmp/pid_file_path.pid')
+        self._slave_child_process = self._create_mock_process(3333)
+        self._slave_process = self._create_mock_process(
```
josephharrington: A general question for you: do you think it is confusing at all when reading tests and see a `self._some_attribute` reference and wonder where it came from? I know it is usually pretty easy to find with a quick search so it can't be too much of an issue.

The reason I ask is that I've started leaning towards only setting `self` attributes in `setUp()` and then making helper methods return whatever objects they create instead of setting more attributes. It is nice because if you see `self._some_attribute` you usually can assume it was set in `setUp()` (and so is set for all tests). But the downside is a little bit more verbosity in the test since it requires extra assignment code.

test/unit/subcommands/test_stop_subcommand.py
```diff
@@ -1,167 +1,175 @@
...
+            proc.cmdline.return_value = cmdline
+        proc.children.return_value = child_processes if child_processes else []
+        return proc
 
-    def test_stop_subcommand_kills_child_proc_with_sigkill_if_still_running_after_sigterm(self):
-        main_proc_pid = 9999
-        main_threshold = NUM_OF_IS_RUNNING_CHECKS_TILL_SIGKILL - 1
-        child_threshold = NUM_OF_IS_RUNNING_CHECKS_TILL_SIGKILL
-        child_pid = main_proc_pid + 1
+    def _expect_kill_processes(self):
```
josephharrington: What do you think of renaming this to something other than "expect". I was thinking that it's not really setting up any expectations or making any assertions.

test/unit/subcommands/test_stop_subcommand.py
```diff
@@ -1,167 +1,175 @@
...
-        self.os_kill_patch.assert_called_with(main_proc_pid, signal.SIGTERM)
-        self.assertEquals(1, self.os_kill_patch.call_count)
+        self._mock_psutil_process.side_effect = [
+            self._slave_process,
+            self._master_process,
+        ]
 
-    def test_stop_subcommand_kills_proc_with_sigkill(self):
-        main_proc_pid = 9999
+        self._stop_subcommand.SIGTERM_SIGKILL_GRACE_PERIOD_SEC = 0
```
josephharrington: Would it be better to mock out time.time() instead?

josephharrington:
A few minor comments but overall looks great! Much better tests. :) :+1:

test/unit/subcommands/test_stop_subcommand.py
aptxkid: Is there any reason why you prefer mocking out `time` here? I like the current one as it exercise more code and explicitly says no grace period, which is exactly what the test wants.

test/unit/subcommands/test_stop_subcommand.py
aptxkid: You're right. Will do.

test/unit/subcommands/test_stop_subcommand.py
aptxkid: In general, `self.anything` can be confusing as they are sort of the 'global' variable inside the class. Here, specifically, I agree it can be cleaner, they don't even need to be instance variables if they are just returned by the method.

Issue #127, opened 2015-04-21
Many idle clusterrunner processes on master


tjlee0909:
When performing a 'ps aux | grep clusterrunner' on the master, you see results such as this:

jenkins   3009  0.0  1.2 2338040 960624 ?      Sl   Apr13   0:00 /home/jenkins/.clusterrunner/dist/clusterrunner master --port 43000
jenkins   3825  0.0  1.1 2128648 835540 ?      Sl   Apr13   0:00 /home/jenkins/.clusterrunner/dist/clusterrunner master --port 43000
jenkins   4380  0.0  2.0 2749068 1511444 ?     Sl   Apr16   0:00 /home/jenkins/.clusterrunner/dist/clusterrunner master --port 43000
jenkins   6512  0.0  0.8 1983204 617116 ?      Sl   Apr10   0:00 /home/jenkins/.clusterrunner/dist/clusterrunner master --port 43000
jenkins   6959  0.0  0.3 1797332 257012 ?      Sl   Apr09   0:00 
...

From PaulSouthworth:
sudo strace -p 29768
Process 29768 attached - interrupt to quit
futex(0x1033610, FUTEX_WAIT_PRIVATE, 0, NULL
-- that looks like waiting forever

This ticket is to figure out why these processes continue proliferating, even though they are idle, and get rid of it.

tjlee0909:
David Duke suggested:  http://stackoverflow.com/questions/10115126/python-requests-close-http-connection
============
10:54 s = requests.session()
s.config['keep_alive'] = False

tjlee0909:
Potentially fixed by https://github.com/box/ClusterRunner/pull/168

josephharrington:
Aw yeah, definitely fixed by #168.

PR #168, opened 2015-06-11
Remove pexpect

josephharrington:
In order to support Windows, we need to refactor our ClusterRunner git
code to function without pexpect. Our use of pexpect was just around
responding to prompts from git and ssh. This change works around those
prompts by modifying the way that git executes so that all prompts
will fail in an automated way.

### DIFF
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -1,17 +1,10 @@
-from multiprocessing import Manager, Pool  # pylint: disable=no-name-in-module
 import os
-import pexpect
-import psutil
-import queue
 import shutil
-import signal
-from time import sleep
 from urllib.parse import urlparse
 
 from app.project_type.project_type import ProjectType
 from app.util import fs, log
 from app.util.conf.configuration import Configuration
-from app.util.unhandled_exception_handler import UnhandledExceptionHandler
 
 
 class Git(ProjectType):
@@ -90,8 +83,8 @@ class Git(ProjectType):
         self._hash = hash
         self._repo_directory = self.get_full_repo_directory(self._url)
         self._timing_file_directory = self.get_timing_file_directory(self._url)
-        self._git_remote_command_executor = _GitRemoteCommandExecutor()
         self._local_ref = None
+        self._logger = log.get_logger(__name__)
 
         # We explicitly set the repo directory to 700 so we don't inadvertently expose the repo to access by other users
         fs.create_dir(self._repo_directory, self.DIRECTORY_PERMISSIONS)
@@ -102,7 +95,6 @@ class Git(ProjectType):
         # This is done in order to switch between the master's and the slave's copies of the repo while not
         # having to do something hacky in order to user the master's generated atoms on the slaves.
         actual_project_directory = os.path.join(self._repo_directory, project_directory)
-
         try:
             os.unlink(build_project_directory)
         except FileNotFoundError:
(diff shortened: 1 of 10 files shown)


boxcla:

test/framework/comparators.py
```diff
@@ -5,15 +5,15 @@ class AnyStringMatching(object):
     """
     A helper object that compares equal to any string matching the specified pattern.
     """
-    def __init__(self, pattern):
-        self._pattern = pattern
+    def __init__(self, regex_pattern):
+        self._matcher = re.compile(regex_pattern)
```
josephharrington: I refactored this class so that the regex is checked in the constructor instead of during comparison. The exception that gets raised when an invalid regex is used is not clear at all, so checking the regex pattern in the constructor will at least give a more helpful stack trace.

requirements.txt
```diff
@@ -14,6 +14,3 @@ PyYAML==3.11
 requests==2.3.0
 termcolor==1.1.0
 tornado==3.2.2
-
-# Use a version of pexpect that has fixed an issue with using pexpect from a subprocess.
-git+https://github.com/pexpect/pexpect.git@ae950f0b759f5ecb5237c7366e866225373afebc
```
josephharrington: Woo!

app/project_type/git.py
```diff
@@ -144,45 +136,63 @@ def _fetch_project(self):
...
+        :param command: The git command to execute, e.g., "fetch origin"
+        :type command: string
+        :param message: The human readable error message to log if the command fails
+        :type message: string
+        :return: The output of the process (stdout and stderr)
         :rtype: string
         """
-        return self._execute_and_raise_on_failure(command, message, self._repo_directory)
+        strict_host_key_setting = 'yes' if Configuration['git_strict_host_key_checking'] else 'no'
+        git_ssh_args = '-o BatchMode=yes -o StrictHostKeyChecking={}'.format(strict_host_key_setting)
```
tjlee0909: Here are some more options to consider:

-o PasswordAuthentication=no
-o UserKnownHostsFile=/dev/null

josephharrington: I think `BatchMode=yes` implies `PasswordAuthentication=no`. For `UserKnownHostsFile=/dev/null` are you saying we'd just do that when strict host checking is turned off?

tjlee0909: Just googled it, and you're right about the BatchMode=yes and PasswordAuthentication=no.

The UserKnownHostsFile, however, is not quite the same. If you don't set the UserKnownHostsFile=/dev/null, but have StrictHostKeyChecking=no, there is a (very) small corner case you can run into where a host that already exists in ~/.ssh/known_hosts has changed its host key, and would consequently warn you about a man-in-the-middle attack.

tjlee0909:
:+1: Nice!

app/project_type/git.py
josephharrington: Ah, yeah that makes sense. Good find! I will add it.

aptxkid:
Very Nice! :+1:

app/project_type/git.py
josephharrington: Actually, from manual testing it looks like `BatchMode=yes` will still prevent a prompt even in the case where there is a remote host key mismatch. If `StrictHostKeyChecking=yes` then the ssh command will fail immediately, and if `StrictHostKeyChecking=no` then ssh will succeed despite the key mismatch (although it will print out a warning).

Given that behavior, I lean towards not including the `UserKnownHostsFile=/dev/null` option. I think the only effect of including it is that is suppresses the mismatched hosts warning that gets printed to stdout during git operations. It doesn't do anything to prevent a prompt.

tjlee0909: Thanks for trying it out. +1

tjlee0909:
Possible linked to https://github.com/box/ClusterRunner/issues/127

PR #172, opened 2015-06-12
Fix flaky test `test_shutting_down_after_running_a_build_does_not_rai…

### LINKED ISSUE #169: Flaky test `test_shutting_down_after_running_a_build_does_not_raise_exception`
`test_shutting_down_after_running_a_build_does_not_raise_exception` is flaky and I can reproduce the failure by decorating the test with `genty_repeat` locally.

aptxkid:
…se_exception`

`slave.setup_build` is actually async as well and it will set `teardown_done_event` as well. So if the real `slave.teardown_build` is slow (because it's async as well), `self.assertTrue(teardown_done_event.wait()` will pass while `_do_build_teardown_and_reset` has not be executed even once. And now `MainThread` is calling `trigger_graceful_app_shutdown`. Remember `slave.teardown_build` is slow? Now both the teardown thread and mainthread can be in the `_do_build_teardown_and_reset` method at the same time. And it can fail the test if both pass the `if not self._project_type:` check. Waiting `slave.setup_build` to finish before all the rest solves the problem.

This fixes #169


boxcla:

josephharrington:
:+1: This test is too complicated for mere mortals.

Issue #171, opened 2015-06-12
Deny duplicate cluster slave registration requests on master


tjlee0909:
If the same slave service (same host, same port) manage to register with the master, then this inevitably leads to the master issuing multiple subjobs (potentially from different builds) to the same slave host.

Right now this is handled as a fatal error on the slave side (if it gets subjobs from different builds), then not only does the slave service die, but the master service dies as well.

^[[0m^[[1m^[[31m[2015-06-11 17:09:53.962] ERROR   Thread-438329   unhandled_excep Unhandled exception handler caught exception.
Traceback (most recent call last):
  File "/home/jenkins/ClusterRunnerBuild/app/util/safe_thread.py", line 18, in run
    super().run()
  File "/usr/local/lib/python3.4/threading.py", line 868, in run
    self._target(*self._args, **self._kwargs)
  File "/home/jenkins/ClusterRunnerBuild/app/master/slave.py", line 97, in _async_start_subjob
    response = self._network.post_with_digest(execution_url, post_data, Secret.get(), error_on_failure=True)
  File "/home/jenkins/ClusterRunnerBuild/app/util/network.py", line 60, in post_with_digest
    error_on_failure=error_on_failure)
  File "/home/jenkins/ClusterRunnerBuild/app/util/decorators.py", line 38, in function_with_retries
    return function(*args, **kwargs)
  File "/home/jenkins/ClusterRunnerBuild/app/util/network.py", line 48, in post
    return self._request('POST', *args, **kwargs)
  File "/home/jenkins/ClusterRunnerBuild/app/util/network.py", line 125, in _request
    format(url, str(resp.status_code), resp.text))
app.util.network._RequestFailedError: Request to http://pod4101-automation1017.pod.box.net:43001/v1/build/2212/subjob/65 failed with status_code 400 and response "{"error": "Attempted to start subjob 65 for build 2212, but current build id is 2220."}"
^[[0m^[[32m[2015-06-11 17:09:54.037] DEBUG   Thread-438329   unhandled_excep Executing teardown callback: <function ServiceSubcommand._write_pid_file.<locals>.remove_pid_file at 0x7f94104b1598>
[...]

PR #173, opened 2015-06-12
Deny duplicate cluster slave registration requests on master #171

tjlee0909:
Addresses https://github.com/box/ClusterRunner/issues/171.

If a ClusterRunner slave that has already been registered with the master attempts to reconnect, we need to make sure the master's internal bookkeeping should be consistent.

### DIFF
--- a/app/master/cluster_master.py
+++ b/app/master/cluster_master.py
@@ -105,22 +105,50 @@ class ClusterMaster(object):
 
         raise ItemNotFoundError('Requested slave ({}) does not exist.'.format(slave_id))
 
-    def connect_new_slave(self, slave_url, num_executors):
+    def connect_slave(self, slave_url, num_executors):
         """
-        Add a new slave to this master.
+        Connect a slave to this master.
+
+        If this slave has already been registered with the master and isn't working on a build, make sure that slave
+        is marked as alive and ready to accept builds.
 
         :type slave_url: str
         :type num_executors: int
-        :return: The slave id of the new slave
-        :rtype: int
+        :return: The response with the slave id of the slave.
+        :rtype: dict[str, str]
         """
-        slave = Slave(slave_url, num_executors)
-        self._all_slaves_by_url[slave_url] = slave
-        self._slave_allocator.add_idle_slave(slave)
+        if slave_url in self._all_slaves_by_url:
+            self._logger.warning('Slave on {} requested to connect to master, even though already connected', slave_url)
+            slave = self._all_slaves_by_url.get(slave_url)
+            slave.set_is_alive(True)
 
-        self._logger.info('Slave on {} connected to master with {} executors. (id: {})',
-                          slave_url, num_executors, slave.id)
-        return {'slave_id': str(slave.id)}
+            # If the slave has made a request to reconnect to the master, it's safe to say that whatever
+            # subjob/build that slave was working on at the time is no longer running. In order to prevent
+            # that build from hanging indefinitely (as the slave will never respond to the master with the
+            # subjob results for that build) we must cancel the build.
+            if slave.current_build_id is not None:
+                self._logger.warning('Newly reconnected slave {} was working on build {}, cancelling build.',
+                                     slave_url, slave.current_build_id)
+                try:
+                    build = self.get_build(slave.current_build_id)
+                    build.cancel()
+                except ItemNotFoundError:
+                    pass
+
+            slave.current_build_id = None
+            # It's a safe operation to try to add the same slave to the slave allocator multiple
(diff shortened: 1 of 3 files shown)


boxcla:

test/unit/master/test_cluster_master.py
```diff
@@ -57,6 +58,45 @@ def test_get_slave_returns_expected_value_given_valid_arguments(self):
...
+        master = ClusterMaster()
+
+        master.connect_slave('never-before-seen.turtles.gov', 10)
+
+        self.assertEqual(1, len(master.all_slaves_by_id()), 'Exactly one slave should be registered with the master.')
+        self.assertIsNotNone(master.get_slave(slave_id=None, slave_url='never-before-seen.turtles.gov'),
+                             'Registered slave does not have the expected url.')
+
+    def test_connect_slave_with_existing_dead_slave_marks_it_as_alive(self):
+        master = ClusterMaster()
```
aptxkid: The 'arrange' part of these three tests can be shared somewhat to reduce the number of lines.

aptxkid:
Looks like the test failed due to a missing rename. Otherwise, LGTM. :+1:

tjlee0909:
I looked at the error line and I'm positive that I did properly rename all of the calling functions. I don't think that test should have failed.

tjlee0909:
Turned out that I had to fetch & rebase.

app/master/cluster_master.py
```diff
@@ -105,22 +105,39 @@ def get_slave(self, slave_id=None, slave_url=None):
...
-                          slave_url, num_executors, slave.id)
-        return {'slave_id': str(slave.id)}
+        :return: The response with the slave id of the slave.
+        :rtype: dict[str, str]
+        """
+        if slave_url in self._all_slaves_by_url:
+            self._logger.warning('Slave on {} requested to connect to master, even though already connected', slave_url)
+            slave = self._all_slaves_by_url.get(slave_url)
+            slave.set_is_alive(True)
+            if slave.current_build_id is None:
```
josephharrington: If the slave is hard killed while working on a build (so `slave.current_build_id` is not None), and then reconnected relatively soon (before the master detects that the slave died), then it seems like the slave might be stuck after that (and whatever build it was working on might be stuck to) since from the master's perspective the slave is still working on that build?

josephharrington: To be clear, this is still way better than what we had before and should prevent the crash we experienced. I was just trying to decide if there would now be a less severe bug around slaves getting stuck upon reconnection.

tjlee0909: This is a good point, I've forgotten that this Slave object's current_build_id is just the master's last status update from that slave.

Do you think what would make the most sense is to force the current_build_id of the slave to None no matter what?

tjlee0909: It also means that the build that the slave was working on (before it was force-killed) will never complete because the Master will continually wait for that subjob to be posted back.

tjlee0909: Yeah, I thought about it, and if you see a re-connection request from the slave coming into the master, you have to assume that any previously running slave process must be dead, and that the subjob that it was working on (if it was working on anything) will never return.

So the things to do are:
- Set the slave's current_build_id to None, and have it be ready to start working on new builds.
- Interrupt the existing running build that the slave was building (if it was working on any) and fail it somehow so that it doesn't permanently hang.

josephharrington: Agreed. Theoretically we could just do `build.cancel()` -- I'm not 100% sure that would "unstick" the build but it should.

tjlee0909: Oh sweet, we have a public method, mark_failed() on the Build object.

tjlee0909: Ok, I'll do that.

test/unit/master/test_cluster_master.py
```diff
@@ -57,6 +58,48 @@ def test_get_slave_returns_expected_value_given_valid_arguments(self):
...
+        existing_slave_id = existing_slave.id
+
+        connect_response = master.connect_slave('existing-slave.turtles.gov', 10)
+
+        self.assertEqual(str(existing_slave_id), connect_response['slave_id'],
+                         'The re-connected slave should not have generated a new slave id.')
+        self.assertTrue(existing_slave.is_alive(use_cached=True),
+                        'The re-connected slave should have been marked as alive once reconnected.')
+        self.assertEquals(2, self.mock_slave_allocator.add_idle_slave.call_count,
+                        'Expected slave to be added to the idle slaves list.')
```
josephharrington: Nice!

josephharrington:
I think we've fixed the flaky tests now, so if you rebase/repush the tests should pass.

PR #175, opened 2015-06-16
Use psutil to kill and terminte processes

aptxkid:
SIGKILL is not supported on Windows. This commit makes it skip the kill and log a warning.

### DIFF
--- a/app/subcommands/stop_subcommand.py
+++ b/app/subcommands/stop_subcommand.py
@@ -1,5 +1,4 @@
 import os
-import signal
 import time
 
 import psutil
@@ -86,10 +85,10 @@ class StopSubcommand(Subcommand):
 
     def _terminate_running_procs(self, procs_to_termintate):
         for proc in [p for p in procs_to_termintate if p.is_running()]:
-            self._logger.info("Sending SIGTERM to PID {}", proc.pid)
-            os.kill(proc.pid, signal.SIGTERM)
+            self._logger.info("Terminating process with PID {}", proc.pid)
+            proc.terminate()
 
     def _kill_running_procs(self, procs_to_kill):
         for proc in [p for p in procs_to_kill if p.is_running()]:
-            self._logger.info("Sending SIGKILL to PID {}", proc.pid)
-            os.kill(proc.pid, signal.SIGKILL)
+            self._logger.info("Killing process with PID {}", proc.pid)
+            proc.kill()
--- a/test/unit/subcommands/test_stop_subcommand.py
+++ b/test/unit/subcommands/test_stop_subcommand.py
@@ -1,7 +1,7 @@
+from functools import partial
 from io import StringIO
 import os
 from os.path import join
-import signal
 from unittest.mock import call, mock_open, Mock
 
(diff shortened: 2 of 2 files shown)


boxcla:

app/subcommands/stop_subcommand.py
```diff
@@ -91,5 +91,9 @@ def _terminate_running_procs(self, procs_to_termintate):
 
     def _kill_running_procs(self, procs_to_kill):
         for proc in [p for p in procs_to_kill if p.is_running()]:
-            self._logger.info("Sending SIGKILL to PID {}", proc.pid)
-            os.kill(proc.pid, signal.SIGKILL)
+            if 'SIGKILL' in dir(signal):
```
josephharrington: What do you think of changing this to `if hasattr(signal, 'SIGKILL'):` instead? [This SO question](http://stackoverflow.com/questions/17723569/whats-the-difference-between-hasattr-and-attribute-in-dir) makes me think that it might the preferred approach.

app/subcommands/stop_subcommand.py
```diff
@@ -91,5 +91,9 @@ def _terminate_running_procs(self, procs_to_termintate):
...
+                self._logger.info("Sending SIGKILL to PID {}", proc.pid)
+                os.kill(proc.pid, signal.SIGKILL)
+            else:
+                self._logger.warning("Can't kill PID {} as SIGKILL is not supported on the platform. This is expected"
+                                     "if ClusterRunner is running on Windows.", proc.pid)
```
josephharrington: Want to move this functionality into [`app/util/process_utils.py`](../../blob/master/app/util/process_utils.py)? I think there are a few places in the app where we might need this change. :)

josephharrington: Actually (sorry) I looked at the `psutil` docs a little bit and am wondering if you had run across the `proc.kill()` method -- it looks like it handles both unix and windows?

test/unit/subcommands/test_stop_subcommand.py
```diff
@@ -169,13 +169,15 @@ def terminate_only_master_process_successfully(pid, _):
         self._stop_subcommand.run(None)
 
         # Assert
+        expected_call_args_list = []
+        expected_call_args_list.append(call(slave_child_process.pid, signal.SIGTERM))
+        expected_call_args_list.append(call(slave_process.pid, signal.SIGTERM))
```
josephharrington: Just curious why you chose to start with an empty list and append elements vs just doing a list literal? (I was guessing you chose that so it matched the semantics of the if-statement below, and thought if so then you could still make it match with a list literal by using `extend()` instead of `append()`.)

aptxkid:
@josephharrington All good comments! I will address all.

aptxkid:
`test_basic_directory_configs_end_to_end` functional test is flaky. Opened https://github.com/box/ClusterRunner/issues/180

tjlee0909:
:+1:

josephharrington:
:+1:

PR #183, opened 2015-06-17
Fix coveralls Windows incompatibility

josephharrington:
Coveralls is only needed on Travis, so we shouldn't need it in the
general requirements.txt file. In fact if this is present in
requirements.txt then pip install on Windows will fail.


boxcla:

aptxkid:
:+1:

PR #186, opened 2015-06-18
Only replace invalid path characters from netloc

aptxkid:
':' is not a valid character in path on Windows. But we cannot replace all colons on a Windows path (e.g. c:\\users\\me).

### DIFF
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -19,34 +19,59 @@ class Git(ProjectType):
     DIRECTORY_PERMISSIONS = 0o700
 
     @staticmethod
-    def get_full_repo_directory(url):
+    def _generate_path_from_repo_url(base_sys_path, url):
         """
-        Generates a directory to house the repo based on the origin url
-        :return: A path to clone the git repo in
+        Generate a sys path based on the base_sys_path and the git repo url. It also removes some invalid
+        characters from the final generated sys path.
+
+        e.g. _generate_path_from_repo_url('/tmp', 'ssh://source_control.cr.com:1234/master-repo') returns
+        /tmp/source_control.cr.com1234/masterrepo
+
+        :param base_sys_path: The base sys path of the generated sys path
+        :type base_sys_path: str
+        :param url: The remote 'origin' url of the git repo
+        :type url: str
+
+        :return: Sys path of combining the base_sys_path and generated sys path from url
         :rtype: str
         """
         url_components = urlparse(url)
         url_full_path_parts = url_components.path.split('/')
         repo_name = url_full_path_parts[-1].split('.')[0]
         url_folder_path_parts = url_full_path_parts[:-1]
-        repo_directory = os.path.join(Configuration['repo_directory'], url_components.netloc, *url_folder_path_parts)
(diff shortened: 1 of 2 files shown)


boxcla:

app/project_type/git.py
```diff
@@ -19,6 +19,19 @@ class Git(ProjectType):
...
+        """
+        Replace the following characters in netloc.
+        ':' -> '-' (':' is not a valid character in path on Windows)
+
+        :param netloc: Network location
+        :type netloc: str
+        :return: Path after replacement
+        :rtype: str
+        """
+        return netloc.replace(':', '-')
```
josephharrington: I think we might need to replace dashes also. The comment in the git history of the `remove_invalid_path_characters` method says "The PHP Intl extension doesn't appear to work if the repo is in a directory with a dash in the path."

And we should document that in a code comment as well. :)

test/unit/project_type/test_git.py
```diff
@@ -66,11 +66,11 @@ def test_execute_command_in_project_type_specifies_cwd_if_doesnt_exist(self):
 
     def test_get_full_repo_directory(self):
```
josephharrington: Want to add a test case for a Windows style path (since that is the situation this PR is fixing)? :wink:

app/project_type/git.py
aptxkid: I see. I went back and forth on this one. I think it's still the best to preserve the old replace behavior.

test/unit/project_type/test_git.py
aptxkid: I've a separate PR to do this in order to keep each PR small.

app/project_type/git.py
```diff
@@ -19,34 +19,59 @@ class Git(ProjectType):
...
         """
-        url_components = urlparse(url)
-        timings_directory = os.path.join(
-            Configuration['timings_directory'],
-            url_components.netloc,
-            url_components.path.strip('/')
-        )
-        return fs.remove_invalid_path_characters(timings_directory)
+
+        return Git._generate_path_from_repo_url(Configuration['timings_directory'], url)
```
josephharrington: Can you give an example of what the timing data would be previously and what it will be now with this change?

aptxkid: @josephharrington It's almost the same as far as to the existing tests' concern. Before this change, `get_full_repo_directory` and `get_timing_file_directory` will construct the path from url and remove colons at the end. Now we only get rid of colons that's in the netloc of the url. This makes a difference on Windows as the path being constructed will contain colon (c:\\bla) and we don't want to remove the colon here. I didn't add the test for windows style path here to make the PR small. So this PR is only about setting up the stage for the next PR while preserving all the existing contracts in test.

josephharrington: So the existing timing directories will not change at all?

aptxkid: If you're asking about our environment, no, it won't change. We should still be able to load the timing file from existing location.

aptxkid:
This PR has been sitting here for quite a long time. I am merging it right now. Will respond to any comments.

PR #189, opened 2015-06-22
Make job_configs for multiple platforms

aptxkid:
Before this commit, job_configs has three configs that only works in POSIX systems. This commit makes it so that each config contains a dictionary pointing to different real configs for different underlying platform so running the functional test on Windows will run Windows command and on Mac will run POSIX commands.

### DIFF
--- a/test/functional/job_configs.py
+++ b/test/functional/job_configs.py
@@ -23,8 +23,8 @@ class FunctionalTestJobConfig(object):
 
 # This is a very basic job where each atom just creates a simple text file.
 BASIC_JOB = FunctionalTestJobConfig(
-    config="""
-
+    config={
+        'posix': """
 BasicJob:
     commands:
         - echo $TOKEN > $ARTIFACT_DIR/result.txt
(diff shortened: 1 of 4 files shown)


boxcla:

test/functional/job_configs.py
```diff
@@ -58,6 +67,9 @@ def __init__(self, config, expected_to_fail, expected_num_subjobs, expected_num_
         - TOKEN: seq 0 4 | xargs -I {} echo "This is atom {}"
 
 """,
+        'nt': """
+""",
```
josephharrington: Why is this config empty?

aptxkid: This PR is more about the 'how' to have platform specific commands in functional test configs than 'what' exactly those commands would be. I am rewriting those shell commands in batch, but it's gonna take a little while. This is also good for keep each PR small. What you think?

aptxkid: @josephharrington  I have one last job config to rewrite anyway. I will update the PR with all batch commands under key 'nt'.

josephharrington: Sounds good. That's what I figured but I wanted to make sure.

josephharrington:
:+1:

PR #190, opened 2015-06-23
Fix yaml filename inconsistencies in docstrings

josephharrington:
Several places in the code had docstrings referring to
"cluster_runner.yaml" which is misleading since the file we look for
by default is actually "clusterrunner.yaml".


boxcla:

aptxkid:
:+1:

PR #192, opened 2015-06-24
Add JOB_WITH_SETUP_AND_TEARDOWN config for 'nt'

aptxkid:
I rewrote the shell commands into batch commands. It's almost an one-to-one mapping except 'sleep 1'. 'timeout /t 1' fails with ERROR: Input redirection is not supported, exiting the process immediately. I figured it should not block this change as the purpose of sleep is to just delay the execution a little bit so I replaced with 'ping' localhost.


boxcla:

josephharrington:
This is insanely great. Nice work! :+1:

PR #193, opened 2015-06-24
Replace Popen in app with Popen_with_delayed_expansion

aptxkid:
Popen_with_delayed_expansion is a cross-platform wrapper around Popen which make sure all the environment variables in the command will be expanded only at execution time. Due to how ClusterRunner chains all commands together, this is almost always the behavior we want.


boxcla:

josephharrington:
:+1:

PR #194, opened 2015-06-24
Introduce environment variable setter command factory

aptxkid:
It helps construct platform specific command for setting environment variables.

### DIFF
--- a/app/util/process_utils.py
+++ b/app/util/process_utils.py
@@ -27,6 +27,14 @@ def kill_gracefully(process, timeout=2):
     return process.returncode, stdout, stderr
 
 
+def _is_windows():
+    """
+    :return: Whether ClusterRunner is running on Windows or not>
+    :rtype: bool
+    """
+    return os.name == 'nt'
+
+
 def Popen_with_delayed_expansion(cmd, *args, **kwargs):
     """
     A thin wrapper around subprocess.Popen which ensures that all environment variables in the cmd are expanded at
@@ -40,7 +48,7 @@ def Popen_with_delayed_expansion(cmd, *args, **kwargs):
     :return: Popen object, just like the Popen object returned by subprocess.Popen
     :rtype: :class:`Popen`
     """
-    if os.name == 'nt':
+    if _is_windows():
         cmd_with_deplayed_expansion = ['cmd', '/V', '/C']
         if isinstance(cmd, str):
             cmd_with_deplayed_expansion.append(cmd)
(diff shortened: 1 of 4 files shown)


boxcla:

app/util/process_utils.py
```diff
@@ -48,3 +48,21 @@ def Popen_with_delayed_expansion(cmd, *args, **kwargs):
...
+    :type name: str
+    :param value: The value of the environment variable
+    :type value: str
+    :return: Platform specific command for setting the environment variable
+    :rtype: str
+    """
+    if os.name == 'nt':
+        return 'set {}={}&&'.format(name, value)
+    else:
+        return 'export {}="{}";'.format(name, value)
```
josephharrington: Do you know if there is any benefit to embedding the env var setter into the command vs. using Popen's `env` param? I've wondered about that for awhile -- it would seem cleaner to use the `env` param.

aptxkid: I thought about that as well. But it might require a (big?) refactor. e.g. Atom has this `command_string` property which is just a command string. I prefer setting environment variables in Popen as well. But this unblocks Windows with minimum effort AFAIK. I suggest we add an issue to track it so we can get rid of it later.

josephharrington: I was referring more to the code that does it in project_type.py -- changing that would be a pretty small change. But yeah changing the Atom class to support that pattern would go pretty deep.

app/util/process_utils.py
```diff
@@ -48,3 +48,21 @@ def Popen_with_delayed_expansion(cmd, *args, **kwargs):
...
+    is designed to be chained with other commands.
+
+    :param name: The name of the environment variable
+    :type name: str
+    :param value: The value of the environment variable
+    :type value: str
+    :return: Platform specific command for setting the environment variable
+    :rtype: str
+    """
+    if os.name == 'nt':
```
josephharrington: Should we have something like an `is_windows()` helper? I doubt we'd need to change the implementation but it might be a tad easier to read.

josephharrington:
:+1: with a few tiny comments.

app/util/process_utils.py
aptxkid: @josephharrington Good idea! Will do.

app/util/process_utils.py
aptxkid: I found this from https://docs.python.org/3.4/library/subprocess.html#subprocess.Popen:
"Note If specified, env must provide any variables required for the program to execute. On Windows, in order to run a side-by-side assembly the specified env must include a valid SystemRoot.
"

This is kind of interesting actually. E.g. In order to run nosetests, of course, you need to have `python` in `$PATH`. Right now, CR is operating with an assumption that the machine is already in a good state to run the job command, which is a fine assumption. If we want to go down the `env` route then we need to set all the environment variable explicitly, which right now are inherited from parent process. This actually can be beneficial if we make everything explicit (e.g. even install Python in setup if it's not in $PATH, etc.) But this is a breaking change so I doubt we are going to make this change right now.

app/util/process_utils.py
```diff
@@ -27,6 +27,14 @@ def kill_gracefully(process, timeout=2):
     return process.returncode, stdout, stderr
 
 
+def _is_windows():
+    """
```
aptxkid: I am leaving this method 'private' to this package for now since it's only being used in this package at this moment. We can totally move it to a better place if there are more usages.

app/util/process_utils.py
josephharrington: The way I've done it in the past is just to copy `os.environ` which is the current environment, and then add in extra env vars as desired.

josephharrington:
:+1:

PR #195, opened 2015-06-25
Use join & expanduser to make test_git cross-platform

aptxkid:
It used to hard code paths like '/tmp/repos', which only works on POSIX systems.


boxcla:

josephharrington:
Nice. :+1:

PR #196, opened 2015-06-25
Skip cloning from master if 'get_project_from_master' is not set

aptxkid:


boxcla:

app/util/conf/base_config_loader.py
```diff
@@ -87,6 +87,11 @@ def configure_defaults(self, conf):
         conf.set('git_askpass_exe', join(bin_dir, 'git_askpass.sh'))
         conf.set('git_ssh_exe', join(bin_dir, 'git_ssh.sh'))
 
+        # How slaves get the project
+        # Slaves would get the project from master if set to True. Otherwise it would just get the project in
+        # the same way how the master gets the project.
+        conf.set('get_project_from_master', True)
```
josephharrington: Should we default to false? With this setting we should also be able to unskip the functional test `test_git_type_demo_project_config`. :smile:

aptxkid: Default to false is a breaking change though...

josephharrington: Lu and I discussed offline and decided to add the conf setting without changing the default behavior, and postpone changing the default for a future change.

test/unit/project_type/test_git.py
```diff
@@ -158,11 +160,13 @@ def test_execute_git_command_auto_sets_strict_host_option_correctly(self, strict
         self.assertIn(expected_call, popen_mock.call_args_list, 'Executed git command should include the correct '
                                                                 'option for StrictHostKeyChecking.')
 
+    @skipIf(os.name == 'nt', 'Skipping test for cloning repo from master on Windows')
```
josephharrington: is_windows()? :wink:

test/unit/project_type/test_git.py
```diff
@@ -175,6 +179,14 @@ def test_slave_param_overrides_returns_expected(self):
         self.assertEqual(expected_overrides, actual_overrides, 'Slave param overrides from Git object should match'
                                                                'expected.')
 
+    def test_slave_param_overrides_returns_empty_dict_when_get_project_from_master_is_disabled(self):
```
josephharrington: What do you think of just asserting that 'url' and 'branch' are not keys in the overrides dict? It might make this test more stable against future unrelated additions to slave param overrides.

PR #200, opened 2015-06-26
Skip clusterrunner.conf file mode on Windows

aptxkid:
Permission model on Windows is different than Mac. Python doesn't support Windows file permission very well.
Skip the check on Windows for now. Since the clusterrunner.conf file should live in the user's home directory,
which is usually not accessible by others, this is OK for now.

### DIFF
--- a/app/util/conf/config_file.py
+++ b/app/util/conf/config_file.py
@@ -3,6 +3,7 @@ import os
 import stat
 
 from app.util import fs
+from app.util.process_utils import is_windows
 
 
 class ConfigFile(object):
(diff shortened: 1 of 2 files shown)


boxcla:

app/util/conf/config_file.py
```diff
@@ -21,7 +21,7 @@ def read_config_from_disk(self):
         if not os.path.isfile(self._filename):
             raise FileNotFoundError('Conf file {} does not exist'.format(self._filename))
         file_mode = stat.S_IMODE(os.stat(self._filename).st_mode)
-        if file_mode != self.CONFIG_FILE_MODE:
+        if os.name == 'nt' or file_mode != self.CONFIG_FILE_MODE:
```
josephharrington: Want to add a comment about why we skip Windows?

josephharrington:
:+1: with a minor comment

Issue #198, opened 2015-06-25
Reconnecting slave can deadlock master's server


josephharrington:
Steps to reproduce this issue locally:
```bash
# Launch a master and a slave
clusterrunner master &
clusterrunner slave &
# Shutdown that slave via the shutdown endpoint
clusterrunner shutdown --master-url localhost:43000 --slave-id 1
# Reconnect the same slave
clusterrunner slave &
# Now both the master and slave web servers are locked up. All requests to them will hang.
```

The reason this happens is due to a bug around marking a slave idle that has been put into "shutdown mode".
1. User tells master to put slave X in shutdown mode. 
2. Slave shuts down. (Master still has it flagged as in shutdown mode.)
3. User reconnects slave. This is when the deadlock happens.
4. __*** Slave makes "connect" api call to master. ***__
&nbsp;&nbsp;&nbsp;&nbsp;a. Master sees that slave was already connected previously. `slave_allocator.add_idle_slave()` gets called which in turn calls `slave.mark_as_idle()`. 
&nbsp;&nbsp;&nbsp;&nbsp;b. `slave.mark_as_idle()` sees that the slave is in shutdown mode and calls `slave.kill()`. (This should not happen at all in this case.)
5. __*** Master makes "kill" api call to slave. ***__ This starts the slave shutdown process, part of which is disconnecting from the master.
6. __*** Slave makes "disconnect" api call to master. ***__ 

The last call will block forever since the master server is still handling the _first_ "connect" request. Since there are pending requests to both the master and the slave, both web servers deadlock on each other and clusterrunner completely locks up. Any other connected slaves would be able to receive requests, but cannot report results back to the master.

There are several issues exposed here:
- Network calls can hang forever. We should add a timeout for all network calls (see #197).
- Slave state is not completely reset upon reconnection. Should we just create a new `Slave` instance entirely instead of reusing the old one?
- Too many synchronous network calls. This is especially dangerous when an outgoing request is made synchronously during handling of an incoming request.

josephharrington:
Fixed by #202

PR #202, opened 2015-06-26
When previously connected slave reconnects, delete old instance.

tjlee0909:
...and instantiate the slave again and re-insert into the master's
bookkeeping system.

This fixes the bug/issue from https://github.com/box/ClusterRunner/issues/198 that leads to the service freezing up upon reconnectin an existing slave.

### DIFF
--- a/app/master/cluster_master.py
+++ b/app/master/cluster_master.py
@@ -109,46 +109,38 @@ class ClusterMaster(object):
         """
         Connect a slave to this master.
 
-        If this slave has already been registered with the master and isn't working on a build, make sure that slave
-        is marked as alive and ready to accept builds.
-
         :type slave_url: str
         :type num_executors: int
         :return: The response with the slave id of the slave.
         :rtype: dict[str, str]
         """
+        # If a slave had previously been connected, and is now being reconnected, the cleanest way to resolve this
+        # bookkeeping is for the master to forget about the previous slave instance and start with a fresh instance.
         if slave_url in self._all_slaves_by_url:
-            self._logger.warning('Slave on {} requested to connect to master, even though already connected', slave_url)
-            slave = self._all_slaves_by_url.get(slave_url)
-            slave.set_is_alive(True)
+            self._logger.warning('Slave on {} requested to connect to master, even though previously connected. ' +
+                                 'Removing existing slave instance from the master\'s bookkeeping.', slave_url)
+            old_slave = self._all_slaves_by_url.get(slave_url)
 
-            # If the slave has made a request to reconnect to the master, it's safe to say that whatever
-            # subjob/build that slave was working on at the time is no longer running. In order to prevent
-            # that build from hanging indefinitely (as the slave will never respond to the master with the
-            # subjob results for that build) we must cancel the build.
-            if slave.current_build_id is not None:
-                self._logger.warning('Newly reconnected slave {} was working on build {}, cancelling build.',
-                                     slave_url, slave.current_build_id)
+            # If a slave has requested to reconnect, we have to assume that whatever build the dead slave was
+            # working on no longer has valid results.
+            if old_slave.current_build_id is not None:
+                self._logger.info('{} has build [{}] running on it. Attempting to cancel build.', slave_url,
+                                  old_slave.current_build_id)
                 try:
-                    build = self.get_build(slave.current_build_id)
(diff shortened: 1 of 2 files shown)


boxcla:

app/master/cluster_master.py
```diff
@@ -109,46 +109,24 @@ def connect_slave(self, slave_url, num_executors):
...
-        else:
-            slave = Slave(slave_url, num_executors)
-            self._all_slaves_by_url[slave_url] = slave
-            self._slave_allocator.add_idle_slave(slave)
-            self._logger.info('Slave on {} connected to master with {} executors. (id: {})',
-                              slave_url, num_executors, slave.id)
-            slave_id = slave.id
-        return {'slave_id': str(slave_id)}
+            self._logger.warning('Slave on {} requested to connect to master, even though previously connected. ' +
+                                 'Removing existing instance from master._all_slaves_by_url.', slave_url)
```
aptxkid: Probably not to include implementation details like `_all_slaves_by_url` in log message

app/master/cluster_master.py
```diff
@@ -109,46 +109,24 @@ def connect_slave(self, slave_url, num_executors):
...
+            self._all_slaves_by_url.pop(slave_url)
```
aptxkid: Just curious. What would happen to the master if the reconnect slave was running a job?

app/master/cluster_master.py
tjlee0909: Good point.

app/master/cluster_master.py
tjlee0909: Ah, you know what, that's a great point. I think if the existing slave is alive, then:

- if the slave is in shutdown mode, then we should set it to not be in shutdown mode
- if the slave is not in shutdown mode, just reject the request and return the existing slave id json response

What do you think? Also, great catch.

josephharrington: Can you guys think of a situation where the action of reconnecting a slave would not also be a signal to the master that any existing slave with the same url should be considered dead? Like if we reconnect a slave, then shouldn't we always assume that any build it was running previously should be canceled?

tjlee0909: A scenario that I can think of (and, sadly, it might be a legitimate scenario) is that one of us derps it on the keyboard, and tries to connect the same slave to the same master multiple times during a deploy in rapid succession of each other. In that particular scenario, the slave wouldn't be running a build though, since it was just connected.

In any case, if a healthy slave is working on a build and connected to the master, and it's is_alive() is true--shouldn't we just ignore the reconnect requests? I was thinking of something like this:

if slave.is_alive():
    if slave.is_shutdown():
        slave.set_shutdown(false)
    else:
        do nothing, do not instantiate new slave
else:
   remove existing slave instance and instantiate new one

aptxkid: Is there a chance of race condition here while getting slave's status?

tjlee0909: Joey and I had an offline conversation. We're going to keep what we have, but additionally we're going to perform a check on the old slave instance and see if it has a current_build_id set; we will then cancel it.

app/master/cluster_master.py
```diff
@@ -109,46 +109,38 @@ def connect_slave(self, slave_url, num_executors):
...
-            # If the slave has made a request to reconnect to the master, it's safe to say that whatever
-            # subjob/build that slave was working on at the time is no longer running. In order to prevent
-            # that build from hanging indefinitely (as the slave will never respond to the master with the
-            # subjob results for that build) we must cancel the build.
-            if slave.current_build_id is not None:
-                self._logger.warning('Newly reconnected slave {} was working on build {}, cancelling build.',
-                                     slave_url, slave.current_build_id)
+            self._logger.warning('Slave on {} requested to connect to master, even though previously connected. ' +
+                                 'Removing existing slave instance from the master\'s bookkeeping.', slave_url)
+            removed_slave = self._all_slaves_by_url.pop(slave_url)
```
josephharrington: We might want to leave the old slave in the dict and just overwrite it with the new one. I'm wondering if there might be a tiny race between this line and when we add the new one below -- if some other thread tries to access the slave in `_all_slaves_by_url` it will get an exception. (I don't have an exact case for this in mind, it's just a feeling.)

josephharrington:
One minor comment, but otherwise looks good. :+1:

PR #204, opened 2015-06-28
Add event log for failed build prepare

josephharrington:


boxcla:

tjlee0909:
:+1:

Cool, do you have any plans yet on where you want to use this new event?

josephharrington:
@tjlee0909 this is actually going to be used right away to fix a bug in the timeline dashboard where if prepare fails, we never get an end event and the timespan bar extends to the right indefinitely.

PR #205, opened 2015-06-29
Skip cloning from master if 'get_project_from_master' is not set

aptxkid:
Default value of 'get_project_from_master' is True so this is not a breaking change.


boxcla:

PR #206, opened 2015-06-29
Fix tests in test_git with hard coded env setter

aptxkid:
Commands for setting environment variables are platform specific. Tests are failing on Windows, as
they have harded coded commands for setting environment variables. This commit changes to use the
`get_envrionment_variable_setter_command` method in process_utils module to make the tests
cross-platform.


boxcla:

PR #209, opened 2015-06-29
Patch Popen_with_delayed_expansion instead of Popen

aptxkid:
The test used to patch subprocess.Popen. It fails on Windows as we invoke different commands
on Popen depending on the platform. Patching Popen_with_deplayed_expansion instead so the test is
cross platform.


boxcla:

josephharrington:
:+1:

PR #210, opened 2015-06-29
Remove the trailing spaces from the command constructd by chaining a …

aptxkid:
…list of commands

The trailing space is fine in bash not less so in windows command prompt.


boxcla:

josephharrington:
:+1:

PR #211, opened 2015-06-30
Add 'exit_code' to atom's API representation

aptxkid:


boxcla:

app/master/build.py
```diff
@@ -228,6 +229,27 @@ def complete_subjob(self, subjob_id, payload=None):
...
+    def _get_atom_artifact_sys_path(self, subjob_id, atom_id):
+        return os.path.join(
+            self._build_results_dir(),
+            Subjob.ATOM_DIR_FORMAT.format(
+                subjob_id,
+                atom_id,
+            ),
+        )
+
+    def _parse_payload_for_atom_exit_code(self, subjob_id):
```
tjlee0909: I'm going to use this same method to set the syspath for the console_output_file in the Atom class.

Because of that, do you mind renaming this function to something slightly more generic, maybe just "_parse_paylod_for_atom"?

tjlee0909: Actually nm, that might not be the best idea. I'll think about it after I implement it, you can leave this alone for now, and I'll rename it if needed.

tjlee0909:
:+1:

app/master/build.py
```diff
@@ -228,6 +229,27 @@ def complete_subjob(self, subjob_id, payload=None):
             self.mark_failed('Error occurred while completing subjob {}.'.format(subjob_id))
             raise
 
+    def _get_atom_artifact_sys_path(self, subjob_id, atom_id):
+        return os.path.join(
+            self._build_results_dir(),
+            Subjob.ATOM_DIR_FORMAT.format(
+                subjob_id,
+                atom_id,
```
josephharrington: I wonder if this code should be more central -- this same `.format()` call is already happening in `Subjob`. Generally it gives me an uncomfortable feeling to have unformatted strings very far away from the place they are formatted -- it's a form of tight coupling.

app/master/build.py
```diff
@@ -228,6 +229,27 @@ def complete_subjob(self, subjob_id, payload=None):
...
+        )
+
+    def _parse_payload_for_atom_exit_code(self, subjob_id):
+        subjob = self.subjob(subjob_id)
+        for atom_id in range(len(subjob.atoms)):
+            with open(
+                os.path.join(
+                    self._get_atom_artifact_sys_path(subjob_id, atom_id),
+                    Subjob.EXIT_CODE_FILE,
+                ),
```
josephharrington: This is minor, but I was wondering what you think -- would this chunk be easier to read if we created the filename in a variable on a previous line?

aptxkid: Totally. Will do.

app/master/build.py
aptxkid: Good point. I might just put this logic in `Subjob`.

PR #212, opened 2015-07-01
Instead of writing stdout/ stderr to temp files, write to artifact path

tjlee0909:
On the slave service, we are writing stdout and stderr to two separate
temporary files, and later copying the contents of them to its final
path in the artifact directory.

This will not only be slightly more efficient, but it allows us to
implement returning the console output of in-progress builds from
slaves.

### DIFF
--- a/app/project_type/project_type.py
+++ b/app/project_type/project_type.py
@@ -3,7 +3,7 @@ import inspect
 import os
 import re
 import signal
-from subprocess import TimeoutExpired
+from subprocess import TimeoutExpired, STDOUT
 from tempfile import TemporaryFile
 from threading import Event
 import time
@@ -170,7 +170,8 @@ class ProjectType(object):
         """
         return command
 
-    def execute_command_in_project(self, command, extra_environment_vars=None, timeout=None, **popen_kwargs):
+    def execute_command_in_project(self, command, extra_environment_vars=None, timeout=None, output_file=None,
+                                   **popen_kwargs):
         """
         Execute a command in the context of the project
 
@@ -180,6 +181,9 @@ class ProjectType(object):
         :type extra_environment_vars: dict[str, str]
         :param timeout: A maximum number of seconds before the process is terminated, or None for no timeout
         :type timeout: int | None
+        :param output_file: The file to write console output to (both stdout and stderr). If not specified,
+                            will generate a TemporaryFile. This method will close the file.
+        :type output_file: BufferedRandom | None
         :param popen_kwargs: additional keyword arguments to pass through to subprocess.Popen
         :type popen_kwargs: dict[str, mixed]
         :return: a tuple of (the string output from the command, the exit code of the command)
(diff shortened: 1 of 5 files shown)


boxcla:

app/project_type/project_type.py
```diff
@@ -190,17 +194,50 @@ def execute_command_in_project(self, command, extra_environment_vars=None, timeo
...
-        stdout_file = TemporaryFile()
-        stderr_file = TemporaryFile()
+        output_file = output_file if output_file is not None else TemporaryFile()
         pipe = Popen_with_delayed_expansion(
             command,
             shell=True,
-            stdout=stdout_file,
-            stderr=stderr_file,
+            stdout=output_file,
+            stderr=STDOUT,  # Redirect stderr to stdout, as we do not care to distinguish the two.
```
tjlee0909: I made a judgment call to just have stdout and stderr be piped to the same file, as we just combine the output from the two streams anyway.

Also, without having just a single console output file, it makes it impossible to implement streaming console output, and we never have the combined output contents in a single file until the very end.

app/project_type/project_type.py
```diff
@@ -190,17 +194,50 @@ def execute_command_in_project(self, command, extra_environment_vars=None, timeo
...
+                'Command exited with non-zero exit code.\nCommand: {}\nExit code: {}\nConsole output: {}\n',
+                command, exit_code, logged_console_output)
+        else:
+            self._logger.debug('Command completed with exit code {}.', exit_code)
+
+        exit_code = exit_code if exit_code is not None else -1  # Make sure we always return an int.
+        combined_command_output = '\n'.join([console_output] + clusterrunner_error_msgs)
+        return combined_command_output, exit_code
+
+    def _wait_for_pipe_to_close(self, pipe, command, timeout):
```
tjlee0909: git/github royally messed up the code diff here. I basically had ripped out a 50-line chunk of code from execute_command_in_project() into a new method, _wait_for_pipe_to_close().

tjlee0909:
this is a WIP, but I wanted early feedback on it.

app/project_type/project_type.py
josephharrington: Sounds good to me.

app/project_type/project_type.py
```diff
@@ -190,17 +194,50 @@ def execute_command_in_project(self, command, extra_environment_vars=None, timeo
...
+        """
+        Wait for the pipe to close (after comment completes) or until timeout. If timeout is reached, then
```
josephharrington: By "comment" did you mean "command" here?

app/slave/subjob_executor.py
```diff
@@ -131,14 +131,15 @@ def _execute_atom_command(self, atomic_command, atom_environment_vars, atom_arti
         :rtype: int
         """
         fs_util.create_dir(atom_artifact_dir)
+        # This console_output_file must be opened in 'w+b' mode in order to be interchangeable with the
+        # TemporaryFile instance that gets instantiated in self._project_type.execute_command_in_project.
+        console_output_file = open(os.path.join(atom_artifact_dir, Subjob.OUTPUT_FILE), mode='w+b')
```
josephharrington: One thing that popped into my head here was that we need to make sure this gets closed. I can't tell if the `TemporaryFile` objects we were using before do more to make sure they get closed than a simple `open()`, but it might be more important since it's not going to be as big of a deal if we're holding open a random tempfile.

Also, I see in the happy path case we should close the file in `ProjectType._read_file_contents_and_close()` but if we hit an exception before that we'd miss that close call (it's not in a finally block).

I wonder if we should just use the `with open(...)` syntax here around the call to `self._project_type.execute_command_in_project` just to make sure it gets closed?

josephharrington:
Looks good overall. :+1:

app/project_type/project_type.py
tjlee0909: I did, thanks.

app/slave/subjob_executor.py
tjlee0909: Sounds good.

PR #213, opened 2015-07-02
Add `state` to atom API

aptxkid:
An Atom can be in one of the three possible states: NOT_STARTED, IN_PROGRESS, COMPLETED

### DIFF
--- a/app/master/atom.py
+++ b/app/master/atom.py
@@ -1,4 +1,12 @@
+from enum import Enum
+
 from app.util.process_utils import get_environment_variable_setter_command
+
+
+class AtomState(str, Enum):
+    NOT_STARTED = 'NOT_STARTED'
+    IN_PROGRESS = 'IN_PROGRESS'
+    COMPLETED = 'COMPLETED'
 
 
 class Atom(object):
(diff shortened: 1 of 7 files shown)


boxcla:

app/master/subjob.py
```diff
@@ -31,8 +32,25 @@ def __init__(self, build_id, subjob_id, project_type, job_config, atoms):
         self.project_type = project_type
         self.job_config = job_config
         self._atoms = atoms
+        self._set_atom_state(AtomState.NOT_STARTED)
         self.timings = {}  # a dict, atom_ids are the keys and seconds are the values
 
+    def _set_atom_state(self, state):
```
tjlee0909: Could you add pydocs here please?

tjlee0909:
:+1: one minor comment, but LGTM

app/master/atom.py
```diff
@@ -16,12 +25,14 @@ def __init__(
         :type expected_time: float | None
         :type actual_time: float | None
         :type exit_code: int | None
+        :type state: `:class:AtomState` | None
```
josephharrington: Since we don't use Sphinx, you should just be able to do `AtomState` here (and actually IntelliJ should detect that syntax and use it for type hinting). We don't use the `:class:` syntax anywhere else in the app.

PR #215, opened 2015-07-05
Record build lifecycle timestamps and add to API

josephharrington:
Start recording timestamps for various phases of a build. The timestamps
are stored in the Build object instance, and exposed via the build's API
representation.

This timestamp data will enable both dashboard improvements and a build
timeout feature.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -1,7 +1,9 @@
+from collections import OrderedDict
 from enum import Enum
 import os
 from queue import Queue, Empty
 from threading import Lock
+import time
 import uuid
 
 from app.master.build_artifact import BuildArtifact
@@ -57,6 +59,10 @@ class Build(object):
         self._postbuild_tasks_are_finished = False
         self._timing_file_path = None
 
+        self._state_timestamps = {status: None
+                                  for status in BuildStatus}   # initialize all timestamps to None
+        self._record_state_timestamp(BuildStatus.QUEUED)
+
     def api_representation(self):
         return {
             'id': self._build_id,
@@ -69,6 +75,11 @@ class Build(object):
             'failed_atoms': self._failed_atoms(),  # todo: print the file contents instead of paths
             'result': self._result(),
             'request_params': self.build_request.build_parameters(),
+            # Convert self._state_timestamps to OrderedDict to make raw API response more readable. Sort the entries
+            # by numerically increasing dict value, with None values sorting highest.
+            'state_timestamps': OrderedDict(sorted(
+                [(state.lower(), timestamp) for state, timestamp in self._state_timestamps.items()],
+                key=lambda item: item[1] or float('inf'))),
         }
 
     def generate_project_type(self):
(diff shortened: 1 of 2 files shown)


boxcla:

test/unit/master/test_build.py
```diff
@@ -350,13 +350,134 @@ def test_generate_project_type_raises_error_if_failed_to_generate_project(self):
...
+        build = self._create_test_build(BuildStatus.BUILDING)
+
+        self.assertIsNone(build.get_state_timestamp(BuildStatus.CANCELED),
+                          '"canceled" timestamp should not be set unless build is canceled.')
+
+        build.cancel()
+        self.assertIsNotNone(build.get_state_timestamp(BuildStatus.CANCELED),
+                             '"canceled" timestamp should be set when build is canceled.')
+
+    # todo: almost all tests in this file should use this helper.
```
josephharrington: I'll refactor the rest of the tests in this file to use the added helper code in a separate commit.

app/master/build.py
```diff
@@ -58,6 +60,10 @@ def __init__(self, build_request):
         self._postbuild_tasks_are_finished = False
         self._timing_file_path = None
 
+        self._state_timestamps = {status.lower(): None  # initialize all timestamps to None
```
tjlee0909: Why the .lower()?

app/master/build.py
```diff
@@ -480,6 +500,30 @@ def _generate_unique_symlink_path_for_build_repo(self):
         """
         return os.path.join(Configuration['build_symlink_directory'], str(uuid.uuid4()))
 
+    def get_state_timestamp(self, build_status):
+        """
+        Get the recorded timestamp for a given build status. This may be None if the build has not yet reached
+        an appropriate state.
```
tjlee0909: nitpick: "the specified state."

app/master/build.py
```diff
@@ -99,7 +110,8 @@ def prepare(self, subjobs, job_config):
         :type subjobs: list[Subjob]
         :type job_config: JobConfig
         """
-        if self.project_type is None:
+        self._record_state_timestamp(BuildStatus.PREPARING)
```
tjlee0909: This wouldn't take into account the time it takes to fetch the project. Was that intentional? If you wanted to include the fetch times, we'd have to record this event in the build_request_handler in _build_preparation_loop().

josephharrington: Ha! Good catch -- I definitely meant to include the fetch time. That's more complicated then; I think I'll just remove `PREPARING` for now and save that for a future change.

app/master/build.py
josephharrington: Really this only matters for the api response -- I wanted those field names to be lowercase. But yeah, the keys in the dict don't have to be lowercased. I'll see if it would be simpler just to lowercase for the api response.

tjlee0909: If you had a reason, that's fine, I just wanted to know.

app/master/build.py
josephharrington: Done.

PR #216, opened 2015-07-05
Centralize functionality for calculating the build artifact path

tjlee0909:
Right now we perform artifact/result directory calculations, and each time we re-implement some form of os.path.join(). This commit consolidates the disparate methods into a single static function attached to the BuildArtifact class.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -6,7 +6,6 @@ import uuid
 
 from app.master.build_artifact import BuildArtifact
 from app.master.slave import SlaveMarkedForShutdownError
-from app.master.subjob import Subjob
 from app.util import analytics, util
 from app.util.conf.configuration import Configuration
 from app.util.counter import Counter
@@ -229,19 +228,16 @@ class Build(object):
             self.mark_failed('Error occurred while completing subjob {}.'.format(subjob_id))
             raise
 
-    def _get_atom_artifact_sys_path(self, subjob, atom_id):
-        return os.path.join(
-            self._build_results_dir(),
-            subjob.get_atom_artifact_name(atom_id),
-        )
-
     def _parse_payload_for_atom_exit_code(self, subjob_id):
         subjob = self.subjob(subjob_id)
         for atom_id in range(len(subjob.atoms)):
-            atom_exit_code_file_sys_path = os.path.join(
-                self._get_atom_artifact_sys_path(subjob, atom_id),
-                Subjob.EXIT_CODE_FILE,
+            artifact_dir = BuildArtifact.atom_artifact_directory(
+                self.build_id(),
+                subjob.subjob_id(),
+                atom_id,
+                result_root=Configuration['results_directory']
             )
+            atom_exit_code_file_sys_path = os.path.join(artifact_dir, BuildArtifact.EXIT_CODE_FILE)
             with open(atom_exit_code_file_sys_path, 'r') as atom_exit_code_file:
                 subjob.atoms[atom_id].exit_code = int(atom_exit_code_file.read())
 
@@ -251,9 +247,7 @@ class Build(object):
             return
 
         # Assertion: all payloads received from subjobs are uniquely named.
-        result_file_path = os.path.join(
-            self._build_results_dir(),
-            payload['filename'])
+        result_file_path = os.path.join(self._build_results_dir(), payload['filename'])
 
         try:
             app.util.fs.write_file(payload['body'], result_file_path)
@@ -469,10 +463,7 @@ class Build(object):
         self._artifacts_archive_file = app.util.fs.compress_directory(self._build_results_dir(), 'results.tar.gz')
 
     def _build_results_dir(self):
-        return os.path.join(
-            Configuration['results_directory'],
-            str(self.build_id()),
-        )
+        return BuildArtifact.build_artifact_directory(self.build_id(), result_root=Configuration['results_directory'])
 
     def _generate_unique_symlink_path_for_build_repo(self):
         """
--- a/app/master/build_artifact.py
+++ b/app/master/build_artifact.py
@@ -1,13 +1,18 @@
 import json
 import os
 
-from app.master.subjob import Subjob
+from app.util.conf.configuration import Configuration
 import app.util.fs
 from app.util.log import get_logger
 
 
 class BuildArtifact(object):
     ATOM_ARTIFACT_DIR_PREFIX = 'artifact_'
+    ATOM_DIR_FORMAT = ATOM_ARTIFACT_DIR_PREFIX + '{}_{}'
+    COMMAND_FILE = 'clusterrunner_command'
+    EXIT_CODE_FILE = 'clusterrunner_exit_code'
+    OUTPUT_FILE = 'clusterrunner_console_output'
+    TIMING_FILE = 'clusterrunner_time'
 
     def __init__(self, build_artifact_dir):
         """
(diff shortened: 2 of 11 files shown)


boxcla:

app/master/build.py
```diff
@@ -229,19 +228,16 @@ def complete_subjob(self, subjob_id, payload=None):
...
-        )
-
     def _parse_payload_for_atom_exit_code(self, subjob_id):
         subjob = self.subjob(subjob_id)
         for atom_id in range(len(subjob.atoms)):
-            atom_exit_code_file_sys_path = os.path.join(
-                self._get_atom_artifact_sys_path(subjob, atom_id),
-                Subjob.EXIT_CODE_FILE,
+            artifact_dir = BuildArtifact.artifact_directory(
+                self.build_id(),
```
aptxkid: nitpicking. kwargs for the first three args as well.

app/master/build_artifact.py
```diff
@@ -1,13 +1,18 @@
...
 
-from app.master.subjob import Subjob
+from app.util.conf.configuration import Configuration
 import app.util.fs
 from app.util.log import get_logger
 
 
 class BuildArtifact(object):
     ATOM_ARTIFACT_DIR_PREFIX = 'artifact_'
+    ATOM_DIR_FORMAT = "artifact_{}_{}"
```
aptxkid: Nitpicking. How about `ATOM_ARTIFACT_DIR_PREFIX +'_{}_{}'`?

app/master/build_artifact.py
```diff
@@ -103,3 +110,28 @@ def _write_timing_data_to_file(self, path, timing_data):
...
         """
         app.util.fs.write_file(json.dumps(timing_data), path)
+
+    @staticmethod
+    def artifact_directory(build_id, subjob_id=None, atom_id=None, result_root=None):
+        """
+        To get the full path to an atom artifact id, the caller must specify all id's: build_id, subjob_id, atom_id.
+        To get the path to just the build artifact directory, the caller must only specify the build_id.
+
+        If the caller specifies exactly one of the subjob_id or atom_id, it is a fatal error, and this method
```
aptxkid: How about splitting the method into two since looks like there're two use cases. Of course, underneath, they can share the same private method for implementation.

app/master/build.py
tjlee0909: done

app/master/build_artifact.py
tjlee0909: done

app/master/build_artifact.py
tjlee0909: sounds good

app/master/subjob.py
```diff
@@ -103,29 +98,6 @@ def atomic_commands(self):
         job_command = self.job_config.command
         return ['{} {}'.format(atom.command_string, job_command) for atom in self._atoms]
 
-    def get_atom_artifact_name(self, atom_id):
```
aptxkid: How you think about keeping some of the helper methods of subjob but call the static method as implementation? Keeping the code DRY is awesome. But calling static methods all the time does seem to be very OO.

app/slave/subjob_executor.py
```diff
@@ -2,7 +2,8 @@
 import shutil
 import time
 
-from app.master.subjob import Subjob
+from app.master.build import BuildArtifact
```
aptxkid: Nitpicking (again). It's kind of weird to import packages from `master` for the slave service. How about moving BuildArtifact to someplace else?

app/master/build.py
```diff
@@ -229,19 +228,16 @@ def complete_subjob(self, subjob_id, payload=None):
...
+                subjob.subjob_id(),
+                atom_id,
+                result_root=Configuration['results_directory']
             )
+            atom_exit_code_file_sys_path = os.path.join(artifact_dir, BuildArtifact.EXIT_CODE_FILE)
```
aptxkid: Here you're calling a static method of `BuildArtifact` and immediately join it with `BuildArtifact.EXIT_CODE_FILE`. How you think about making some helpers so few components actually need to know `BuildArtifact.EXIT_CODE_FILE`?

aptxkid:
LGTM with (super) minor comments. :+1: 
Sorry for tons of nitpicking. Getting rid of all the duplicates is awesome!

app/master/subjob.py
tjlee0909: I think static is only bad when used incorrectly, and otherwise is better than the alternative. If you can give me a good reason why this is not a good use case for it, I'm happy to change it.

app/master/build.py
tjlee0909: I was thinking about that (having five static methods, four for each of the files, and one for the directory) and decided against it, just by feeling. But I'm thinking that might be worth it if it means the calling code doesn't have to worry about enum's from other classes.

app/slave/subjob_executor.py
tjlee0909: Yup, was thinking about whether to do that in this PR or another one. But might as well do it in this one.

app/master/build.py
tjlee0909: The first argument is a required one, and the last three are optional kwargs. Does it make sense to leave the kwarg for the first one out? I'm not familiar with the Python convention.

app/master/subjob.py
aptxkid: E.g. the following `artifact_dir` method it has the information about `build_id`, 'subjob_id'. Given an instance of a `Subjob`, it's just easier to get the artifact_dir by calling a method with `atom_id` than calling a static method with `build_id`, etc. The code can stay dry as you can call the static method as the implementation of `artifact_dir` here. This might not be a good example. But the main thing I want to share is that if the object already has the information, asking the consumer of the object to fetch the property from the object (`build_id`, etc.) and then call a static method is a little more work than just using the internal properties and call a method on the object instead.

app/slave/subjob_executor.py
aptxkid: I am perfectly fine with a separate PR. I am fan of keeping each PR small.

app/master/build.py
aptxkid: I actually dunno if there's any convention here. I 'personally' like kwargs more as they're more explicit and usually more readable (especially when dealing with many arguments).

app/master/subjob.py
tjlee0909: That's a good point.

On a related note, would you say that it's not worth creating the extra static helper methods to generate the timing file/console output file/exit code file? Because it wouldn't be used in the scenario that we keep the instance method.

app/master/build.py
tjlee0909: How about when the argument names speak for themselves? This is what the code looks like now:

build_id, subjob_id=subjob_id, atom_id=atom_id, result_root=result_root

app/master/build.py
aptxkid: I think it's impossible before this refactor. But it might be possible now to entirely encapsulate those constants, which would be awesome since usually it's better if we can expose fewer things. But I agree that making many static methods might be weird also. Just some thoughts. We don't need to make these changes right now or ever.

app/slave/subjob_executor.py
josephharrington: +1 to everything (moving `BuildArtifact` in separate PR)

test/unit/slave/test_subjob_executor.py
```diff
@@ -47,14 +49,14 @@ def test_execute_subjob_passes_correct_build_executor_index_to_execute_command_i
         atomic_commands = ['command']
         executor.id = 2
         expected_env_vars = {
-            'ARTIFACT_DIR': 'path',
+            'ARTIFACT_DIR': '/tmp/1/artifact_2_0',
```
josephharrington: @aptxkid will this be Windows-incompatible?

josephharrington:
I want to double check with Lu that the tests you added/modified will be compatible with Windows, but otherwise looks good. :+1:

test/unit/slave/test_subjob_executor.py
aptxkid: Almost certainly no. Usually, I would do
`Configuration['artifact_directory'] = os.path.expanduser('~')`

`'ARTIFACT_DIR': join(Configuration['artifact_directory'], '1', 'artifact_2_0')`.

Thanks for catching it, Joey!

tjlee0909: Done, and will change in test_build_artifact.py

PR #218, opened 2015-07-07
Implement console endpoint on slave

tjlee0909:
This entails implementing the console_output class, which is a utility class for getting segments of console output from a file.

### DIFF
--- a/app/common/console_output.py
+++ b/app/common/console_output.py
@@ -1,0 +1,101 @@
+from app.common.console_output_segment import ConsoleOutputSegment
+
+
+class ConsoleOutput(object):
+    """
+    Represents the console output of an atom.
+    """
+    def __init__(self, path):
+        """
+        :param path: sys path to the console output file
+        :type path: str
+        """
+        self.path = path
+
+    def segment(self, max_lines=50, offset_line=None):
+        """
+        Return a segment of the console output.
+
+        :type max_lines: int
+        :param offset_line: The starting line number in the console output from which the segment should
+            return content for. If set to None, then reads content starting from the end.
+        :type offset_line: int | None
+        :rtype: ConsoleOutputSegment
+        """
+        if offset_line is None:
+            return self._parse_from_end(max_lines)
+        else:
+            return self._parse_from_offset(max_lines, offset_line)
+
+    def _parse_from_offset(self, max_lines, offset_line):
+        """
+        Return up to max_lines of output starting from line number offset_line. If max_lines + offset_line
+        doesn't exist on this console output, then return everything beyond offset_line.
+
+        :type max_lines: int
+        :type offset_line: int
+        :rtype: ConsoleOutputSegment
+        """
+        total_lines = 0
+        output_lines = 0
+        console_output = []
+
+        with open(self.path, 'r') as f:
+            # Iterate up to the index offset_line
+            for i in range(0, offset_line):
+                # This is an error, meaning that there aren't even offset_line+1 lines in self.path.
+                if f.readline() == '':
+                    raise ValueError('offset: {} is higher than the total number of lines in file {}'.format(
+                        offset_line, self.path))
+
+                total_lines += 1
+
+            # Retrieve the console_output just between offset_line and offset_line + max_lines
+            for i in range(offset_line, offset_line + max_lines):
+                line = f.readline()
+
+                # We have reached the end of the file, or a line that has not finished being written to.
+                if line == '' or not line.endswith("\n"):
+                    break
+
+                console_output.append(line)
+                output_lines += 1
+                total_lines += 1
+
+            # If there are more lines, then keep on counting in order to populate total_lines properly
+            while f.readline():
+                total_lines += 1
+
+        return ConsoleOutputSegment(offset_line, output_lines, total_lines, ''.join(console_output))
+
+    def _parse_from_end(self, max_lines):
+        """
+        Return console output segment containing the last max_lines of output, or the whole file
+        if less than or equal to max_lines.
+
+        :type max_lines: int
+        :rtype: ConsoleOutputSegment
+        """
+        total_lines = 0
+        output_lines = 0
+        console_output = []
+
+        with open(self.path, 'r') as f:
+            # Figure out how many lines there are, so that we know which offset line to start appending
+            # console output from.
+            for line in f:
+                if line.endswith("\n"):
+                    total_lines += 1
+
+            f.seek(0)
+            offset_line = max(total_lines - max_lines, 0)
+
+            # Move file pointer to the desired offset_line
+            for i in range(offset_line):
+                f.readline()
+
+            for i in range(offset_line, total_lines):
+                console_output.append(f.readline())
+                output_lines += 1
+
+        return ConsoleOutputSegment(offset_line, output_lines, total_lines, ''.join(console_output))
--- a/app/slave/cluster_slave.py
+++ b/app/slave/cluster_slave.py
@@ -1,13 +1,16 @@
 from enum import Enum
+import os
 from queue import Queue
 import requests
 import sys
 import time
 
+from app.common.console_output import ConsoleOutput
+from app.master.build_artifact import BuildArtifact
 from app.project_type.project_type import SetupFailureError
 from app.slave.subjob_executor import SubjobExecutor
 from app.util import analytics, log, util
-from app.util.exceptions import BadRequestError
+from app.util.exceptions import BadRequestError, ItemNotFoundError
 from app.util.network import Network
 from app.util.safe_thread import SafeThread
 from app.util.secret import Secret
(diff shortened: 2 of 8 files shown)


boxcla:

app/web_framework/cluster_slave_application.py
```diff
@@ -116,6 +126,82 @@ def get(self, build_id, subjob_id):
...
+            }
+        self.write(response)
+
+
+class _AtomHandler(_ClusterSlaveBaseAPIHandler):
+    def get(self, build_id, subjob_id, atom_id):
+        response = {
+            'comment': 'not implemented',
+            }
+        self.write(response)
```
tjlee0909: I wasn't quite sure what to put here for these two endpoints (_AtomsHandler and _AtomHandler) that do nothing, so I just copied what we have for these types of non-implemented endpoints.

app/common/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'taejun'
```
josephharrington: :dollar: :dollar: :tophat: :dollar: :dollar:

aptxkid: @tjlee0909  You can change this in IntelliJ settings

tjlee0909: Dope

tjlee0909:
Dangit, I just realized that in PHPUnit, the output looks like: 

PHPUnit 4.0.12 by Sebastian Bergmann.

.....

And if we read output by lines, and not by bytes, then the streaming console output will not show the progressing dots (because they are all on the same line, and we have already requested that line in a previous request).

app/web_framework/cluster_slave_application.py
```diff
@@ -116,6 +126,82 @@ def get(self, build_id, subjob_id):
...
+            response = {'message': '\'offset_line\' must be greater than or equal to zero.'}
+            self._write_status(response, success=False, status_code=400)
+            return
+        if max_lines <= 0:
+            response = {'message': '\'max_lines\' must be greater than zero.'}
+            self._write_status(response, success=False, status_code=400)
+            return
+
+        artifact_dir = BuildArtifact.atom_artifact_directory(build_id, subjob_id, atom_id)
+        output_file = os.path.join(artifact_dir, BuildArtifact.OUTPUT_FILE)
```
josephharrington: How do you feel about moving most of this logic in this method into `ClusterSlave`? Originally the `ClusterMasterApplication` and `ClusterSlaveApplication` classes were intended to just be thin wrappers around `ClusterMaster` and `ClusterSlave`. (The original motivation was that we should ideally be able to swap out the web framework layer without affecting underlying app behavior.)

We've moved away from that a bit and some business logic has crept in to the web framework layer, but I think it's still a good ideal to shoot for.

Something you could also take advantage of is from `ClusterSlave` raise a `BadRequestError` with the error message and the web framework layer will automatically create the `{'message': 'whatever'}` response instead of you needing to construct it manually here.

app/web_framework/cluster_slave_application.py
josephharrington: Could you just do a `pass` similar to `_SubjobsHandler` above? I'm not sure what the reasoning behind the original "not implemented" response was.

app/web_framework/cluster_slave_application.py
```diff
@@ -116,6 +126,82 @@ def get(self, build_id, subjob_id):
...
+
+        # If the console output file doesn't exist where we expect it to be, return a 404.
+        if not os.path.isfile(output_file):
+            response = {'message': 'Output file doesn\'t exist for build_id: {} subjob_id: {} atom_id: {}'.format(
+                build_id, subjob_id, atom_id)}
+            self._write_status(response, success=False, status_code=404)
```
josephharrington: Continuing from my comment above, a 404 can be set by raising an `ItemNotFoundError`. The full list is at the top of the `ClusterBaseHandler` class.

test/integration/common/test_console_output.py
```diff
@@ -0,0 +1,64 @@
+from genty import genty, genty_dataset
+import os
+from tempfile import mkstemp
+from unittest import TestCase
+
+from app.common.console_output import ConsoleOutput
+
+
+@genty
+class TestConsoleOutput(TestCase):
```
josephharrington: Should we create an (empty for now) `BaseIntegrationTestCase` class? I can see us definitely using that as we create more integration tests.

app/web_framework/cluster_slave_application.py
tjlee0909: Done

app/web_framework/cluster_slave_application.py
tjlee0909: That's cool, will use that.

app/web_framework/cluster_slave_application.py
tjlee0909: Sounds good.

test/integration/common/test_console_output.py
tjlee0909: OK

app/common/console_output.py
```diff
@@ -0,0 +1,102 @@
...
+        :param path: sys path to the console output file
+        :type path: str
+        """
+        self.path = path
+
+    def segment(self, max_lines=50, offset_line=None):
+        """
+        Return a segment of the console output.
+
+        :param max_lines:
```
josephharrington: It's pretty obvious what this is supposed to be but did you mean to add a docstring for the param?

tjlee0909: Nah I'll delete it.

app/common/console_output.py
```diff
@@ -0,0 +1,102 @@
...
+                        offset_line, self.path))
+
+                total_lines += 1
+
+            # Retrieve the console_output just between offset_line and offset_line + max_lines
+            for i in range(offset_line, offset_line + max_lines):
+                line = f.readline()
+
+                # We have reached the end of the file, or a line that has not finished being written to.
+                if line == '' or not line.endswith("\n"):
```
josephharrington: Just curious -- is there an easy way to manually test the case where a line has not finished being written to? I tried testing a one-line file (e.g., `echo something > oneline.txt`) but Python still sticks a newline on the end.

test/integration/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'taejun'
```
tjlee0909: DOPE

app/common/console_output.py
tjlee0909: I modified a PHPUnit test to perform a "sleep(60)" on the third test case, and did this:

phpunit test/unit/lib/Jira_Test.php > whatever.txt

and then cntrl-c'd out of it a few seconds into it.
```
taejun@mbp-002672 devtools (dev) $ cat whatever.txt 
PHPUnit 4.0.12 by Sebastian Bergmann.

Configuration read from /Users/taejun/Code/devtools/phpunit.xml

..taejun@mbp-002672 devtools (dev) $ 
```

and I see here that the terminal prompt is on the same line as the cat output.

app/common/console_output.py
```diff
@@ -0,0 +1,101 @@
...
+        :rtype: ConsoleOutputSegment
+        """
+        total_lines = 0
+        output_lines = 0
+        console_output = []
+
+        with open(self.path, 'r') as f:
+            # Figure out how many lines there are, so that we know which offset line to start appending
+            # console output from.
+            for line in iter(f.readline, ''):
```
josephharrington: I think this is equivalent to
```python
for line in f:
    ...
```

tjlee0909:
Aw, the coveralls doesn't include my integration test and assumes that I lowered the coverage percentage.

app/common/console_output.py
```diff
@@ -0,0 +1,101 @@
...
+                if line.endswith("\n"):
+                    total_lines += 1
+
+            f.seek(0)
+            offset_line = max(total_lines - max_lines, 0)
+
+            # Move file pointer to the desired offset_line
```
josephharrington: Just curious -- do you see any downside to calling `_parse_from_offset` at this point?

tjlee0909: I was considering that. The only downside is that we close and reopen the file pointer, but that's not much.

josephharrington:
@tjlee0909 if you change the `test-unit-with-coverage` target in the Makefile to also run the integration test directory it should include it in the coverage.

app/slave/cluster_slave.py
```diff
@@ -324,6 +327,55 @@ def kill(self):
...
+        except ValueError as e:
+            raise BadRequestError(e)
+
+        return {
+            'offset_line': segment.offset_line,
+            'num_lines': segment.num_lines,
+            'total_num_lines': segment.total_num_lines,
+            'content': segment.content,
+        }
+
```
josephharrington: This is pretty clean and great. :smiley_cat:

test/integration/common/test_console_output.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self.assertEquals(segment.num_lines, expected_num_lines)
+        self.assertEquals(segment.offset_line, expected_offset_line)
+        self.assertEquals(segment.total_num_lines, expected_total_num_lines)
+        self.assertEquals(segment.content, expected_content)
+
+    def test_segment_raises_value_error_if_offset_greater_than_total_length(self):
+        with self.assertRaises(ValueError):
+            console_output = ConsoleOutput(self._completed_console_output_file_path)
+            console_output.segment(max_lines=5, offset_line=155)
```
josephharrington: I remember we ran into some weird issues with invalid unicode coming out of phpunit output that caused some exceptions. (We have this test around the issue: https://github.com/box/ClusterRunner/blob/master/test/unit/project_type/test_project_type.py#L78) I wonder if we should add some tests for this new functionality related to unicode console output and potentially malformed output.

I think we're a bit safer in this case because we're handling a client's API call when loading the file, so the worst that could happen is returning a 500 to the client.

app/common/console_output.py
josephharrington: Oh derp, yeah that makes sense. I forgot that echo adds a newline by default. `echo -n something > oneline.txt` works.

josephharrington:
Looks great! I made a few comments, but overall I am happy. :smile: :+1:

app/common/console_output.py
tjlee0909: Done.

app/common/console_output.py
tjlee0909: I tried doing that, and for some reason it swapped back and forth between passing my integration tests and failing it. I've decided to keep this for now.

PR #219, opened 2015-07-07
Refactor and clean up tests for Build object

josephharrington:
The Build class is the largest single class and the most complex object
in the application. Future planned refactorings will help this, but in
the meantime it is hard and usually verbose to create realistic build
objects for testing.

This change refactors some of the Build tests to use a build creation
helper that creates more realistic build objects. It has a side benefit
of making these tests more readable and maintainable.

The tradeoff is that these tests are now more "integration-y", but it
makes sense to move in this direction considering the complexity of the
Build class itself.

### DIFF
--- a/test/unit/master/test_build.py
+++ b/test/unit/master/test_build.py
@@ -2,7 +2,8 @@ from queue import Queue
 from os.path import abspath, join
 import sys
 from threading import Event
-from unittest.mock import MagicMock, Mock, mock_open
+from unittest.mock import MagicMock, Mock, mock_open, call
+from genty import genty, genty_dataset
 
 from app.master.atom import Atom, AtomState
 from app.master.atomizer import Atomizer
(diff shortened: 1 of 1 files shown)


boxcla:

test/unit/master/test_build.py
```diff
@@ -17,230 +18,178 @@
...
-        build._project_type = mock_project_type
+        build = self._create_test_build(BuildStatus.QUEUED)
+        job_config = self._create_job_config()
+        subjobs = self._create_subjobs(count=3, job_config=job_config)
 
-        build.prepare(subjobs, self._create_job_config())
+        build.prepare(subjobs=subjobs, job_config=job_config)
 
-        with self.assertRaises(Exception):
-            build.prepare(subjobs, mock_project_type, self._create_job_config())
```
josephharrington: I think this was actually not testing this correctly before -- the exception that was being raised was around the extra argument to `prepare()`, not the one we were trying to test for.

josephharrington:
This is part 1 of 2. I'm going to refactor the rest of the tests in a separate PR after this one is accepted and merged.

tjlee0909:
I've only looked at two of the tests you modified, as well as your changes to _create_test_build. LGTM :+1:

PR #220, opened 2015-07-07
Call Popen with a cmd sequence instead of command string

aptxkid:
According to https://docs.python.org/3.4/library/subprocess.html#popen-constructor, this is the recommanded way to make cross-platform Popen objects. The way we used to pass command strings to Popen doesn't work well on Windows if the command includes spaces.

### DIFF
--- a/app/client/service_runner.py
+++ b/app/client/service_runner.py
@@ -34,9 +34,7 @@ class ServiceRunner(object):
         self._logger.info('Running master on {}'.format(self._master_url))
         if self.is_master_up():
             return
-        cmd = '{} master --port {} &'.format(
-            self._main_executable,
-            self._port(self._master_url))
+        cmd = [self._main_executable, 'master', '--port', self._port(self._master_url)]
 
         self._run_service(cmd, self._master_url)
 
@@ -56,12 +54,10 @@ class ServiceRunner(object):
         :return:
         """
         self._logger.info('Running slave')
-        cmd = '{} slave --master-url {} '.format(
-            self._main_executable,
-            self._master_url)
+        cmd = [self._main_executable, 'slave', '--master-url', self._master_url]
         if port is not None:
-            cmd += '--port {} '.format(str(port))
-        cmd += '&'
+            cmd.extend(['--port', str(port)])
+
         self._run_service(cmd)
 
     def block_until_build_queue_empty(self, timeout=60):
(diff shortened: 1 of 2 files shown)


boxcla:

app/client/service_runner.py
```diff
@@ -34,9 +34,7 @@ def run_master(self):
         self._logger.info('Running master on {}'.format(self._master_url))
         if self.is_master_up():
             return
-        cmd = '{} master --port {} &'.format(
-            self._main_executable,
-            self._port(self._master_url))
+        cmd = [self._main_executable, 'master', '--port', self._port(self._master_url)]
```
tjlee0909: I don't know what the Window's equivalent of backgrounding the process is, but here you're missing the '&' argument.

app/client/service_runner.py
```diff
@@ -88,7 +85,7 @@ def is_queue_empty():
             raise Exception('Master service did not become idle before timeout.')
 
     def kill(self):
-        self._run_service('{} stop'.format(self._main_executable))
+        self._run_service([self._main_executable, 'stop'])
 
     def _run_service(self, cmd, service_url=None):
```
tjlee0909: I'm not familiar with this Popen_with_delayed_expansion(), but I'm assuming that you can interchange a string with a list of strings?

aptxkid: I don't quite understand this. Are you suggesting break the string into cmd sequence in Popen_with_delayed_expansion instead of here?

tjlee0909: No, I was asking whether it was okay to pass in a list of strings, instead of a string as was there previously.

app/client/service_runner.py
aptxkid: Good catch! I apparently missed it and duplicate the problem in test as well!

josephharrington:
Should we change `Popen_with_delayed_expansion` to refuse string commands?

aptxkid:
@josephharrington Actually, I just found out that when `shell=True` is set, passing command as a string is recomanded. https://docs.python.org/3.4/library/subprocess.html#popen-constructor

app/client/service_runner.py
aptxkid: I see. It's just a wrapper around `subprocess.Popen` which has the exact same interface as Popen. And the only thing it does is to ensure all environment variables in the command will be expanded during execution time. https://docs.python.org/3.4/library/subprocess.html#popen-constructor

app/client/service_runner.py
aptxkid: Actually I am going to remove '&' here since it doesn't seem matter here and '&' is not cross-platform.

PR #221, opened 2015-07-08
Implement the console output endpoint on the master.

tjlee0909:
This PR contains mostly refactorings in order to keep DRY for the small amount of code needed to implement the console output endpoint on the master.

### DIFF
--- a/app/common/cluster_service.py
+++ b/app/common/cluster_service.py
@@ -1,0 +1,62 @@
+import os
+
+from app.common.console_output import ConsoleOutput
+from app.master.build_artifact import BuildArtifact
+from app.util.exceptions import BadRequestError, ItemNotFoundError
+
+
+class ClusterService(object):
+    """
+    The abstract base ClusterRunner service class.
+    """
+
+    def get_console_output(self, build_id, subjob_id, atom_id, result_root, max_lines=50, offset_line=None):
+        """
+        Return the console output if it exists, raises an ItemNotFound error if not.
+
+        On success, the response contains keys: offset_line, num_lines, total_num_lines, and content.
+
+        e.g.:
+        {
+            'offset_line': 0,
+            'num_lines': 50,
+            'total_num_lines': 167,
+            'content': 'Lorem ipsum dolor sit amet,\nconsectetur adipiscing elit,\n...',
+        }
+
+        :type build_id: int
+        :type subjob_id: int
+        :type atom_id: int
+        :param result_root: the sys path to either the results or artifacts directory where results are stored.
+        :type result_root: str
+        :param max_lines: The maximum total number of lines to return. If this max_lines + offset_line lines do not
+            exist in the output file, just return what there is.
+        :type max_lines: int
+        :param offset_line: The line number (0-indexed) to start reading content for. If none is specified, we will
+            return the console output starting from the end of the file.
+        :type offset_line: int | None
+        """
+        if offset_line is not None and offset_line < 0:
+            raise BadRequestError('\'offset_line\' must be greater than or equal to zero.')
+        if max_lines <= 0:
+            raise BadRequestError('\'max_lines\' must be greater than zero.')
+
+        artifact_dir = BuildArtifact.atom_artifact_directory(build_id, subjob_id, atom_id, result_root=result_root)
+        output_file = os.path.join(artifact_dir, BuildArtifact.OUTPUT_FILE)
+
+        if not os.path.isfile(output_file):
+            raise ItemNotFoundError('Output file doesn\'t exist for build_id: {} subjob_id: {} atom_id: {}'.format(
+                build_id, subjob_id, atom_id))
+
+        try:
+            console_output = ConsoleOutput(output_file)
+            segment = console_output.segment(max_lines, offset_line)
+        except ValueError as e:
+            raise BadRequestError(e)
+
+        return {
+            'offset_line': segment.offset_line,
+            'num_lines': segment.num_lines,
+            'total_num_lines': segment.total_num_lines,
+            'content': segment.content,
+        }
--- a/app/master/subjob.py
+++ b/app/master/subjob.py
@@ -29,6 +29,7 @@ class Subjob(object):
         self._atoms = atoms
         self._set_atom_state(AtomState.NOT_STARTED)
         self.timings = {}  # a dict, atom_ids are the keys and seconds are the values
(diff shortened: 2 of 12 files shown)


boxcla:

app/web_framework/cluster_master_application.py
```diff
@@ -152,6 +156,51 @@ def get(self, build_id, subjob_id, atom_id):
         self.write(response)
 
 
+class _AtomConsoleHandler(_ClusterMasterBaseAPIHandler):
```
tjlee0909: @josephharrington - I wasn't sure whether this was appropriate or not to put inside the ClusterMaster class, vs outside.

The whole 'except' clause in the try/except contains a redirect, which I wasn't sure if we wanted to keep out of the ClusterMaster class and contained within the web framework code (ClusterMasterApplication).

app/web_framework/cluster_master_application.py
```diff
@@ -152,6 +156,51 @@ def get(self, build_id, subjob_id, atom_id):
...
+            if slave is None:
+                raise e
+
+            api_url_builder = UrlBuilder(slave.url)
+            slave_console_url = api_url_builder.url(
+                'build/{}/subjob/{}/atom/{}/console'.format(build_id, subjob_id, atom_id))
+            slave_console_url += '?maxLines={}'.format(max_lines)
+
+            if offset_line is not None:
+                slave_console_url += '&offsetLine={}'.format(offset_line)
```
tjlee0909: I'm not familiar with redirects, or Tornado's utility functions for them. I basically wanted to forward the exact same request (with querystring args). I might play around with what tornado returns and replace these 7 lines with something built into Tornado's HTTPRequest object.

app/common/cluster_service.py
```diff
@@ -0,0 +1,62 @@
+import os
```
tjlee0909: This whole class is just a copy/paste of a method that was previously in ClusterSlave.

app/web_framework/cluster_master_application.py
```diff
@@ -152,6 +156,51 @@ def get(self, build_id, subjob_id, atom_id):
...
+            build = self._cluster_master.get_build(int(build_id))
+            subjob = build.subjob(int(subjob_id))
+            slave = subjob.slave
+
+            if slave is None:
+                raise e
+
+            api_url_builder = UrlBuilder(slave.url)
+            slave_console_url = api_url_builder.url(
+                'build/{}/subjob/{}/atom/{}/console'.format(build_id, subjob_id, atom_id))
```
josephharrington: I think you can do something like
```python
api_url_builder.url('build', build_id, 'subjob', subjob_id, 'atom', atom_id, 'console')
```

tjlee0909: I'm afraid you might have commented on an outdated diff sir.

app/web_framework/cluster_master_application.py
josephharrington: I think `urllib.parse.urlencode` will turn a dict into a query string.

Something like
```python
query = {'maxLines': max_lines}
if offset_line is not None:
    query['offsetLine'] = offset_line

query_string = urlencode(query)
```

josephharrington: Also, this is an annoying comment but I just noticed that our only other existing query parameters (for master and slave /eventlog) use underscores instead of camelCase.

app/web_framework/cluster_master_application.py
josephharrington: Ah, yeah I agree with you about placing this logic here -- it's all pretty webby. I don't really see much that could be moved into `ClusterMaster`. It makes sense though -- this is probably some of the heaviest web-centric logic that we've added so far.

app/master/subjob.py
```diff
@@ -40,11 +41,15 @@ def _set_atom_state(self, state):
         for atom in self._atoms:
             atom.state = state
 
-    def mark_in_progress(self):
+    def mark_in_progress(self, slave):
```
josephharrington: Do you think this is better vs just storing the slave_url? (Actual question -- I'm interested in your thinking. :smile:)

app/web_framework/cluster_master_application.py
tjlee0909: Ok, I'm switching to using underscores.

app/web_framework/cluster_master_application.py
tjlee0909: But yeah, going to use that.

app/master/subjob.py
tjlee0909: It's about the same I'd say. You never know in the future when we might need a back reference to the Slave for other stuff.

app/web_framework/cluster_slave_application.py
```diff
@@ -137,11 +137,23 @@ def get(self, build_id, subjob_id, atom_id):
...
+            build_id,
+            subjob_id,
+            atom_id,
+            Configuration['artifact_directory'],
+            max_lines,
+            offset_line
+        )
+        # Because the 'Origin' header in the redirect (from the master) gets set to 'null', the only way
+        # for the client to receive this response is for us to allow any origin to receive this response.
+        self.set_header('Access-Control-Allow-Origin', '*')
```
tjlee0909: I am not a security professional, so I do not know the implications of having this line here, but it seems to be the only thing that works with this redirect business.

josephharrington:
:+1:

app/web_framework/cluster_slave_application.py
josephharrington: This is probably fine for now. At worst, an attack would require a malicious website to make an ajax request to the internal cluster master's console endpoint and even in that unlikely event, all that would be compromised would be console output from a random subjob.

josephharrington: I double checked and the redirect response *does* need to have the correct allow-origin header set for the browser to follow it. This means that even if a malicious website tried to make an ajax request to our cluster master it would fail due to cross origin policy. An attack would need to be an malicious site making an ajax request to a cluster slave, which is even *less* likely to be successful since builds and subjobs are randomly distributed.

From reading around, I'm not really sure how we could do better than this. We could require the "Referer" header to match our allowed origins, but it sounds like users can legitimately disable referer headers as a security setting, so that approach would disable the redirect functionality for anyone with that security setting.

PR #223, opened 2015-07-10
Use is_windows helper in setup.py

aptxkid:
Instead of checking os.name, use is_windows helper to name the executable properly.


boxcla:

josephharrington:
:+1:

PR #227, opened 2015-07-13
Refactor and clean up tests for Build - part 2

josephharrington:
This is a continuation of #219 (commit d67c302).

### DIFF
--- a/app/master/slave.py
+++ b/app/master/slave.py
@@ -95,7 +95,6 @@ class Slave(object):
             raise SlaveMarkedForShutdownError('Tried to start a subjob on a slave in shutdown mode. ({}, id: {})'
                                               .format(self.url, self.id))
 
-        subjob.mark_in_progress(self)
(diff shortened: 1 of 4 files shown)


boxcla:

app/master/slave.py
```diff
@@ -95,7 +95,6 @@ def start_subjob(self, subjob):
             raise SlaveMarkedForShutdownError('Tried to start a subjob on a slave in shutdown mode. ({}, id: {})'
                                               .format(self.url, self.id))
 
-        subjob.mark_in_progress(self)
```
josephharrington: It made more sense to me to call this from `Build` instead of `Slave` but let me know if you disagree. The similar `subjob.mark_completed()` call is already in `Build` so it felt appropriate for the two `subjob.mark_x()` calls to be in the same place. I think in general it makes sense for the build object to do its own bookkeeping on the status of its subjobs/atoms.

PR #230, opened 2015-07-16
Add slave to subjob API response

josephharrington:
Previously the only way to find out where a subjob ran was to grep
the master logs. This will let us skip that step.


boxcla:

wjdhollow:
:+1:

PR #238, opened 2015-07-27
Add Hypothesis tests

drobertduke:
This is an experiment for adding/running hypothesis tests.  These
tests generate a variety of inputs for a function.

These tests exposed a failure where if a build_type is not
specified for a BuildRequest, the is_valid and is_valid_type
methods fail.


boxcla:

josephharrington:
Just curious -- what made you decide to put these in a separate test/hypothesis directory?

test/hypothesis/master/test_cluster_master.py
```diff
@@ -0,0 +1,26 @@
...
+class TestClusterMaster(BaseUnitTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.patch('app.util.fs.create_dir')
+        self.patch('app.util.fs.async_delete')
+        self.patch('os.makedirs')
+
+    @given(dictionaries(text(), text()))
+    def test_handle_request_for_new_build(self, build_params):
```
josephharrington: Are these tests just that the code doesn't raise an exception? If so, how do you feel about naming the tests something to indicate that -- something like `test_handle_request_for_new_build_does_not_raise_exception`.

drobertduke:
I put them in a different directory because I thought we might want to keep them separate from the existing unit tests, but now I think that was wrong.

josephharrington:
:+1:

Issue #191, opened 2015-06-23
clusterrunner_command file should include all env vars


josephharrington:
Here is an example of the current clusterrunner_command file:
```
export TESTPATH="/tmp/clusterrunner_build_symlinks/b42ce3c9-a9ff-40dd-bee0-09e990b9e775/test/unit/lib/Person_Test.php"; cd $PROJECT_DIR && echo "something important" > $ARTIFACT_DIR/coverage_config.xml && phpunit -c coverage_config.xml $TESTPATH
```
Note that the values of `$PROJECT_DIR` and `$ARTIFACT_DIR` are not included in this file, so if the user is debugging this command and would like to know what those values were then they will need to hunt through the clusterrunner master and slave logs to find all the env values.

I think it would be more useful if we also included the values of the ClusterRunner "standard" env vars (e.g., `$PROJECT_DIR`) in this file so that users could just execute the file to reproduce CR's behavior.

An example of what I mean for the above case would be:
```
export ATOM_ID="0";
export BUILD_EXECUTOR_INDEX="10";
export EXECUTOR_INDEX="0";
export MACHINE_EXECUTOR_INDEX="0";
export ARTIFACT_DIR="/home/jenkins/.clusterrunner/artifacts/2318/artifact_7_0";
export PROJECT_DIR="/tmp/clusterrunner_build_symlinks/b42ce3c9-a9ff-40dd-bee0-09e990b9e775"; 
export TESTPATH="/tmp/clusterrunner_build_symlinks/b42ce3c9-a9ff-40dd-bee0-09e990b9e775/test/unit/lib/Person_Test.php";

cd $PROJECT_DIR && echo "something important" > $ARTIFACT_DIR/coverage_config.xml && phpunit -c coverage_config.xml $TESTPATH
```

PR #241, opened 2015-07-28
Display the atom, not atom command, in api representation.

tjlee0909:
In displaying the streaming console output, the atom command is
difficult to read.

And as far as I can tell, we do not use the atom command (from the api
representation) anywhere.

### DIFF
--- a/app/master/atom.py
+++ b/app/master/atom.py
@@ -28,11 +28,11 @@ class Atom(object):
         :type state: `:class:AtomState` | None
         """
         self._env_var_name = env_var_name
-        self._atom_value = atom_value
+        self.atom_value = atom_value
         self.expected_time = expected_time
         self.actual_time = actual_time
         self.exit_code = exit_code
         self.state = state
 
         # Convert atomizer command output into environment variable export commands.
-        self.command_string = get_environment_variable_setter_command(self._env_var_name, self._atom_value)
+        self.command_string = get_environment_variable_setter_command(self._env_var_name, self.atom_value)
--- a/app/master/subjob.py
+++ b/app/master/subjob.py
@@ -75,7 +75,7 @@ class Subjob(object):
     def get_atoms(self):
         return [{
             'id': idx,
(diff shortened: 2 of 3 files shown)


boxcla:

josephharrington:
:+1: (ignore coverage failures)

josephharrington:
@tjlee0909 On second thought -- do you think we should expose the env var name in addition to the atom? I'm torn on this since it's a pretty rare use case to have more than one env var name in the atomizers, but it is supported.

For example, right now I have a job with two different atomizers defined. Before this change I could easily see which var name was getting set from the atom command, but now it just gives the value and not the var name.

Alternatively, we could extend the idea in #191 and also include all the env var values in the subjob API's `command` field.

tjlee0909:
I see, so we have a use case for that now.

One of the projects that I'd like to work on for next quarter is giving developers the ability to take a machine offline (from ClusterRunner, and or Jenkins), and allow them to re-run the test exactly the way it was before. In such a scenario, https://github.com/box/ClusterRunner/issues/191 would be a pre-req.

So, yes to include all the env var values in the subjob API's command field regardless. 

We COULD have two exposed options. One is atom_value (I guess technically, it'd be atom_values if you can have multiple atom variables), and the other could be atom_command. What do you think about that?

josephharrington:
For a single atom we still will only have one atom_key and one atom_value, but in a given build the atom_keys can be different from atom to atom. (This is because specifying multiple atomizers just appends more atoms with different env vars.)

This pattern still feels a bit weird to me because we just leave the other atomizer vars undefined when running the command. (Actually I just realized this could be a source of weird behavior if one of the undefined atomizer var names happened to be something that was already defined in the environment. I might make a separate issue ticket for that.)

I'd be fine with just adding something like an atom_key or atom_var field to the atom api. Aaaand I'd probably also be fine with just tabling this for now until we get around to addressing #191. :sunglasses:

PR #242, opened 2015-07-28
Add install.ps1 for easy installation on Windows

aptxkid:
This file will be uploaded to Box, so that people can easily install ClusterRunner on Windows without even cloning the repo.


boxcla:

josephharrington:
Should this be in a windows subfolder?

PR #243, opened 2015-07-28
Adopt the same installation method of Chocolaty

aptxkid:
Now the only thing needed to get ClusterRunner on Windows is by running a command in CMD.
No cloning the repo required. Just like the way it is on Mac and Linux.


boxcla:

josephharrington:
:+1:

PR #246, opened 2015-08-17
Canceling a build should resolve slave statuses

drobertduke:
Currently if you cancel a build which has slaves allocated,
the build will be terminated but the slaves will remain allocated.
This deallocates all slaves attached to the build.

We also add locking around build preparation and canceling to
avoid race conditions.

Added a cancel subcommand.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -46,6 +46,7 @@ class Build(object):
         self._project_type = None
         self._build_completion_lock = Lock()  # protects against more than one thread detecting the build's finish
         self._subjob_assignment_lock = Lock()  # prevents subjobs from being skipped
+        self._subjob_queuing_lock = Lock()
         self._slaves_allocated = []
         self._num_executors_allocated = 0
         self._num_executors_in_use = 0
@@ -119,9 +120,13 @@ class Build(object):
         self._unstarted_subjobs = Queue(maxsize=len(subjobs))
         self._finished_subjobs = Queue(maxsize=len(subjobs))
 
-        for subjob in subjobs:
-            self._all_subjobs_by_id[subjob.subjob_id()] = subjob
-            self._unstarted_subjobs.put(subjob)
+        with self._subjob_queuing_lock:
+            if self._status() in [BuildStatus.FINISHED, BuildStatus.ERROR, BuildStatus.CANCELED]:
+                self._logger.warn('Could not complete build preparation, build status is {}', self._status())
+                return
+            for subjob in subjobs:
+                self._all_subjobs_by_id[subjob.subjob_id()] = subjob
+                self._unstarted_subjobs.put(subjob)
 
         self._max_executors = job_config.max_executors
         self._max_executors_per_slave = job_config.max_executors_per_slave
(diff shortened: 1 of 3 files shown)


boxcla:

drobertduke:
There are no tests yet, looking for feedback first.

I'm not excited about the complexity I added around locking in prepare() and cancel().  Any suggestions?  Ultimately I'd like to offload this kind of logic to a multi-threaded state machine.

I added a 'cancel' subcommand.  I thought of extending the 'build' subcommand with a cancel option, but currently we are using 'build' like a verb in the CLI but as a noun in the rest of the app.

app/master/build.py
```diff
@@ -119,9 +120,13 @@ def prepare(self, subjobs, job_config):
...
-        for subjob in subjobs:
-            self._all_subjobs_by_id[subjob.subjob_id()] = subjob
-            self._unstarted_subjobs.put(subjob)
+        with self._subjob_queuing_lock:
+            if self._status() in [BuildStatus.FINISHED, BuildStatus.ERROR, BuildStatus.CANCELED]:
+                self._logger.warn('Could not complete build preparation, build status is {}', self._status())
+                return
+            for subjob in subjobs:
+                self._all_subjobs_by_id[subjob.subjob_id()] = subjob
+                self._unstarted_subjobs.put(subjob)
```
josephharrington: Did you run into a situation where you needed this lock? I was thinking that the line `if not self._preparation_coin.spend():` protected against this being executed more than once (implying that this couldn't be executed by multiple threads simultaneously) but I may not be remembering correctly.

josephharrington: Oops, nevermind -- I see you use it below also.

josephharrington:
Want to split the cancel subcommand stuff into a separate commit? I think that is independent from the fixing the issues with cancel and probably easier to test / work out bugs for / etc.

app/master/build.py
```diff
@@ -327,13 +332,17 @@ def cancel(self):
...
+            while self._unstarted_subjobs is not None and not self._unstarted_subjobs.empty():
+                try:
+                    # A subjob may be asynchronously pulled from this queue, so we need to avoid blocking when empty.
+                    self._unstarted_subjobs.get(block=False)
+                except Empty:
+                    break
+
+        # Tell each allocated slave to teardown
+        for slave in self._slaves_allocated:
+            self.execute_next_subjob_or_teardown_slave(slave)
```
tjlee0909: Can't you just call self._free_slave_executor(slave) directly?

tjlee0909:
Just to be explicit, the race condition you are preventing is the one when the build preparation has progressed to the point where it is adding subjobs to the unstarted subjobs, and at the same time there is a build cancellation request?

app/master/build.py
josephharrington: I'm not sure it will be as easy as calling `execute_next_subjob_or_teardown_slave` for each slave. It's kind of misleading -- a better name for that method would probably be `execute_next_subjob_or_free_executor`.

When cancel is called, there may currently be subjobs executing on the slave (in our case, up to 10). I think in that case, calling `execute_next_subjob_or_teardown_slave` for each slave will just make the master think there are now 9 busy executors on each slave. In fact if all 10 of those subjobs then report results back, the last one will probably cause a request exception since we'll try to free 1 executor when the master thinks all 10 were already freed.

josephharrington:
Just FYI, I have a refactor of the Build class in progress that might also address this issue at the same time.

PR #248, opened 2015-08-19
Generate session id in cluster master, and return in every response.

tjlee0909:
The header key is: 'Session-Id'.

This key is not being used anywhere yet. This change is to just generate the session id, and return it with every response.

### DIFF
--- a/app/web_framework/cluster_master_application.py
+++ b/app/web_framework/cluster_master_application.py
@@ -2,6 +2,7 @@ import http.client
 import os
 import tornado.web
 import urllib.parse
+import uuid
 
 from app.util import analytics
 from app.util.conf.configuration import Configuration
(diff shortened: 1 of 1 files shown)


boxcla:

app/web_framework/cluster_master_application.py
```diff
@@ -248,14 +260,16 @@ class _BuildResultHandler(ClusterBaseHandler, tornado.web.StaticFileHandler):
 
     From the Tornado docs: "for heavy traffic it will be more efficient to use a dedicated static file server".
     """
-    def initialize(self, route_node=None, cluster_master=None):
+    def initialize(self, route_node=None, cluster_master=None, session_id=None):
```
tjlee0909: This API endpoint failed in the end-to-end tests before I added the session_id parameter. It looks like the whole reason why this customization exists is to set the 'path' parameter to None in the super method.

tjlee0909:
Actually, scracth this, I realize that this makes it difficult for the validation of the session token (coming up in a later PR). Going to redo this.

Issue #245, opened 2015-08-12
Implement a session token for ClusterRunner


tjlee0909:
ClusterRunner currently retains no state in between service restarts. All first builds after a restart starts with build id 1.

This becomes problematic when integrating ClusterRunner with a different service that persists state forever. We need the concept of a session token, that is unique per service restart, that would be used to authenticate requests and make sure that the requests are for the correct builds.

josephharrington:
+1
It would be nice if specifying the session token in API requests was optional. If I want to pull up the api response for a specific build, I shouldn't *need* to go figure out the current session id. 

I was also thinking that http if-match and etag headers might be appropriate for this.

tjlee0909:
Agreed about making it optional, meaning that if someone doesn't specify the session ID, it means they don't care about whether or not they might be looking at the wrong session.

I'm wondering how we will return the session ID. Should it be returned with the build id in the POST build requests?

I'm thinking that all GETS that do provide the session ID, and the session ID is not the same, then we will return a 40X code.

I'm not quite sure what we want to do for POSTS regarding session ID's.

I don't know what an etag header is, I'll look it up.

josephharrington:
The downside of a header is that it wouldn't be completely contained within the url. That might be a dealbreaker.

PR #249, opened 2015-08-19
Generate session id in cluster master, and return in every response.

tjlee0909:
Additionally, if the session id is specified in the header then enforce that it matches the current session id. If not, return a 404.

### DIFF
--- a/app/util/session_id.py
+++ b/app/util/session_id.py
@@ -1,0 +1,16 @@
+import uuid
+
+
+class SessionId(object):
+    SESSION_HEADER_KEY = 'Session-Id'
+    _session_id = None
+
+    @classmethod
+    def get(cls):
+        """
+        :return: the unique, generated session id string.
+        :rtype: str
+        """
+        if cls._session_id is None:
+            cls._session_id = str(uuid.uuid4())
+        return cls._session_id
--- a/app/web_framework/cluster_master_application.py
+++ b/app/web_framework/cluster_master_application.py
@@ -4,9 +4,11 @@ import tornado.web
 import urllib.parse
 
 from app.util import analytics
+from app.util import log
 from app.util.conf.configuration import Configuration
 from app.util.decorators import authenticated
-from app.util.exceptions import ItemNotFoundError
+from app.util.exceptions import ItemNotFoundError, PreconditionFailedError
+from app.util.session_id import SessionId
 from app.util.url_builder import UrlBuilder
 from app.web_framework.cluster_application import ClusterApplication
 from app.web_framework.cluster_base_handler import ClusterBaseAPIHandler, ClusterBaseHandler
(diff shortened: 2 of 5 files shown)


boxcla:

app/util/session_id.py
```diff
@@ -0,0 +1,16 @@
+import uuid
+
+
+class SessionId(object):
+    SESSION_HEADER_KEY = 'Session-Id'
```
josephharrington: Just curious -- did you make up this header name or did you find other similar uses of it? All I found was http://www.w3.org/TR/WD-session-id-960221.html

tjlee0909:
Implements https://github.com/box/ClusterRunner/issues/245

app/util/session_id.py
tjlee0909: I saw that the convention for header keys was 'First-Letter-Uppercase-With-Dashes', and just went with that.

app/web_framework/cluster_master_application.py
```diff
@@ -64,9 +66,31 @@ def initialize(self, route_node=None, cluster_master=None):
...
+        no longer exists.
+        """
+        session_id = self.request.headers.get(SessionId.SESSION_HEADER_KEY)
+
+        if session_id is not None and session_id != SessionId.get():
+            raise ItemNotFoundError('Specified session id: {} has expired and is invalid.'.format(session_id))
+
+        super().prepare()
+
+    def write(self, response):
```
josephharrington: There's a `set_default_headers()` method you could do this in instead -- it probably doesn't matter too much for this specific case but the reason I mention it is because I think we're already setting the CORS header via that method in the base class.

app/web_framework/cluster_master_application.py
```diff
@@ -64,9 +66,31 @@ def initialize(self, route_node=None, cluster_master=None):
...
+    def prepare(self):
+        """
+        If the request has specified the session id, which is optional, and the session id does not match
+        the current instance's session id, then the client is asking for a resource that has expired and
+        no longer exists.
+        """
+        session_id = self.request.headers.get(SessionId.SESSION_HEADER_KEY)
+
+        if session_id is not None and session_id != SessionId.get():
+            raise ItemNotFoundError('Specified session id: {} has expired and is invalid.'.format(session_id))
```
josephharrington: Should we make this a 412 (Precondition Failed) instead of a 404?

tjlee0909: Yeah, we could go with that.

app/util/session_id.py
josephharrington: Ah, okay. That spec above is just a draft so I'm not too concerned about overlap.

app/web_framework/cluster_master_application.py
tjlee0909: That would also set the Session-Id on the slave services, and because the slave service is on another process, the Session-Id returned there would be different from the Session-Id generated on the master service.

josephharrington:
:+1: A few minor comments but looks good to me.

app/web_framework/cluster_master_application.py
tjlee0909: Unless you meant adding it to set_default_headers() in this class, not the ClusterBaseAPIHandler()

tjlee0909: Ah, it's a built-in tornado method. Sure, will add it there.

josephharrington: Yeah I meant still in the subclass, just doing it in `set_default_headers()` instead of `prepare()`.

app/web_framework/cluster_master_application.py
```diff
@@ -64,9 +66,29 @@ def initialize(self, route_node=None, cluster_master=None):
...
+        if session_id is not None and session_id != SessionId.get():
+            raise PreconditionFailedError('Specified session id: {} has expired and is invalid.'.format(session_id))
+
+        super().prepare()
+
+    def set_default_headers(self):
+        """
+        Inject the session id in the header.
+        """
+        self.set_header(SessionId.SESSION_HEADER_KEY, SessionId.get())
```
josephharrington: You might need to call `super().set_default_headers()`.

tjlee0909: Will do.

josephharrington:
:+1: :+1:  :princess:

PR #250, opened 2015-08-20
Replace directory path in atoms with the environment variable

tjlee0909:
This way, when viewing the atom commands in the API (among other places), you will see $PROJECT_DIR/test/path.php instead of /tmp/clusterrunner/blahblahgarbage/test/path.php.

### DIFF
--- a/app/master/subjob.py
+++ b/app/master/subjob.py
@@ -129,12 +129,8 @@ class Subjob(object):
             timings_file_path = os.path.join(artifact_dir, BuildArtifact.TIMING_FILE)
             if os.path.exists(timings_file_path):
                 with open(timings_file_path, 'r') as f:
-                    # Strip out the project directory from atom timing data in order to have all
-                    # atom timing data be relative and project directory agnostic (the project
-                    # directory will be a generated unique path for every build).
-                    atom_key = atom.command_string.replace(self.project_type.project_directory, '')
                     atom.actual_time = float(f.readline())
-                    timings[atom_key] = atom.actual_time
+                    timings[atom.command_string] = atom.actual_time
             else:
                 self._logger.warning('No timing data for subjob {} atom {}.',
                                      self._subjob_id, atom_id)
--- a/test/unit/master/test_atomizer.py
+++ b/test/unit/master/test_atomizer.py
@@ -1,37 +1,37 @@
-from unittest.mock import MagicMock
+from unittest.mock import Mock
 from app.master.atomizer import Atomizer, AtomizerError
 
-from app.project_type.project_type import ProjectType
 from app.util.process_utils import get_environment_variable_setter_command
 from test.framework.base_unit_test_case import BaseUnitTestCase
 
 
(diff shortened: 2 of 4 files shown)


boxcla:

app/master/subjob.py
```diff
@@ -129,12 +129,8 @@ def read_timings(self):
             timings_file_path = os.path.join(artifact_dir, BuildArtifact.TIMING_FILE)
             if os.path.exists(timings_file_path):
                 with open(timings_file_path, 'r') as f:
-                    # Strip out the project directory from atom timing data in order to have all
-                    # atom timing data be relative and project directory agnostic (the project
-                    # directory will be a generated unique path for every build).
-                    atom_key = atom.command_string.replace(self.project_type.project_directory, '')
                     atom.actual_time = float(f.readline())
-                    timings[atom_key] = atom.actual_time
+                    timings[atom.command_string] = atom.actual_time
```
josephharrington: I think this will invalidate our current timing data on some jobs but that's probably not a big deal unless we restart during peak load. :grimacing:

test/unit/master/test_atomizer.py
```diff
@@ -1,37 +1,37 @@
...
-_FAKE_ATOMIZER_COMMAND_OUTPUT = 'test_a.py\ntest_b.py\ntest_c.py\n'
+_FAKE_ATOMIZER_COMMAND_OUTPUT = '/tmp/test/directory/test_a.py\n/tmp/test/directory/test_b.py\n/tmp/test/directory/test_c.py\n'
 _SUCCESSFUL_EXIT_CODE = 0
 _FAILING_EXIT_CODE = 1
 
 
 class TestAtomizer(BaseUnitTestCase):
     def test_atomizer_returns_expected_atom_list(self):
-        mock_project = MagicMock(spec_set=ProjectType)
+        mock_project = Mock()
```
josephharrington: I think you could keep the spec on `ProjectType` by doing `Mock(spec=ProjectType)`. Using spec instead of spec_set relaxes the restriction on setting attributes but still verifies that any mocked methods must be valid ProjectType methods.

josephharrington:
A small comment on the tests, but looks good to me. This is great! :sunglasses:

app/master/subjob.py
tjlee0909: Yeah, I think I'll restart ClusterRunner sometime later tonight and run a few jobs through there before leaving the maintenance window.

test/unit/master/test_atomizer.py
tjlee0909: I removed the spec_set because when I do, I can't set the project_directory, as I do in line 18.

I get this error:

AttributeError: Mock object has no attribute 'project_directory'

josephharrington: Right, I was saying to use `spec` instead of `spec_set`.

tjlee0909: Ah, I didn't read.

PR #251, opened 2015-08-20
Update timing data keys, not overwrite the file entirely.

tjlee0909:
This will be necessary when we implement overridden atoms from build
requests. In such a build request, we will not have ran all the atoms,
and if we were to keep the existing behavior, we would wipe out all of
the atom timing data except for the select few that were manually
specified in the build request.

This method also leaves the door open in the future if we want to always
update the timing data, regardless of overall build success/failure, but
only for successful atoms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
 # each line defined in env will create a new parallel test job on Travis
 env:
   - TARGETS="lint test-unit-with-coverage"
+  - TARGETS="test-integration-with-coverage"
   - TARGETS="test-functional test-unit-via-clusterrunner"
 
 install:
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,11 @@ test-unit:
 
 test-unit-with-coverage:
 	$(call print_msg, Running unit tests with coverage... )
-	nosetests -v --with-xcoverage --cover-package=app test/unit test/integration
+	nosetests -v --with-xcoverage --cover-package=app test/unit
+
+test-integration-with-coverage:
+	$(call print_msg, Running unit tests with coverage... )
+	nosetests -v --with-xcoverage --cover-package=app test/integration
 
 test-unit-via-clusterrunner:
 	$(call print_msg, Running unit tests via ClusterRunner... )
--- a/app/master/build_artifact.py
+++ b/app/master/build_artifact.py
@@ -47,9 +47,10 @@ class BuildArtifact(object):
             self._logger.debug('Created new timing file in {}', timing_file_path)
             return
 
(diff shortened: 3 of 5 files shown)


boxcla:

app/master/build_artifact.py
```diff
@@ -104,12 +105,26 @@ def generate_failures_file(self):
...
+        does not replace the entire timing data file, but rather only replaces the timing data for
+        individual atom keys.
+        :param timing_file_path: str
+        :param new_timing_data: dict[str, float]
+        """
+        with open(timing_file_path) as timing_file:
+            timing_data = json.load(timing_file)
+
+        timing_data.update(new_timing_data)
+        app.util.fs.write_file(json.dumps(timing_data), timing_file_path)
```
josephharrington: Maybe call `self._write_timing_data_to_file(timing_file_path, timing_data)` here instead?

app/master/build_artifact.py
```diff
@@ -47,9 +47,10 @@ def write_timing_data(self, timing_file_path, timing_data):
             self._logger.debug('Created new timing file in {}', timing_file_path)
```
tjlee0909: I realized we can combine the above if block with the one below.

so

if not os.path.is.file(...) or len(self.get_failed_commands()) == 0:
...

Makefile
```diff
@@ -25,7 +25,11 @@ test-unit:
 
 test-unit-with-coverage:
 	$(call print_msg, Running unit tests with coverage... )
-	nosetests -v --with-xcoverage --cover-package=app test/unit test/integration
+	nosetests -v --with-xcoverage --cover-package=app test/unit
+
+test-integration-with-coverage:
+	$(call print_msg, Running unit tests with coverage... )
+	nosetests -v --with-xcoverage --cover-package=app test/integration
```
tjlee0909: I had to break these out because travis was spuriously failing my integration tests consistently. I suspect is has something to do with leftover global state.

.travis.yml
```diff
@@ -6,6 +6,7 @@ python:
 # each line defined in env will create a new parallel test job on Travis
 env:
   - TARGETS="lint test-unit-with-coverage"
+  - TARGETS="lint test-integration-with-coverage"
```
josephharrington: You could remove lint from this one so it just runs once.

Makefile
josephharrington: Sounds good. I thought this would mess with coveralls but it doesn't look like coverage went down at all so it must be fine.

josephharrington:
:+1: with a tiny comment about the .travis.yml change.

PR #252, opened 2015-08-21
Support atom overrides in builds/ POST endpoint.

tjlee0909:
If the atom_overrides build parameter is specified in the API request, the 'atomizer' section of the clusterrunner.yaml config gets bypassed.

Having this parameter be specified allows us to re-run just failed atoms, for example.

### DIFF
--- a/app/master/build_request_handler.py
+++ b/app/master/build_request_handler.py
@@ -3,6 +3,7 @@ import os
 from queue import Queue
 from threading import Lock
 
+from app.master.atom import Atom
 from app.master.atom_grouper import AtomGrouper
 from app.master.build_request import BuildRequest
 from app.master.subjob import Subjob
@@ -148,7 +149,13 @@ class BuildRequestHandler(object):
         :type project_type: project_type.project_type.ProjectType
         :rtype: list[Subjob]
         """
-        atoms_list = job_config.atomizer.atomize_in_project(project_type)
+        # Users can override the list of atoms to be run in this build. If the atoms_override
+        # was specified, we can skip the atomization step and use those overridden atoms instead.
+        if project_type.atoms_override is not None:
+            atoms_string_list = project_type.atoms_override
+            atoms_list = [Atom(atom_string_value) for atom_string_value in atoms_string_list]
+        else:
+            atoms_list = job_config.atomizer.atomize_in_project(project_type)
 
         # Group the atoms together using some grouping strategy
         timing_file_path = project_type.timing_file_path(job_config.name)
--- a/test/unit/master/test_build_request_handler.py
+++ b/test/unit/master/test_build_request_handler.py
@@ -1,0 +1,41 @@
+from genty import genty, genty_dataset
+from unittest.mock import Mock
+
+from app.project_type.project_type import ProjectType
+from app.master.atom import Atom
+from app.master.build_request_handler import BuildRequestHandler
+from app.master.atomizer import Atomizer
+from app.master.job_config import JobConfig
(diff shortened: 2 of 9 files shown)


boxcla:

app/master/build_request_handler.py
```diff
@@ -148,7 +149,13 @@ def _compute_subjobs_for_build(self, build_id, job_config, project_type):
         :type project_type: project_type.project_type.ProjectType
         :rtype: list[Subjob]
         """
-        atoms_list = job_config.atomizer.atomize_in_project(project_type)
+        # Users can override the list of atoms to be run in this build. If the atoms_override
+        # was specified, we can skip the atomization step and use those overridden atoms instead.
+        if project_type.atoms_override is not None:
+            atoms_string_list = project_type.atoms_override
+            atoms_list = [Atom(atom_string_value) for atom_string_value in atoms_string_list]
```
josephharrington: Do you think we should do any prevalidation that the overridden atoms are valid or should we just let the build go ahead and fail slave-side? I suppose it would be the same effect as a misconfigured job config in that case (e.g., specifying different atomizer variable in the commands section and the atomizer section).

test/unit/master/test_build_request_handler.py
```diff
@@ -0,0 +1,40 @@
...
+from test.framework.base_unit_test_case import BaseUnitTestCase
+
+
+@genty
+class TestBuildRequestHandler(BaseUnitTestCase):
+    @genty_dataset(
+        atoms_override_specified=(['override1', 'override2'], None, False),
+        atoms_override_not_specified=(None, [Atom('atom_value_1'), Atom('atom_value_2')], True),
+    )
+    def test_compute_subjobs_for_build(self, atoms_override, atomizer_output, atomizer_called):
```
josephharrington: Want to add a behavior for the test name? Maybe something like `test_compute_subjobs_for_build_only_calls_atomizer_if_no_overrides`? I always feel like it helps test readability when you know from the name what the test is trying to verify.

josephharrington:
:+1:

test/unit/master/test_build_request_handler.py
tjlee0909: This test uses a genty dataset that tests both the override-specified case, and the override not-specified case.

app/master/build_request_handler.py
tjlee0909: It's funny you should ask.

For the feature that I want to build, I'd like it if we:
- Ran the atomization first.
- Checked to see if the atoms_override list is a proper subset of the original atomization.
- Only run the atoms_override that is a subset of both the atoms_override and original atomization.

However, I feel that this behavior is unexpected given the API, and that I do believe the existing functionality (slave fails if the atom is invalid, and essentially treat it as we would a bad atomization command section).

test/unit/master/test_build_request_handler.py
josephharrington: Right, if you wanted to be totally specific you could use "...if_and_only_if...". My main point was that without a behavior in the test name it feels like it's saying we're testing literally everything the `compute_subjobs_for_build` method does.

tjlee0909: Ah, right. Yeah, I'll change that.

app/master/build_request_handler.py
josephharrington: Ah, hmm -- yeah it might be cool if we did not require atoms to be output by the atomizer -- more flexibility, etc. I support leaving it how you currently have it and improving it later if it causes terribly confusing errors or behaviors.

PR #253, opened 2015-08-25
Server-side changes to allow cors.

tjlee0909:
Currently, client-side pre-flight requests sends OPTION type requests
that get rejected by the server.

http://stackoverflow.com/questions/1099787/jquery-ajax-post-sending-options-as-request-method-in-firefox


boxcla:

josephharrington:
:+1:

PR #256, opened 2015-08-26
Change 'failed_atoms' endpoint API representation for a build

tjlee0909:
to return an API representation of the whole atom.

### DIFF
--- a/app/master/atom.py
+++ b/app/master/atom.py
@@ -15,6 +15,7 @@ class Atom(object):
             actual_time=None,
             exit_code=None,
             state=None,
+            id=None
     ):
         """
         :type command_string: str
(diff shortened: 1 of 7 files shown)


boxcla:

tjlee0909:
Testing to come.

app/master/atom.py
```diff
@@ -28,3 +28,12 @@ def __init__(
...
         self.state = state
+
+    def api_representation(self):
+        return {
+            'command_string': self.command_string,
+            'expected_time': self.expected_time,
+            'actual_time': self.actual_time,
+            'exit_code': self.exit_code,
+            'state': self.state,
+        }
```
josephharrington: Nice. In subjob.py there is a `get_atoms()` method that we could probably get rid of in favor of this now.

app/master/build.py
```diff
@@ -445,18 +450,25 @@ def _status(self):
...
-        :rtype: list [str] | None
+        The atoms that failed. Returns None if the build hasn't completed yet. Returns empty set if
+        build has completed and no atoms have failed.
+        :rtype: list[Atom] | None
         """
-        if self._is_canceled:
-            return []
+        if self._failed_atoms is None and self.is_finished:
+            # Why was this conditional here in the first place? Don't we only care about whether or not
+            # the build is finished? I think we can remove this.
```
josephharrington: Are you wondering about the `_is_canceled` check? It would be awesome to remove the check but obvs we should do a manual test first. I wonder if build_artifact was raising an exception since a canceled build doesn't have an artifact dir?

app/master/build.py
```diff
@@ -445,18 +450,25 @@ def _status(self):
         else:
             return BuildStatus.BUILDING
 
-    def _failed_atoms(self):
+    def failed_atoms(self):
```
josephharrington: Can this stay private? Doesn't look like it is called anywhere else...

PR #258, opened 2015-08-31
Build refactor - Create BuildScheduler class

josephharrington:
This is one of a few changes I have queued up to refactor our Build
class to not be terrible.

- Break out anything in the `Build` class related to slave interaction
  into a new `BuildScheduler` class.

  A `BuildScheduler` has a one-to-one relationship with a `Build`.
  Every build gets a scheduler. The build does not know anything about
  the scheduler, and the scheduler only interacts with the build via
  its public interface.

- Create a `BuildSchedulerPool` class to help with managing
  `BuildScheduler` instances. This prevents `ClusterMaster` from
  needing to keep yet another id-to-instance dict.

  The `BuildSchedulerPool` will also eventually take care of cleaning
  up old scheduler instances.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -7,8 +7,7 @@ import time
 import uuid
 
 from app.master.build_artifact import BuildArtifact
-from app.master.slave import SlaveMarkedForShutdownError
-from app.util import analytics, util
+from app.util import util
 from app.util.conf.configuration import Configuration
 from app.util.counter import Counter
 from app.util.exceptions import ItemNotFoundError
@@ -40,21 +39,15 @@ class Build(object):
 
         self._error_message = None
         self.is_prepared = False
+        self._setup_is_started = False
         self._preparation_coin = SingleUseCoin()  # protects against separate threads calling prepare() more than once
         self._is_canceled = False
 
         self._project_type = None
         self._build_completion_lock = Lock()  # protects against more than one thread detecting the build's finish
-        self._subjob_assignment_lock = Lock()  # prevents subjobs from being skipped
-        self._slaves_allocated = []
-        self._num_executors_allocated = 0
-        self._num_executors_in_use = 0
-
-        self._max_executors = float('inf')
-        self._max_executors_per_slave = float('inf')
 
         self._all_subjobs_by_id = {}
-        self._unstarted_subjobs = None
+        self._unstarted_subjobs = None  # WIP: Move subjob queues to BuildScheduler class.
         self._finished_subjobs = None
         self._failed_atoms = None
         self._postbuild_tasks_are_finished = False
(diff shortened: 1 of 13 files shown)


boxcla:

app/master/build.py
```diff
@@ -40,21 +39,15 @@ def __init__(self, build_request):
 
         self._error_message = None
         self.is_prepared = False
+        self._setup_is_started = False
```
josephharrington: This is a temporary flag -- I'll get rid of it in another change in favor of using the state machine.

app/master/build.py
```diff
@@ -307,6 +220,10 @@ def _mark_subjob_complete(self, subjob_id):
             self._logger.info("All results received for build {}!", self._build_id)
             SafeThread(target=self._perform_async_postbuild_tasks, name='PostBuild{}'.format(self._build_id)).start()
 
+    def mark_started(self):
+        self._setup_is_started = True
+        self._record_state_timestamp(BuildStatus.BUILDING)
```
josephharrington: Just for reference, the `_record_state_timestamp` correspond pretty closely to where the state machine will be making transitions.

app/master/build_scheduler.py
```diff
@@ -0,0 +1,122 @@
...
+    This class handles the logic of taking a Build and distributing its subjobs to
+    slaves. There is a one-to-one relationship between instances of this class and
+    instances of Build.
+
+    The data flow between this class and Build is unidirectional; the build instance
+    itself shouldn't know anything about its scheduler or slaves or where subjobs
+    are executing. All that goes here.
+
+    Ususally this class is instantiated and managed by a BuildSchedulerPool.
+    """
```
josephharrington: Almost everything in this class is pulled unchanged from the original methods in `Build`.

app/master/build_scheduler.py
```diff
@@ -0,0 +1,122 @@
...
+        """ :rtype: int """
+        return self._build.build_id()
+
+    def needs_more_slaves(self):
+        """
+        Determine whether or not this build should have more slaves allocated to it.
+
+        :rtype: bool
+        """
+        return self._num_executors_allocated < self._max_executors and not self._build._unstarted_subjobs.empty()
```
josephharrington: The `self._build._unstarted_subjobs` reference is a temporary violation of the build's public interface. I plan to fix this in a future change, probably by moving the subjob queues into this class.

Some of the state computations in the `Build` class still depend on these queues so it's not as easy to remove them into this class before the state machine refactor.

app/master/slave.py
```diff
@@ -68,7 +70,7 @@ def setup(self, build):
         setup_url = self._slave_api.url('build', build.build_id(), 'setup')
         post_data = {
             'project_type_params': slave_project_type_params,
-            'build_executor_start_index': build.num_executors_allocated,
+            'build_executor_start_index': executor_start_index,
```
drobertduke: This is good

drobertduke:
It's happening! :+1: :pizza:

PR #259, opened 2015-09-15
Build refactor - Create SubjobCalculator class

josephharrington:
- Move prepare logic from BuildRequestHandler into Build.
- Move subjob computing logic from BuildRequestHandler into new class.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -7,6 +7,8 @@ import time
 import uuid
 
 from app.master.build_artifact import BuildArtifact
+from app.master.build_request import BuildRequest
+from app.project_type.project_type import ProjectType
 from app.util import util
 from app.util.conf.configuration import Configuration
 from app.util.counter import Counter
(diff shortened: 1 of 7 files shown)


boxcla:

app/master/build_request_handler.py
```diff
@@ -107,110 +100,13 @@ def _prepare_build_async(self, build, project_lock):
...
-
-        self._logger.info('Successfully fetched project for build {}.', build_id)
-        job_config = project_type.job_config()
-
-        if job_config is None:
-            build.mark_failed('Build failed while trying to parse clusterrunner.yaml.')
-            return
-
-        subjobs = self._compute_subjobs_for_build(build_id, job_config, project_type)
-        build.prepare(subjobs, job_config)
```
josephharrington: This logic all moved essentially unchanged into build.py.

app/master/build_request_handler.py
```diff
@@ -107,110 +100,13 @@ def _prepare_build_async(self, build, project_lock):
...
-                    atom_time_map = json.load(json_file)
-                except ValueError:
-                    self._logger.warning('Failed to load timing data from file that exists {}', timing_file_path)
-
-        if atom_time_map is not None and len(atom_time_map) > 0:
-            atom_grouper = TimeBasedAtomGrouper(atoms, max_executors, atom_time_map, project_directory)
-        else:
-            atom_grouper = AtomGrouper(atoms, max_executors)
-
-        return atom_grouper.groupings()
```
josephharrington: This logic all moved essentially unchanged into subjob_calculator.py.

app/master/build.py
```diff
@@ -103,18 +105,30 @@ def generate_project_type(self):
...
+
+        if not isinstance(self.project_type, ProjectType):
+            raise RuntimeError('Build {} has no project set.'.format(self._build_id))
 
         if not self._preparation_coin.spend():
             raise RuntimeError('prepare() was called more than once on build {}.'.format(self._build_id))
 
+        self._logger.info('Fetching project for build {}.', self._build_id)
+        self.project_type.fetch_project()
+        self._logger.info('Successfully fetched project for build {}.', self._build_id)
```
josephharrington: The main motivation for this change is when adding the state machine to include a "preparing" state. To enable that, the `self.project_type.fetch_project()` call has to be done within the Build class.

Also, it makes more sense to me to include all of the logic for build preparation in one place instead of doing half in BuildRequestHandler and half in Build.

tjlee0909: Yes! This came up before, and it bothered me that "prepare()" method didn't do the most important thing in build preparation: fetching the ref.

app/master/build_request_handler.py
```diff
@@ -107,110 +100,13 @@ def _prepare_build_async(self, build, project_lock):
             analytics.record_event(analytics.BUILD_PREPARE_START, build_id=build.build_id(),
                                    log_msg='Build preparation loop is handling request for build {build_id}.')
             try:
-                self._prepare_build(build)
+                build.prepare(self._subjob_calculator)
```
tjlee0909: I recognize the benefit (especially in testing) of injecting dependencies like this, instead of instantiating them directly inside of the method itself.

I'm not asking you to change this or anything, but should we make this pattern the norm?

tjlee0909:
:+1:

app/master/build_request_handler.py
josephharrington: Ah, yeah good question -- I considered that briefly (just instantiating this inside the Build object), but kind of waffled on it.

It's generally easier to test with injected dependencies, but with Python specifically it's not too hard to mock out even inline instantiations, so it's probably not a good enough reason for us to go adding a bunch of extra parameters to our constructors and methods. (I could see a stronger argument for dependency injection to support us creating a user plugin architecture, but that's not really on the roadmap.)

Mostly I did this just because the subjob calculator really only had to be instantiated once. Actually now the more I talk about it, the more I lean towards just hiding it from the build request handler and just instantiating it for each build. You're right that it is a bit of a break from our existing style. I might change this after all. :sunglasses:

PR #260, opened 2015-09-21
added subjob id to atom

ethomas2:
added subjob id field to atom's api representation. This will make the is-it-your-fault-service cleaner

### DIFF
--- a/app/master/atom.py
+++ b/app/master/atom.py
@@ -15,7 +15,8 @@ class Atom(object):
             actual_time=None,
             exit_code=None,
             state=None,
-            atom_id=None
+            atom_id=None,
+            subjob_id=None
     ):
         """
         :type command_string: str
@@ -24,12 +25,14 @@ class Atom(object):
         :type exit_code: int | None
         :type state: `:class:AtomState` | None
         :type atom_id: int | None
+        :type subjob_id: int | None
         """
         self.command_string = command_string
         self.expected_time = expected_time
         self.actual_time = actual_time
         self.exit_code = exit_code
         self.state = state
+        self.subjob_id = subjob_id
         self.id = atom_id
 
     def api_representation(self):
@@ -40,4 +43,5 @@ class Atom(object):
             'exit_code': self.exit_code,
             'state': self.state,
             'id': self.id,
+            'subjob_id': self.subjob_id,
         }
--- a/app/master/subjob.py
+++ b/app/master/subjob.py
@@ -27,9 +27,21 @@ class Subjob(object):
         self.project_type = project_type
         self.job_config = job_config
         self._atoms = atoms
+        self._set_atoms_subjob_id(atoms, subjob_id)
         self._set_atom_state(AtomState.NOT_STARTED)
         self.timings = {}  # a dict, atom_ids are the keys and seconds are the values
         self.slave = None  # The slave that had been assigned this subjob. Is None if not started.
+
+    def _set_atoms_subjob_id(self, atoms, subjob_id):
+        """
+        Set the subjob_id on each atom
+        :param atoms: an array of atoms to set the subjob_id on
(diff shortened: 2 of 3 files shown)


boxcla:
Hi @ethomas2, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

josephharrington:
Want to add a unit test for this change?

app/master/subjob.py
```diff
@@ -27,10 +27,15 @@ def __init__(self, build_id, subjob_id, project_type, job_config, atoms):
         self.project_type = project_type
         self.job_config = job_config
         self._atoms = atoms
+        self._set_atoms_subjob_id()
```
tjlee0909: My only concern is that if someone accidentally changes the order in the constructor of the self._atoms = atoms and the self._set_atoms_subjob_id(), then the functionality will fail.

If you just write a unit test to make sure the atoms get the subjob id's set, then I think you're all good.

app/master/atom.py
```diff
@@ -30,6 +31,7 @@ def __init__(
         self.actual_time = actual_time
         self.exit_code = exit_code
         self.state = state
+        self.subjob_id = subjob_id
```
aptxkid: Update the docstring for subjob_id

app/master/subjob.py
```diff
@@ -27,10 +27,15 @@ def __init__(self, build_id, subjob_id, project_type, job_config, atoms):
         self.project_type = project_type
         self.job_config = job_config
         self._atoms = atoms
+        self._set_atoms_subjob_id(atoms, subjob_id)
         self._set_atom_state(AtomState.NOT_STARTED)
         self.timings = {}  # a dict, atom_ids are the keys and seconds are the values
         self.slave = None  # The slave that had been assigned this subjob. Is None if not started.
 
+    def _set_atoms_subjob_id(self, atoms, subjob_id):
```
tjlee0909: Could you add the docstrings for atoms and subjob_id?

app/master/atom.py
```diff
@@ -24,12 +25,14 @@ def __init__(
         :type exit_code: int | None
         :type state: `:class:AtomState` | None
         :type atom_id: int | None
+        :type subjob_id: int
```
tjlee0909: Technically, this hsould be "int | None"

tjlee0909:
Posted two tiny comments about pydoc, but once that's fixed :+1:

test/unit/master/test_subjob.py
```diff
@@ -36,6 +36,12 @@ def setUp(self):
             ],
         )
 
+    def test_set_atoms_subjob_id(self):
+        atoms = [Mock(), Mock()]
+        self._subjob._set_atoms_subjob_id(atoms, 4)
+        for atom in atoms:
+            self.assertEqual(atom.subjob_id, 4)
+
```
josephharrington: This test is okay but I think we can do better. I think generally we should try to test through public interfaces as much as possible.

It would be better to do a test that passes the list of atoms to the Subjob constructor and then checks that the subjob_id on each atom was set.

josephharrington:
@ethomas2 Sorry for my late comment -- would you want to revise the test whenever you get a chance? :sunglasses:

test/unit/master/test_subjob.py
josephharrington: Also, it would be great if we keep with the convention of naming tests like "test_X_does_Y_when_Z" or something similar -- basically including a description of the tested behavior in the test name.

I prefer to name tests that way because it helps us think about tests as a "spec" for the application. If we just name our tests "test_<method_name>" then it is not clear what exactly we're expecting the test to do, and it is tempting for that test to just expand to test *everything* that the method does instead of splitting tests into behavioral units.

PR #261, opened 2015-09-22
refactoring test_subjob.py

ethomas2:


boxcla:

ethomas2:
@josephharrington made the changes you suggested. Since this commit really belongs with the last one, is there a proper way to squash the two that doesn't involve a git push -f to master? The -f seems a little heavy handed

tjlee0909:
-f is what I usually do.

test/unit/master/test_subjob.py
```diff
@@ -36,9 +36,9 @@ def setUp(self):
             ],
         )
 
-    def test_set_atoms_subjob_id(self):
+    def test_subjob_constructor_sets_subjob_id_on_atoms(self):
         atoms = [Mock(), Mock()]
-        self._subjob._set_atoms_subjob_id(atoms, 4)
+        Subjob(1, 4, Mock(), Mock(), atoms)
```
aptxkid: How about using keyword arguments? It makes it much more readable.

aptxkid:
Push -f to master is disruptive to people already have the latest change. I would suggest just merge this PR separately.

test/unit/master/test_subjob.py
ethomas2: whoa, didn't even know you could use keyword arguments unless the function declaration looked like
def f(a=foo, b=bar)

josephharrington:
+1 to just merging separately.

LGTM :+1: :+1: :+1: Thanks for doing this!

aptxkid:
:+1:

PR #264, opened 2015-10-08
Tweak CORS settings that only allows 'GET' call

aptxkid:


boxcla:

tjlee0909:
:+1:

test/unit/web_framework/test_cluster_base_handler.py
```diff
@@ -16,7 +16,7 @@ class TestClusterBaseHandler(BaseUnitTestCase):
         when_cors_conf_value_has_not_been_set=('http://alice.in-wonder.land:8080', None),
         when_request_origin_is_not_allowed=('http://alice.in-wonder.land:8080', 'http://bill\.in-wonder\.land:\d*'),
     )
-    def test_set_default_headers_should_not_set_origin_header(self, request_origin, cors_conf_value):
+    def test_set_default_headers_should_not_set_access_control_headers(self, request_origin, cors_conf_value):
```
josephharrington: Should we modify this test to make sure the new headers (Access-Control-Allow-Headers and Access-Control-Allow-Methods) are also not set?

josephharrington:
:+1:

test/unit/web_framework/test_cluster_base_handler.py
aptxkid: Good catch! Done.

PR #265, opened 2015-10-12
Move job validation and unpacking logic into JobConfig class.

tjlee0909:
Currently the instantiation of the JobConfig object is tied to the ClusterRunnerConfig class. This change moves that construction logic into the JobConfig's class method, to form a sort of factory method.

This change is necessary for an upcoming change where I would like to construct JobConfig objects directly, without having to go through the ClusterRunnerConfig class. The ClusterRunnerConfig class is tied to the yaml format, as it accepts the raw yaml contents of a whole clusterrunner.yaml file as its constructor argument. 

This is problematic for the feature we would like to implement, where you can post the fields of the config via the POST request in a JSON body, not in yaml format.

### DIFF
--- a/app/master/cluster_runner_config.py
+++ b/app/master/cluster_runner_config.py
@@ -1,23 +1,13 @@
-import sys
 import yaml
 
-from app.master.atomizer import Atomizer
 from app.master.job_config import JobConfig
 from app.util import log
 
 
-# clusterrunner.yaml config section names
-SETUP_BUILD = 'setup_build'
-TEARDOWN_BUILD = 'teardown_build'
-COMMANDS = 'commands'
-ATOMIZERS = 'atomizers'
-MAX_EXECUTORS = 'max_executors'
-MAX_EXECUTORS_PER_SLAVE = 'max_executors_per_slave'
-
-
 class ClusterRunnerConfig(object):
-    DEFAULT_MAX_EXECUTORS = sys.maxsize
-
+    """
+    This class represents all of the ClusterRunner job definitions that live inside a single clusterrunner.yaml file.
+    """
     def __init__(self, raw_yaml_contents):
         """
         :param raw_yaml_contents: Raw string contents of project clusterrunner.yaml file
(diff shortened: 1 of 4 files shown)


boxcla:

app/master/cluster_runner_config.py
```diff
@@ -1,23 +1,13 @@
-import sys
```
tjlee0909: The majority of this change is copy/pasting the contents of cluster_runner_config.py to job_config.py.

app/master/job_config.py
```diff
@@ -18,3 +35,119 @@ def __init__(self, name, setup_build, teardown_build, command, atomizer, max_exe
...
+        :type config_dict: dict
+        :rtype: JobConfig
+        """
+        setup_build = cls._shell_command_list_to_single_command(config_dict.get(SETUP_BUILD))
+        teardown_build = cls._shell_command_list_to_single_command(config_dict.get(TEARDOWN_BUILD))
+        command = cls._shell_command_list_to_single_command(config_dict[COMMANDS])
+        atomizer = Atomizer(config_dict[ATOMIZERS])
+        max_executors = config_dict.get(MAX_EXECUTORS, cls.DEFAULT_MAX_EXECUTORS)
+        max_executors_per_slave = config_dict.get(MAX_EXECUTORS_PER_SLAVE, cls.DEFAULT_MAX_EXECUTORS)
+        return JobConfig(name, setup_build, teardown_build, command, atomizer, max_executors, max_executors_per_slave)
```
josephharrington: Minor thing, but I think you can do `return cls(name, setup_build, ...)` here instead of reference the containing classname explicitly.

josephharrington:
Nice. :+1:

app/master/job_config.py
```diff
@@ -18,3 +35,119 @@ def __init__(self, name, setup_build, teardown_build, command, atomizer, max_exe
         self.atomizer = atomizer
         self.max_executors = max_executors
         self.max_executors_per_slave = max_executors_per_slave
+
+    @classmethod
+    def construct_from_dict(cls, name, config_dict):
```
aptxkid: How about just put all the logic in `__init__`? Is there some other places that JobConfig is instantiated directly without using this factory method?

tjlee0909: Has this discussion with Lu offline. Short answer: I could have gone either way, no big deal either way. I do have a habit of not putting real work in the constructor, just from my experience with non-python languages due to testability concerns.

PR #267, opened 2015-10-14
Accept formatted JSON in the build request to specify the job config.

tjlee0909:
An example of what the JSON params look like when making a git-type build request:
```php
$params = [
        'type' => 'git',
        'url' => 'ssh://path.to.source.control/projectname',
        'branch' => 'master',
        'job_name' => 'CustomJobNameDoesntMatter',
        'config' => [
            'setup_build' => [
                'do some setup 1',
                'do some setup 2',
            ],
            'commands' => [
                'run some commands 1',
                'run some commands 2',
            ],
            'atomizers' => [
                [ 'TESTPATH' => 'find $PROJECT_DIR/ -name ...' ]
            ],
            'max_executors' => 100,
        ]
];
```

### DIFF
--- a/app/master/build_request.py
+++ b/app/master/build_request.py
@@ -19,9 +19,18 @@ class BuildRequest(object):
         "type": "git",
         "url": "ssh://github.com/drobertduke/some-project",
         "branch": "master",
-        "job_name": "clusterrunner_configured_job",
+        [OPTIONAL] "job_name": "clusterrunner_configured_job",
+            - This field is optional if the "config" section is specified
         [OPTIONAL] "atoms_override": ['export VAR="overridden_atom_value_1";', ...],
         [OPTIONAL] "hash": "123456789123456789123456789"
+        [OPTIONAL] "config": {
+            "commands" : [...],
+            "atomizers" : {...},
+            "setup_build": [...],
+            "teardown_build": [...],
+            "max_executors": ...,
+            "max_executors_per_slave": ...
+        }
     }
     """
     def __init__(self, build_parameters):
--- a/app/project_type/project_type.py
+++ b/app/project_type/project_type.py
@@ -9,6 +9,7 @@ from threading import Event
 import time
 
 from app.master.cluster_runner_config import ClusterRunnerConfig
+from app.master.job_config import JobConfig
 from app.util import log
 from app.util.conf.configuration import Configuration
 from app.util.process_utils import Popen_with_delayed_expansion, get_environment_variable_setter_command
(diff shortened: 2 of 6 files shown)


boxcla:

tjlee0909:
An example of what the JSON params look like when making a git-type build request:

$params = [
        'type' => 'git',
        'url' => 'ssh://path.to.source.control/projectname',
        'branch' => 'master',
        'job_name' => 'CustomJobNameDoesntMatter',
        'config' => [
            'setup_build' => [
                'do some setup 1',
                'do some setup 2',
            ],
            'commands' => [
                'run some commands 1',
                'run some commands 2',
            ],
            'atomizers' => [
                [ 'TESTPATH' => 'find $PROJECT_DIR/ -name ...' ]
            ],
            'max_executors' => 100,
        ]
];

app/master/build_request.py
```diff
@@ -22,6 +22,14 @@ class BuildRequest(object):
         "job_name": "clusterrunner_configured_job",
```
josephharrington: "job_name" becomes optional if "config" is specified, right?

tjlee0909: Functionally, yeah. But I was thinking about what would display in all of the dashboards we have, and was worried that they will just break. What do you think I should do?

josephharrington: Ah, true. I suppose ClusterRunner could just set the job_name value to some constant like "_adhoc_" if config was specified without a job name.

tjlee0909: I'm actually more worried about the 'job_name' field in the request param, more than the name field in the JobConfig, as that might be the value that some dashboards are utilizing.

That's something that should be fixed though, if it does that.

josephharrington: Ah yeah, that's what I meant. Maybe outside scope for this specific change then.

tjlee0909: That's correct, I'm trying to weasel myself out of more work.

test/functional/job_configs.py
```diff
@@ -24,21 +24,16 @@ def __init__(self, config, expected_to_fail, expected_num_subjobs, expected_num_
...
+        'posix':
+            {
+                'commands': ['echo $TOKEN > $ARTIFACT_DIR/result.txt'],
+                'atomizers': [{'TOKEN': 'seq 0 4 | xargs -I {} echo "This is atom {}"'}]
+            },
+        'nt':
+            {
+                'commands': ['echo !TOKEN!> !ARTIFACT_DIR!\\result.txt'],
+                'atomizers': [{'TOKEN': 'FOR /l %n in (0,1,4) DO @echo This is atom %n'}]
+            },
```
josephharrington: I'd like to keep these functional tests as yaml strings if possible -- I think it's a good thing that the test data is in the raw format. Maybe we should change the framework instead?

tjlee0909: Sure, I can do that.

josephharrington:
:+1:

aptxkid:
:+1:

PR #274, opened 2015-11-30
Build refactor - Use FSM to track build state

josephharrington:
- Add an FSM implementation (based on Fysom) to track the state of a 
build.
- Remove a bunch of flags related to tracking state from the Build
class and use the FSM instead.
- Fix tests.

I've left a bunch of WIP (work-in-progess) notes since this is part of a much larger refactoring and I didn't want to do it all in one commit.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -3,10 +3,10 @@ from enum import Enum
 import os
 from queue import Queue, Empty
 from threading import Lock
-import time
 import uuid
 
 from app.master.build_artifact import BuildArtifact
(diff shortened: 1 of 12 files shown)


boxcla:

josephharrington:
Note that I haven't added any tests since this should be pretty much a pure refactoring of the current build logic. I have also checked change this against our test cluster/job and confirmed that everything still works.

tjlee0909:
lgtm :+1: obviously fix tests.

tjlee0909:
Wunderbar

PR #275, opened 2015-12-13
Bump pylint to 1.5.1

josephharrington:
This fixes currently failing lint jobs. I also regenerated the
pylintrc file (`pylint --generate-rcfile`) and added back some of our
custom rule ignores.

Pylint 1.4.2 has an incompatibility with the newest version of one of
its dependencies (astroid). This has started causing our lint jobs to
fail with a bunch of errors that look like:

```
E: 37, 8: Undefined variable 'self' (undefined-variable)
```

Since we have only specified exact versions of our dependencies in
requirements.txt (and not the exact versions of dependencies'
dependencies, etc.) it's still possible for third party libraries to
break something without a change on our side.

We should switch over to using peep
(https://pypi.python.org/pypi/peep) or something similar.

### DIFF
--- a/app/client/service_runner.py
+++ b/app/client/service_runner.py
@@ -1,6 +1,7 @@
-import requests
 from subprocess import DEVNULL
 import time
+
+import requests
 
 from app.util.conf.configuration import Configuration
 from app.util.log import get_logger
--- a/pylintrc
+++ b/pylintrc
@@ -7,9 +7,6 @@
 # pygtk.require().
 #init-hook=
 
-# Profiled execution.
-profile=no
-
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
 ignore=CVS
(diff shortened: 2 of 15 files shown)


boxcla:

pylintrc
```diff
@@ -38,8 +60,7 @@ load-plugins=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 
-disable=R, missing-docstring, fixme, invalid-name, bad-continuation, attribute-defined-outside-init, arguments-differ,
-    attribute-defined-outside-init, star-args, locally-disabled, unused-argument, unused-variable, no-init,
+disable=R, missing-docstring, fixme, invalid-name, arguments-differ, locally-disabled, unused-argument, unused-variable,
```
josephharrington: Most additions and changes in this file are just new defaults from `pylint --generate-rcfile`. The only changes I made to the new default rcfile were to these disable= rules and some of the flags we had already changed (e.g., colorized output, no full reports).

app/client/service_runner.py
```diff
@@ -1,7 +1,8 @@
-import requests
 from subprocess import DEVNULL
 import time
 
+import requests
+
```
josephharrington: There's a new lint rule `wrong-import-order` that enforces the pep8 recommendation that third party lib imports be separated from stdlib imports. Intellij can fix this automatically with "Optimize Imports" and that's how I fixed all these.

PR #277, opened 2015-12-14
Fix delay in windows functional tests

josephharrington:
The functional tests actually weren't sleeping before. The `ping`
behavior is unintuitive and requires `-n 2` for a 1 second delay.


boxcla:

aptxkid:
Yikes! Thanks, @josephharrington

PR #278, opened 2015-12-14
Add SIGINFO logging when service is hard killed

josephharrington:
The functional testing framework sometimes is unable to kill and
restart the service gracefully in between tests. This change sends a
SIGINFO signal before doing the hard kill. The debug logs triggered by
this signal should give us info about what thread is preventing
graceful shutdown.


boxcla:

tjlee0909:
:+1:

PR #279, opened 2015-12-14
Move pid files to app base directory

josephharrington:
We put the pid files in /tmp/ originally but at least in our
environment they keep getting deleted (surprise!). We don't want these
to be deleted by anyone but clusterrunner, so they shouldn't be in
/tmp/.

### DIFF
--- a/app/util/conf/base_config_loader.py
+++ b/app/util/conf/base_config_loader.py
@@ -32,12 +32,6 @@ class BaseConfigLoader(object):
         conf.set('main_executable_path', sys.argv[0])
         conf.set('version', autoversioning.get_version())
 
-        # Where the clusterrunner service will save the process id to. These settings are set in base_config_loader
-        # and not in master_config_loader and slave_config_loader because CLI tools such as "clusterrunner stop"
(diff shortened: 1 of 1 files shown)


boxcla:

tjlee0909:
I'd almost say just leave out all of the old_* PID id file stuff. Theoretically, they'll just be cleaned up anyhow. Also, we don't use clusterrunner stop anymore in Box's environment, so the switchover/grace period temporary change is unnecessary I think.

Otherwise, :+1:

aptxkid:
I had similar questions about if we need the `old_pid` stuff.

josephharrington:
@tjlee0909 The place I ran into it was in our internal end-to-end test. It failed to deploy the two-node cluster for the test and my guess it was because `clusterrunner stop` didn't work because the pid file had been deleted.

I definitely want to remove the old pid file references soon; I just selfishly wanted to throw that in there to get some practice making backwards compatible changes that require multiple commits. :sunglasses:

tjlee0909:
You do you Joey. :+1:

Issue #263, opened 2015-10-08
When atomizer returns 0 atoms, build hangs


tjlee0909:

tjlee0909:
This might actually not be an issue, going to have to perform more testing later.

josephharrington:
This is definitely an issue. I can reproduce locally on my mac with 
```bash
cd clusterrunner
./main.py master & ./main.py slave &
./main.py build directory --project-directory . --job-name Blah
```
where job "Blah" looks like:
```yaml
Blah:
    commands:
        - echo blahhhh
    atomizers:
        - ATOM: find . -name "something_nonexistent"
```

josephharrington:
The current behavior is that the build gets stuck in the `QUEUED` state forever. It makes it into the slave allocation loop but immediately exits since it has no subjobs. Currently the only way a build can leave `QUEUED` is if it is canceled, failed, or starts building (which only happens when slave is allocated) .

We should either fail the build, or move it immediately to a finished state. We could do this in the `BuildRequestHandler` immediately after preparing the build -- just don't put a build with zero subjobs into the `_builds_waiting_for_slaves` queue.

tjlee0909:
Fixed!

PR #283, opened 2016-01-01
Fail build immediately if no atoms found.

tjlee0909:
This commit addresses https://github.com/box/ClusterRunner/issues/263

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -143,6 +143,8 @@ class Build(object):
             raise RuntimeError('Build failed while trying to parse clusterrunner.yaml.')
 
         subjobs = subjob_calculator.compute_subjobs_for_build(self._build_id, job_config, self.project_type)
+        if len(subjobs) == 0:
+            raise RuntimeError('Atomizer for build found no atoms to run.')
 
         self._unstarted_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Move this into BuildScheduler?
         self._finished_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Remove this and just record finished count.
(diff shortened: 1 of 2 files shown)


boxcla:

josephharrington:
Nice! This is pretty clean. What were your thoughts on failing a build with no atoms vs. letting the build pass immediately? The argument in favor of passing the build is that the atomizer command was technically successful -- zero exit code -- but (maybe just for this specific build) it didn't find any atoms.

I don't think we have a good use case in production right now, but here's a contrived one: say I wanted to set up a clusterrunner job to run just on changed php files. The atomizer could be something like:
```yaml
TESTPATH: git diff --name-only HEAD~1 | grep .php
```
In a case like that I'd probably want the build to succeed even if there were no atoms. 

On the other hand I'd agree that if someone has made a mistake in their atomizer command that they might unknowingly create a job that does nothing, so failing the build is an easy way to alert for that type of mistake. I'd hope that someone creating a job for the first time is checking more than just if the build was successful or not, but it's probably not clear for new users how to do that.

I was curious about your thoughts on the above. :+1: in general.

tjlee0909:
I agree, technically the build didn't fail. And I am also super worried that no one will parse the build results other than the failed_tests file. I decided that this is the lesser of two evils.

At the moment, the contrived case you have is an actual production use case right now. All of the 'PHPUnit-Enumerated' type jobs (there are multiple of them) parse the HEAD commit modified files in the atomization command. As you can guess, when such a job runs with the current CR version, the build hangs and the corresponding jenkins job times out. We have currently gotten around that ATM by running the atomizer command in a separate call before invoking CR, and just exiting 0 if no tests to run were found.

tjlee0909:
Ok, had a separate conversation with Joey, and letting the build pass (in the case of no atoms found) sounds like the correct way to go. The client should be aware that, for this ClusterRunner job configuration it is possible to have 0 atoms run, and if they want to fail the build in such a scenario, then the client can check the number of subjob result directories inside the returned result tarball.

I will close this PR and create a new one that passes the build.

Issue #273, opened 2015-11-23
Build scheduler allocates too many slaves to builds with few atoms


josephharrington:
If a build has a single atom, it is very likely to get more than one slave assigned.

There is this method in `BuildScheduler`:
```python
def needs_more_slaves(self):
        """Determine whether or not this build should have more slaves allocated to it."""
        return self._num_executors_allocated < self._max_executors and not self._build._unstarted_subjobs.empty()
```

This will correctly stop allocating slaves once all subjobs have been started. But because build setup takes at least a few seconds, it's very likely that many slaves will be allocated before any subjobs start. So builds with very few atoms are likely to have more slaves allocated than necessary.

This is not a huge problem, but is inefficient and *could* be a huge problem for builds with a long setup time.

tjlee0909:
I see, so this seems to be a problem when:

- max_executors is set to something high, lets say 100.
- The number of subjobs/atoms is small, lets say 1.

I think we should just do a check on the total number of subjobs vs number of executors allocated, and limit it based on that (in addition to what we have). Anything more complicated is probably not worth the effort and additional complexity we'd introduce (e.g.: handling the case where one slave is burning through subjobs, but 10 other slaves are taking forever to clone the repo).

josephharrington:
Yep, totally agree with the check you suggest. There's no point in allocating more executors than subjobs.

Just for reference, another thing that made this more likely was that the default value of `max_executors` is `sys.maxsize`.

PR #284, opened 2016-01-01
Don't allocate more executors to a build than total subjobs

tjlee0909:
This addresses https://github.com/box/ClusterRunner/issues/273

### DIFF
--- a/app/master/build_scheduler.py
+++ b/app/master/build_scheduler.py
@@ -49,7 +49,13 @@ class BuildScheduler(object):
 
         :rtype: bool
         """
-        return self._num_executors_allocated < self._max_executors and not self._build._unstarted_subjobs.empty()
+        if self._num_executors_allocated >= self._max_executors:
+            return False
+        if self._build._unstarted_subjobs.empty():
+            return False
+        if self._num_executors_allocated >= len(self._build.all_subjobs()):
+            return False
+        return True
 
     def allocate_slave(self, slave):
         """
--- a/test/unit/master/test_build.py
+++ b/test/unit/master/test_build.py
@@ -181,16 +181,18 @@ class TestBuild(BaseUnitTestCase):
             build.complete_subjob(subjob.subjob_id(), payload=self._FAKE_PAYLOAD)
 
     @genty_dataset(
-        max_executors_reached=(1, False),
-        max_executors_not_reached=(30, True),
+        max_executors_reached=(1, False, 100),
+        max_executors_not_reached=(30, True, 100),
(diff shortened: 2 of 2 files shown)


boxcla:

app/master/build_scheduler.py
```diff
@@ -49,7 +49,13 @@ def needs_more_slaves(self):
...
         :rtype: bool
         """
-        return self._num_executors_allocated < self._max_executors and not self._build._unstarted_subjobs.empty()
+        if self._num_executors_allocated >= self._max_executors:
+            return False
+        if self._build._unstarted_subjobs.empty():
+            return False
+        if self._num_executors_allocated >= len(self._build.all_subjobs()):
+            return False
+        return True
```
tjlee0909: I thought that breaking up the conditionals was much easier to read than all of the 'ands' and 'nots' chained together.

tjlee0909:
I posted a comment about this in https://github.com/box/ClusterRunner/issues/273 , but to reiterate: 

I'm going for low-hanging fruit here with the somewhat common manifestation of this bug where the max_executors is set to much higher than the actual number of subjobs calculated by the atomizer.

app/master/build_scheduler.py
```diff
@@ -49,7 +49,13 @@ def needs_more_slaves(self):
 
         :rtype: bool
         """
-        return self._num_executors_allocated < self._max_executors and not self._build._unstarted_subjobs.empty()
+        if self._num_executors_allocated >= self._max_executors:
+            return False
+        if self._build._unstarted_subjobs.empty():
+            return False
+        if self._num_executors_allocated >= len(self._build.all_subjobs()):
```
tjlee0909: Note to self: maybe should compare to self._build._unstarted_subjobs.qsize() instead.

josephharrington: I think I actually prefer checking the total number of subjobs. What were you thinking would be the advantage of checking `_unstarted_subjobs.qsize()`?

josephharrington:
Hallelujah! :+1:

app/master/build_scheduler.py
tjlee0909: In the scenario where the slaves being allocated are trickling in (being released from other builds onto this one), then the length of all subjobs is an overestimate, whereas the unstarted_subjobs.qsize() is more of an exact fit.

It's really not a big deal though. I'm happy with leaving it as is and merging it.

josephharrington: Yeah agreed / not a big deal. :metal:

PR #286, opened 2016-01-08
If build has no atoms, immediately pass the build.

tjlee0909:
Addresses https://github.com/box/ClusterRunner/issues/263

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -115,10 +115,6 @@ class Build(object):
         if self._project_type is None:
             raise BuildProjectError('Build failed due to an invalid project type.')
 
-        # Create the build results directory.
-        if not os.path.exists(self._build_results_dir()):
-            os.mkdir(self._build_results_dir())
-
     def prepare(self, subjob_calculator):
         """
         :param subjob_calculator: Used after project fetch to atomize and group subjobs for this build
@@ -155,7 +151,7 @@ class Build(object):
             self._unstarted_subjobs.put(subjob)
 
         self._timing_file_path = self._project_type.timing_file_path(job_config.name)
-
+        app.util.fs.create_dir(self._build_results_dir())
         self._state_machine.trigger(BuildEvent.FINISH_PREPARE)
 
     def build_id(self):
@@ -257,13 +253,20 @@ class Build(object):
         # We use a local variable here which was set inside the _build_completion_lock to prevent a race condition
         if should_trigger_postbuild_tasks:
             self._logger.info("All results received for build {}!", self._build_id)
-            SafeThread(target=self.perform_async_postbuild_tasks, name='PostBuild{}'.format(self._build_id)).start()
+            SafeThread(target=self._perform_async_postbuild_tasks, name='PostBuild{}'.format(self._build_id)).start()
 
     def mark_started(self):
         """
         Mark the build as started.
         """
         self._state_machine.trigger(BuildEvent.START_BUILDING)
+
+    def finish(self):
+        """
+        Perform postbuild task and mark this build as finished.
+        """
+        # This method also transitions the FSM to finished after the postbuild tasks are complete.
+        self._perform_async_postbuild_tasks()
 
     def mark_failed(self, failure_reason):
         """
@@ -434,7 +437,7 @@ class Build(object):
             return BuildResult.FAILURE
         return None
 
-    def perform_async_postbuild_tasks(self):
+    def _perform_async_postbuild_tasks(self):
         """
         Once a build is complete, certain tasks can be performed asynchronously.
         """
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -112,9 +112,12 @@ class Build(object):
         project_type_params = self.build_request.build_parameters()
         project_type_params.update({'build_project_directory': build_specific_project_directory})
         self._project_type = util.create_project_type(project_type_params)
-
         if self._project_type is None:
             raise BuildProjectError('Build failed due to an invalid project type.')
+
+        # Create the build results directory.
+        if not os.path.exists(self._build_results_dir()):
+            os.mkdir(self._build_results_dir())
 
     def prepare(self, subjob_calculator):
         """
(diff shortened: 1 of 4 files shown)


boxcla:

tjlee0909:
Looking for feedback before writing tests.

app/master/build.py
```diff
@@ -112,10 +112,13 @@ def generate_project_type(self):
         project_type_params = self.build_request.build_parameters()
         project_type_params.update({'build_project_directory': build_specific_project_directory})
         self._project_type = util.create_project_type(project_type_params)
-
         if self._project_type is None:
             raise BuildProjectError('Build failed due to an invalid project type.')
 
+        # Create the build results directory.
+        if not os.path.exists(self._build_results_dir()):
+            os.mkdir(self._build_results_dir())
```
tjlee0909: I added this here because, in the scenario when there are no atoms and we want to immediately perform the aynchronoous post-build actions, we need to first tar-up the build results directory. But because the build results directory is usually only generated as we receive subjob results from slaves, that errors out.

I figured this is the place where it makes the most sense to generate the initial empty build results directory (generate_project_type).

app/master/build_request_handler.py
```diff
@@ -101,6 +101,11 @@ def _prepare_build_async(self, build, project_lock):
                                    log_msg='Build preparation loop is handling request for build {build_id}.')
             try:
                 build.prepare(self._subjob_calculator)
+                # If the atomizer found no work to do, perform build cleanup and skip the slave allocation.
+                if len(build.all_subjobs()) == 0:
+                    self._logger.info('Build {} has no work to perform and is exiting.', build.build_id())
+                    build.mark_started()
```
tjlee0909: If this line isn't here, the build_fsm raises an exception when the perform_async_postbuild_tasks() tries to progress the FSM to the completed state.

We have two options here:

- either modify the FSM rules to allow the build to go from PREPARED -> FINISHED.
- or do what I did here, where we mark the build as started immediately before performing the asynchronous postbuild tasks.

app/master/build_request_handler.py
```diff
@@ -101,6 +101,11 @@ def _prepare_build_async(self, build, project_lock):
...
+                    build.perform_async_postbuild_tasks()
```
tjlee0909: I was contemplating creating a "mark_completed()" function in the Build class, but it would end up entirely just being a wrapp for the perform_async_postbuild_tasks() invocation, so I decided to make this method public instead.

josephharrington: +1 to adding something like `mark_completed()` or `finish()`. Performing postbuild tasks feels like an internal decision that the build should be making.

Also, ideally I'd rather us add the PREPARED -> FINISHED transition. Ideally we'd just call something like `finish()` on the build

I'm actually also fine with what you have right now. I still need to finish up a bunch of the work-in-progress notes I left around the FSM, which involves cleaning up the public interface anyway.

tjlee0909: Ok, I can create a finish() public method.

Also, could you tell me later how to regenerate that FSM diagram?

app/master/build.py
josephharrington: I think there is a helper to create a directory (`app.util.fs.create_dir`) that be a bit cleaner?

Also, I'm curious about your thoughts behind putting result dir creation in `generate_project_type()` (vs. e.g., `prepare()`).

app/master/build_request_handler.py
josephharrington: Looking at this more -- I think I'd rather us keep `perform_async_postbuild_tasks` private. I think it was written intentionally to be executed on its own thread (started internally by the build). It feels a bit weird to execute it in one context on its own thread asynchronously and in another context synchronously, blocking other builds from being prepared.

I'm not really concerned about the *performance* of blocking other builds from preparing while doing postbuild tasks since this would only happen on builds with no subjobs, so no results, and the postbuild would be very fast. It's more of a conceptual thing.

The most direct way to solve that is probably to move the part in `Build` that starts the PostBuild thread into its own method, and then call that here instead.

josephharrington: And haha, I just created that FSM diagram manually. "Column Selection Mode" in IntelliJ is your best friend for that.

josephharrington: Here's a swing at adding the PREPARED -> FINISHED transition. https://gist.github.com/josephharrington/337baca2c80f072a57f7

josephharrington: Another option, more along the lines of what you had originally, would be inside the build's `finish()` method to just progress to the BUILDING state first if the build is in the PREPARED state. I'm not opposed to that either.

app/master/build.py
tjlee0909: prepare() is objectively the better place to put it, as the build results directory has nothing to do with the project type.

app/master/build_request_handler.py
tjlee0909: I think I'll add the PREPARED -> FINISHED transition. It feels like lying to say the build is BUILDING, when we really are going to from preparation to completion.

tjlee0909:
Going to write some tests. Meanwhile, does that match what you were thinking Joey? I just functionally tested it in our test cluster, and it passes a NoAtom build.

app/master/build_request_handler.py
```diff
@@ -101,6 +101,10 @@ def _prepare_build_async(self, build, project_lock):
                                    log_msg='Build preparation loop is handling request for build {build_id}.')
             try:
                 build.prepare(self._subjob_calculator)
+                # If the atomizer found no work to do, perform build cleanup and skip the slave allocation.
+                if len(build.all_subjobs()) == 0:
+                    self._logger.info('Build {} has no work to perform and is exiting.', build.build_id())
+                    build.finish()
```
josephharrington: Do we want to `finish()` even if the build `has_error`?

josephharrington:
Looks good to me!

app/master/build_request_handler.py
tjlee0909: No, we don't. Thanks for pointing that out.

app/master/build_request_handler.py
```diff
@@ -102,6 +102,10 @@ def _prepare_build_async(self, build, project_lock):
             try:
                 build.prepare(self._subjob_calculator)
                 if not build.has_error:
+                    # If the atomizer found no work to do, perform build cleanup and skip the slave allocation.
+                    if len(build.all_subjobs()) == 0:
+                        self._logger.info('Build {} has no work to perform and is exiting.', build.build_id())
+                        build.finish()
```
josephharrington: I know you already merged but I just noticed this -- should this return after calling finish()?

josephharrington: Or maybe always log the build_prepare_finish event but just not stick it on the queue?

tjlee0909: Dangit! Sorry, I've been pretty distracted lately. I'll fix that right now, this is bad.

PR #297, opened 2016-02-26
Add Prometheus instrumentation

gcurtis:
This change uses the Prometheus client SDK to add some instrumentation
to the master. So far, only HTTP request latency is instrumented.

There is also a new endpoint at /v1/metrics which exposes the metrics in
in Prometheus text exposition format. The metrics endpoint required a
new method in ClusterBaseHandler so that a raw text string could be
written.

### DIFF
--- a/app/web_framework/cluster_base_handler.py
+++ b/app/web_framework/cluster_base_handler.py
@@ -103,6 +103,10 @@ class ClusterBaseAPIHandler(ClusterBaseHandler):
             response['child_routes'] = self.get_child_routes()
         super().write(response)
 
+    def write_text(self, response):
+        super().set_header('Content-Type', 'text/plain; charset=utf-8')
+        super().write(response)
+
     def _write_status(self, additional_response=None, success=True, status_code=200):
         status = self.SUCCESS_STATUS if success else self.FAILURE_STATUS
         response = {'status': status}
--- a/app/web_framework/cluster_master_application.py
+++ b/app/web_framework/cluster_master_application.py
@@ -4,6 +4,9 @@ import urllib.parse
 
 import tornado.web
 
+import prometheus_client
+from prometheus_client import Histogram
+
 from app.util import analytics
 from app.util import log
 from app.util.conf.configuration import Configuration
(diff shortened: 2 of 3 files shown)


boxcla:
Verified that @gcurtis has signed the CLA. Thanks for the pull request!

app/web_framework/cluster_master_application.py
```diff
@@ -19,6 +22,9 @@
 # pylint: disable=attribute-defined-outside-init
 #   Handler classes are not designed to have __init__ overridden.
 
+request_latency = Histogram('request_latency_seconds', 'Latency of HTTP requests in seconds')  # pylint: disable=no-value-for-parameter
```
gcurtis: The `_MetricWrapper` decorator in the Prometheus client SDK seemed to be throwing off pylint.

app/web_framework/cluster_master_application.py
```diff
@@ -33,6 +39,7 @@ def __init__(self, cluster_master):
         root_route = \
             RouteNode(r'/', _RootHandler).add_children([
                 RouteNode(r'v1', _APIVersionOneHandler).add_children([
+                    RouteNode(r'metrics', _MetricsHandler),
```
gcurtis: We may want to have this endpoint disabled by default and enabled with a command line option. I'm not sure what the cleanest way of doing that is.

josephharrington: Maybe we could just leave the endpoint active in either case, and if it's turned off just have it return an error code.

app/web_framework/cluster_base_handler.py
```diff
@@ -103,6 +103,10 @@ def write(self, response):
             response['child_routes'] = self.get_child_routes()
         super().write(response)
 
+    def write_text(self, response):
+        super().set_header('Content-Type', 'text/plain; charset=utf-8')
```
josephharrington: Could just do `self.set_header()` instead of `super()`.

josephharrington:
I can dig it. :+1: We could do the conf switch in a separate change.

Looks like it passed the functional tests so it's probably fine, but if you want want extra testing we have an internal test cluster that TJ set up that you can run it through.

tjlee0909:
Where does the data for these metrics get stored? Doing this sounds great, but out of curiosity, was there a particular reason why you are doing this now?

gcurtis:
@tjlee0909 the metrics are stored in-memory and then collected by a Prometheus server. The server gets the metrics by periodically scraping the /v1/metrics endpoint.

For your second question - I wanted to a get a proof of concept working and ClusterRunner seemed like a good choice :-)

tjlee0909:
k.

:+1:

gcurtis:
Closing this for now until we have time to revisit it.

Issue #298, opened 2016-03-22
Delete artifact directory for build on master after artifact tarball is created


tjlee0909:
The Cluster-Master occasionally runs out of inodes due to the massive amount of files in the .clusterrunner/master/artifacts/ directory and subdirectories.

Fortunately, these files and directories are no longer needed after generating the build artifact. We should delete this as a part of async-postbuild tasks.

josephharrington:
Woohoo!

PR #299, opened 2016-03-22
Delete artifact directory for build on master after completion.

tjlee0909:
That directory's contents are no longer needed once the build artifact tarball has been created, and is just taking up disk space and inodes on the master at that point.

https://github.com/box/ClusterRunner/issues/298

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from enum import Enum
 import os
 from queue import Queue, Empty
+import shutil
 from threading import Lock
 import uuid
 
@@ -442,6 +443,7 @@ class Build(object):
         Once a build is complete, certain tasks can be performed asynchronously.
         """
         self._create_build_artifact()
+        self._delete_temporary_build_artifact_files()
         self._postbuild_tasks_are_finished = True
         self._state_machine.trigger(BuildEvent.POSTBUILD_TASKS_COMPLETE)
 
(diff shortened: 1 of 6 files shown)


boxcla:

wjdhollow:
:shipit:

app/master/build.py
```diff
@@ -451,6 +453,25 @@ def _create_build_artifact(self):
...
+        for path in os.listdir(build_result_dir):
+            # The build result tar-ball (results.tar.gz) is also stored in this same directory, so we must
+            # skip deleting it.
+            if path == 'results.tar.gz':
+                continue
+            full_path = os.path.join(build_result_dir, path)
+            if os.path.isdir(full_path):
+                shutil.rmtree(full_path)
+            else:
+                os.remove(full_path)
```
josephharrington: What do you think of using `fs.async_delete` or some other approach that can remove the files after the build finishes? It seems like these rms can be kind of slow and we may not want to block the build from completing until it finishes. (`_perform_async_postbuild_tasks` currently has to complete before we mark the build as finished.)

tjlee0909: I originally hesitated to do so, because the postbuild tasks are already async. I didn't consider the fact that the build isn't marked complete until the async postbuild tasks are also complete.

So, sounds good to me.

josephharrington: Yeah, I was reading through the code thinking it's a little misleading for us to call them async postbuild tasks. They're async from the perspective of the last slave to report results; that slave doesn't have to wait for postbuild tasks to finish. But from the perspective of the build, they are definitely synchronous since the build is blocked until they complete.

tjlee0909:
Verified functionally on test cluster that behavior is as expcted.

test/unit/master/test_build.py
```diff
@@ -429,6 +430,15 @@ def test_get_failed_atoms_returns_failed_atoms_only(self):
             build._all_subjobs_by_id[3]._atoms[3],
         ])
 
+    def test_delete_temporary_build_artifact_files_skips_results_tarball(self):
+        build = self._create_test_build(BuildStatus.BUILDING)
```
josephharrington: Would creating a build with FINISHED status let you avoid needing to call the private `_delete_temporary_build_artifact_files` directly?

app/master/build.py
```diff
@@ -451,6 +452,21 @@ def _create_build_artifact(self):
...
+        Delete the temporary build result files that are no longer needed, due to the creation of the
+        build artifact tarball (results.tar.gz).
+
+        ONLY call this method after _create_build_artifact() has completed. Otherwise we have lost the build results.
+        """
+        build_result_dir = self._build_results_dir()
+        for path in os.listdir(build_result_dir):
+            # The build result tar-ball (results.tar.gz) is also stored in this same directory, so we must
+            # skip deleting it.
+            if path == 'results.tar.gz':
```
josephharrington: Should we make this a constant somewhere? Not like we'll change it but it might make it slightly easier to track references.

josephharrington:
A few tiny comments, but overall looks great! :+1:

app/master/build.py
tjlee0909: Yeah, I was thinking that.

test/unit/master/test_build.py
tjlee0909: I don't think I understand what you're saying. Do you mean call the public 'finish()' method instead?

josephharrington: I mean like 
`build = self._create_test_build(BuildStatus.FINISHED, ...`
instead of 
`build = self._create_test_build(BuildStatus.BUILDING, ...`

josephharrington: Not sure if that would actually work, just wondering...

PR #301, opened 2016-03-28
Serially delete directories and files in build results.

tjlee0909:
Previously, this was deleted asynchronously. While this sounds okay, it actually causes issue at scale with the number of file descriptors that can be open at a given time for builds with large numbers of atoms.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -2,6 +2,8 @@ from collections import OrderedDict
 from enum import Enum
 import os
 from queue import Queue, Empty
+import shutil
+import time
 from threading import Lock
 import uuid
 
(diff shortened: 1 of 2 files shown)


boxcla:

app/master/build.py
```diff
@@ -461,11 +462,19 @@ def _delete_temporary_build_artifact_files(self):
...
                 continue
-            app.util.fs.async_delete(os.path.join(build_result_dir, path))
+            full_path = os.path.join(build_result_dir, path)
+            # Do NOT use app.util.fs.async_delete() here. That call will generate a temp directory for every
+            # atom, which can be in the thousands per build, and can lead to running up against the ulimit -Hn.
+            if os.path.isdir:
+                shutil.rmtree(full_path, ignore_errors=True)
+            else:
+                os.remove(full_path)
+        self._logger.info('Completed deleting artifact files for {}.', self._build_id)
```
josephharrington: How do you feel about instead of a started and completed log if we just measured the time it takes and log that?

tjlee0909: Sounds like a good idea.

josephharrington:
:+1: sounds good.

PR #303, opened 2016-04-04
Perform shallow clones and fetches on both slave and master.

tjlee0909:
This change is necessary in order to improve CR performance, as well as to reduce the disk footprint of our VM images.

This should be backwards compatible for masters/slaves that already have non-shallow clones.


boxcla:

josephharrington:
:+1:

PR #304, opened 2016-04-04
Have the master always fetch and clone deep, but slave shallow.

tjlee0909:
Turns out you cannot fetch from a shallow repo in the older versions of git. What this means is that for now, we will have to continue to fetch and clone regularly on the master only.


boxcla:

josephharrington:
Looks like tests maybe failed but :+1: to approach.

Issue #296, opened 2016-02-22
Reconnecting a slave can take down the master


josephharrington:
This occurs when an idle slave is hard killed (SIGKILL / kill -9) and then reconnected. (As opposed to a graceful kill with SIGTERM / kill -15.) Local reproduction steps are below.

I believe the cause is that when we connect a slave, we detect if it's a "reconnect" meaning a slave with the same host name and port has been connected before. We do some work to clean up the old slave instance but it is incomplete. 

We remove the slave from `self._all_slaves_by_url` but if it is idle, that instance is still in the SlaveAllocator's idle slaves queue so it can still be assigned to a build. We should probably just call `self._disconnect_slave()` and let that method handle cleaning up the old slave.

Reproduction steps:

```bash
./main.py master &
./main.py slave  --num-executors 5 &
kill -9 `cat ~/.clusterrunner/.clusterrunner_slave.pid`
./main.py slave  --num-executors 5 &

# At this point you'll see the "Removing existing slave instance from the master's bookkeeping." log message.

# Now start two concurrent builds...
./main.py build directory --project-directory examples/directory\ job/ & 
./main.py build directory --project-directory examples/directory\ job/ &

# At this point the master will die with "Attempted to start subjob 0 for build 1, but current build id is 2."
# The slave will keep running. Both build commands will eventually time out since they can't contact the master.
```

tjlee0909:
Nice debugging work there.

josephharrington:
Just for a bit more background, there's a line in the SlaveAllocator where we discard slaves from the idle slave queue:
```python
                # Remove dead and shutdown slaves from the idle queue
                if claimed_slave.is_shutdown() or not claimed_slave.is_alive(use_cached=False):
                    continue
```

The semantics around `slave.is_alive()` are a bit weird to me -- I don't think consuming code should need to care about when it's using a cached value vs. a non-cached value since most of the time that decision requires more context than consuming code should care about.

I was thinking of refactoring that to something like `slave.can_accept_builds()` and then the slave could internally take care of deciding whether to use cached values or not. Just a thought -- it might make it easier to handle the case when a slave is reconnected.

[...]

PR #305, opened 2016-04-09
Add functional test for slave reconnection issue

josephharrington:
The added test in this commit is marked skipped for now due to #296
but will be unskipped in an upcoming commit when that issue is
fixed.

The functional test ensures a slave can die ungracefully and after
that host reconnects that everything will still work as expected.

### DIFF
--- a/app/master/slave.py
+++ b/app/master/slave.py
@@ -29,6 +29,9 @@ class Slave(object):
         self._slave_api = UrlBuilder(slave_url, self.API_VERSION)
         self._logger = log.get_logger(__name__)
 
+    def __str__(self):
+        return '<slave #{} - {}>'.format(self.id, self.url)
+
     def api_representation(self):
         return {
             'url': self.url,
--- a/test/functional/test_cluster_basic.py
+++ b/test/functional/test_cluster_basic.py
@@ -33,23 +33,7 @@ class TestClusterBasic(BaseFunctionalTestCase):
         master.block_until_build_finished(build_id, timeout=30)
         slave.block_until_idle(timeout=20)  # ensure slave teardown has finished before making assertions
 
-        if test_job_config.expected_to_fail:
-            self.assert_build_has_failure_status(build_id=build_id)
-        else:
-            self.assert_build_has_successful_status(build_id=build_id)
-
-        self.assert_build_status_contains_expected_data(
-            build_id=build_id,
-            expected_data={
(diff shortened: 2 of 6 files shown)


boxcla:

app/master/slave.py
```diff
@@ -29,6 +29,9 @@ def __init__(self, slave_url, num_executors):
         self._slave_api = UrlBuilder(slave_url, self.API_VERSION)
         self._logger = log.get_logger(__name__)
 
+    def __str__(self):
+        return '<slave #{} - {}>'.format(self.id, self.url)
```
tjlee0909: Nice.

test/functional/test_cluster_basic.py
```diff
@@ -82,3 +66,55 @@ def test_git_type_demo_project_config(self):
...
+        test_config = JOB_WITH_SETUP_AND_TEARDOWN
+        job_config = yaml.safe_load(test_config.config[os.name])['JobWithSetupAndTeardown']
+        master = self.cluster.start_master()
+
+        # Start a slave, hard kill it, then reconnect it.
+        self.cluster.start_slave(num_executors_per_slave=5, start_port=43001)
+        self.cluster.kill_slaves(kill_gracefully=False)
+
+        # Make sure the slave restarts with the same port.
+        slave = self.cluster.start_slave(num_executors_per_slave=5, start_port=43001)
```
tjlee0909: So this test, if run on the current codebase, takes down the master? I thought the problem occurred when the slave is running BUILD_A, and then gets a setup request sent from the master for BUILD_B before BUILD_A has completed.

josephharrington: Yep you're right -- and the reason the master sends a setup command to a slave already running a build is because the master actually has two `Slave` instances for the same host in the idle slaves queue. The master treats them as different slaves and so assigns them to different builds. This occurs when an idle slave goes offline without doing the "disconnect" call to the master.

This test does the same thing as the repro steps I posted in the original issue. I think there are much much smaller race conditions possible that would cause the same issue if the slave was killed and restarted at the exact right moment, but those seem very hard to hit compared to this one.

tjlee0909: Oohh, I see. Ok, that's why you started two builds.

tjlee0909:
:+1: digging the functional TDD

PR #306, opened 2016-04-10
Enable Session-Id headers for slaves

josephharrington:
The master already returns a session id in response headers, and the
API will raise a 412 if the Session-Id request header specifies a
differing session id.

This change moves that logic one level up so that slaves will also
start generating and returning a slave-specific session id in their
responses. The master now also tracks the session id for each slave.

A subsequent change will add verification when the master sends
requests to the slave to ensure that the master is talking to the
correct slave instance.

### DIFF
--- a/app/web_framework/cluster_master_application.py
+++ b/app/web_framework/cluster_master_application.py
@@ -8,8 +8,7 @@ from app.util import analytics
 from app.util import log
 from app.util.conf.configuration import Configuration
 from app.util.decorators import authenticated
-from app.util.exceptions import ItemNotFoundError, PreconditionFailedError
-from app.util.session_id import SessionId
+from app.util.exceptions import ItemNotFoundError
 from app.util.url_builder import UrlBuilder
 from app.web_framework.cluster_application import ClusterApplication
 from app.web_framework.cluster_base_handler import ClusterBaseAPIHandler, ClusterBaseHandler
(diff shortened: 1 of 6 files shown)


boxcla:

app/web_framework/cluster_master_application.py
```diff
@@ -296,7 +275,8 @@ class _SlavesHandler(_ClusterMasterBaseAPIHandler):
     def post(self):
         slave_url = self.decoded_body.get('slave')
         num_executors = int(self.decoded_body.get('num_executors'))
-        response = self._cluster_master.connect_slave(slave_url, num_executors)
+        session_id = self.decoded_body.get('session_id')
```
tjlee0909: Does this method return None if the session_id doesn't exist?

tjlee0909:
Cool, assuming the comment I left is irrelevant, LGTM

app/web_framework/cluster_master_application.py
tjlee0909: Just want to make sure that if we were to deploy the master first before the slaves, that it doesn't blow up.

josephharrington: Yep, `get` will return None so this should be backward compatible (the master would work with old slave versions). That should also be true with the next change that adds session id validation.

I just realized we should probably do more validation on this api data though. I'll add a todo.

PR #307, opened 2016-04-11
Verify slave session id on is_alive check

### LINKED ISSUE #296: Reconnecting a slave can take down the master

josephharrington:
Previously when the master checked if a slave is alive, it only
checked if the slave service is responsive at that url. This change
makes the master also verify that the slave service that responds at
that url still has the same session id that the slave reported when
it connected to the master.

This will cause the `slave.is_alive()` check on the master to start
returning false if the slave responds but has the wrong session id.
Previously this would return true.

This is done by setting a header in the master's request to the
slave that includes the expected session id. If the slave sees this
and it does not match its current session id, it will return a 412
and the master will consider the slave with the old session id to be
offline.

This fixes #296. Now when the slave allocator does the `is_alive()`
call before allocating a slave to a build, it will correctly detect
when a slave's session id has changed and should be considered
offline.

### DIFF
--- a/app/util/session_id.py
+++ b/app/util/session_id.py
@@ -2,7 +2,9 @@ import uuid
 
 
 class SessionId(object):
+    EXPECTED_SESSION_HEADER_KEY = 'Expected-Session-Id'
     SESSION_HEADER_KEY = 'Session-Id'
+
     _session_id = None
 
     @classmethod
--- a/test/functional/test_cluster_basic.py
+++ b/test/functional/test_cluster_basic.py
@@ -67,7 +67,6 @@ class TestClusterBasic(BaseFunctionalTestCase):
         self.assert_build_artifact_contents_match_expected(
             build_id=build_id, expected_build_artifact_contents=expected_artifact_contents)
 
-    @skip('This test currently fails due to https://github.com/box/ClusterRunner/issues/296.')
     def test_slave_reconnection_does_not_take_down_master(self):
         test_config = JOB_WITH_SETUP_AND_TEARDOWN
         job_config = yaml.safe_load(test_config.config[os.name])['JobWithSetupAndTeardown']
--- a/app/master/slave.py
+++ b/app/master/slave.py
@@ -5,6 +5,7 @@ from app.util.counter import Counter
 from app.util.network import Network
 from app.util.safe_thread import SafeThread
 from app.util.secret import Secret
+from app.util.session_id import SessionId
 from app.util.url_builder import UrlBuilder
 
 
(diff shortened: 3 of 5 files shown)


boxcla:

app/util/session_id.py
```diff
@@ -2,7 +2,9 @@
 
 
 class SessionId(object):
+    EXPECTED_SESSION_HEADER_KEY = 'Expected-Session-Id'
```
josephharrington: I added this as a way to more clearly distinguish between the request header and the response header. Previously a "Session-Id" request header was set to indicate the *expected* session id, and a "Session-Id" response header was set to communicate what was the actual current session id.

It's not too bad but I was thinking using a different header name would be clearer. This is also closer to the pattern followed by some standard headers such as "Accept" and "Content-Type" -- the first one is something you'd put in a request to declare what types you accept, and the second is what you'd find in the response to declare what type it is.

test/functional/test_cluster_basic.py
```diff
@@ -67,7 +67,6 @@ def test_git_type_demo_project_config(self):
         self.assert_build_artifact_contents_match_expected(
             build_id=build_id, expected_build_artifact_contents=expected_artifact_contents)
 
-    @skip('This test currently fails due to https://github.com/box/ClusterRunner/issues/296.')
```
josephharrington: Woo!

tjlee0909:
Took me a while to understand the expected vs sent session ids, but makes sense. :+1:

josephharrington:
@tjlee0909 Which part was unclear? Can I change something so it's more intuitive?

tjlee0909:
Nah, I think just reading through the code, sometimes it's hard to get lost between the slave-side, master-side, and the shared-side.

Issue #313, opened 2016-06-20
Invalid transition when build gets slaves, loses slaves, and gets slaves again


tjlee0909:
[15:35:27] ClusterRunner encountered an error processing build 2:
[15:35:27]  >>> 'Error during build state transition. (FysomError: event BuildEvent.START_BUILDING inappropriate in current state BuildState.BUILDING)'

This occurs when a build is going with slaves, and then all slaves assigned to that build are disconnected, and then some slaves are added back again.

PR #314, opened 2016-06-22
Move _builds_waiting_for_slaves queue to BuildSchedulerPool.

tjlee0909:
This is part 1 of a two-part fix to handle error'ed build state from https://github.com/box/ClusterRunner/issues/313 .

This commit should be a NO-OP and have CR behave identically.


boxcla:

PR #315, opened 2016-06-23
Allocating, de-allocating, and then re-allocating slaves should work.

tjlee0909:
This PR should fix https://github.com/box/ClusterRunner/issues/313

Currently, when slaves are added, removed, and added again to a build, the build fails to complete successfully.

### DIFF
--- a/app/master/build_request_handler.py
+++ b/app/master/build_request_handler.py
@@ -23,15 +23,14 @@ class BuildRequestHandler(object):
     entity) to pull Builds from.
 
     All of the input of builds come through self.handle_build_request() calls, and all of the output
-    of builds go through self.next_prepared_build_scheduler() calls.
+    of builds go through self._scheduler_pool.next_prepared_build_scheduler() calls.
     """
     def __init__(self, scheduler_pool):
         """
-        :type scheduler_pool: BuildSchedulerPool
+        :type scheduler_pool: app.master.build_scheduler_pool.BuildSchedulerPool
         """
         self._logger = get_logger(__name__)
         self._scheduler_pool = scheduler_pool
-        self._builds_waiting_for_slaves = Queue()
         self._request_queue = Queue()
         self._request_queue_worker_thread = SafeThread(
             target=self._build_preparation_loop, name='RequestHandlerLoop', daemon=True)
@@ -54,19 +53,6 @@ class BuildRequestHandler(object):
         self._request_queue.put(build)
         analytics.record_event(analytics.BUILD_REQUEST_QUEUED, build_id=build.build_id(),
                                log_msg='Queued request for build {build_id}.')
-
-    def next_prepared_build_scheduler(self):
-        """
-        Get the scheduler for the next build that has successfully completed build preparation.
-
-        This is a blocking call--if there are no more builds that have completed build preparation and this
-        method gets invoked, the execution will hang until the next build has completed build preparation.
-
-        :rtype: BuildScheduler
-        """
-        build = self._builds_waiting_for_slaves.get()
-        build_scheduler = self._scheduler_pool.get(build)
-        return build_scheduler
 
     def _build_preparation_loop(self):
         """
@@ -111,7 +97,7 @@ class BuildRequestHandler(object):
                     # If there is work to be done, this build must queue to be allocated slaves.
                     else:
                         self._logger.info('Build {} is waiting for slaves.', build.build_id())
-                        self._builds_waiting_for_slaves.put(build)
+                        self._scheduler_pool.add_build_waiting_for_slaves(build)
 
             except Exception as ex:  # pylint: disable=broad-except
                 build.mark_failed(str(ex))  # WIP(joey): Build should do this internally.
--- a/app/master/build_scheduler.py
+++ b/app/master/build_scheduler.py
@@ -20,14 +20,16 @@ class BuildScheduler(object):
     itself shouldn't know anything about its scheduler or slaves or where subjobs
     are executing. All that goes here.
 
-    Ususally this class is instantiated and managed by a BuildSchedulerPool.
+    This class is instantiated and managed by a BuildSchedulerPool.
     """
-    def __init__(self, build):
+    def __init__(self, build, scheduler_pool):
         """
-        :type build: Build
+        :type build: app.master.build.Build
+        :type scheduler_pool: app.master.build_scheduler_pool.BuildSchedulerPool
         """
         self._logger = get_logger(__name__)
         self._build = build
+        self._scheduler_pool = scheduler_pool
 
         job_config = build.project_type.job_config()
         self._max_executors = job_config.max_executors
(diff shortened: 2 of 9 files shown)


boxcla:

app/master/build_scheduler.py
```diff
@@ -101,18 +99,23 @@ def execute_next_subjob_or_free_executor(self, slave):
...
                     subjob.mark_in_progress(slave)
-
                 except SlaveMarkedForShutdownError:
                     self._build._unstarted_subjobs.put(subjob)  # todo: This changes subjob execution order. (Issue #226)
                     # An executor is currently allocated for this subjob in begin_subjob_executions_on_slave.
                     # Since the slave has been marked for shutdown, we need to free the executor.
                     self._free_slave_executor(slave)
+                else:
+                    if is_first_subjob:
+                        self._build.mark_started()
```
josephharrington: Does this alter the point at which `mark_started` gets called? Instead of before slave setup, now it looks like it's called after the first slave setup returns?

app/master/build_request_handler.py
```diff
@@ -55,19 +54,6 @@ def handle_build_request(self, build):
...
-
-        This is a blocking call--if there are no more builds that have completed build preparation and this
-        method gets invoked, the execution will hang until the next build has completed build preparation.
-
-        :rtype: BuildScheduler
-        """
-        build = self._builds_waiting_for_slaves.get()
-        build_scheduler = self._scheduler_pool.get(build)
-        return build_scheduler
-
```
josephharrington: Awesome. I like how this kind of cancels out.

app/master/build_scheduler.py
```diff
@@ -101,18 +99,23 @@ def execute_next_subjob_or_free_executor(self, slave):
             # this method, finds the subjob queue empty, and is torn down.  If that was the last 'living' slave, the
             # build would be stuck.
             with self._subjob_assignment_lock:
+                is_first_subjob = False
+                if self._build._unstarted_subjobs.qsize() == len(self._build.all_subjobs()):
+                    is_first_subjob = True
```
josephharrington:
```python
is_first_subjob = (self._build._unstarted_subjobs.qsize() == len(self._build.all_subjobs())
```
?

test/functional/job_configs.py
```diff
@@ -151,3 +151,34 @@ def __init__(self, config, expected_to_fail, expected_num_subjobs, expected_num_
         File('subjob_file_3.txt', contents='setup.\nsubjob 3.\nteardown.\n'),
     ],
 )
+
+# This is a very basic job where each atom just creates a simple text file.
```
josephharrington: Did you mean to change this comment?

test/functional/master/test_deallocation_and_allocation_of_slaves_mid_build.py
```diff
@@ -0,0 +1,30 @@
...
+            'type': 'directory',
+            'config': yaml.safe_load(JOB_WITH_SLEEPS.config[os.name])['BasicSleepingJob'],
+            'project_directory': project_dir.name,
+        })
+        build_id = build_resp['build_id']
+        master.block_until_build_started(build_id, timeout=10)
+        master.graceful_shutdown_slaves_by_id([1])
+        self.cluster.block_until_n_slaves_dead(num_slaves=1, timeout=10)
+        self.cluster.kill_slaves(kill_gracefully=False)
+        self.assert_build_status_contains_expected_data(build_id, {'status': 'BUILDING'})
```
josephharrington: :heart_eyes:

test/unit/master/test_build.py
```diff
@@ -359,23 +359,22 @@ def test_preparing_build_creates_empty_results_directory(self):
...
         building_timestamp1 = self._get_build_state_timestamp(build, BuildState.BUILDING)
-        scheduler.allocate_slave(slave=mock_slave2)
+        scheduler.execute_next_subjob_or_free_executor(mock_slave1)
         building_timestamp2 = self._get_build_state_timestamp(build, BuildState.BUILDING)
-
-        self.assertIsNotNone(building_timestamp1, '"building" timestamp should be set after first slave allocated.')
+        self.assertIsNotNone(building_timestamp1, '"building" timestamp should be set after first subjob is started.')
         self.assertEqual(building_timestamp1, building_timestamp2,
-                         '"building" timestamp should not change upon further slave allocation.')
+                         '"building" timestamp should not change upon further subjob execution.')
```
josephharrington: Did you mean to change the assert message on line 368 too? (Also see my question above about why this changed.)

app/master/build_scheduler.py
tjlee0909: Done

app/master/build_scheduler.py
tjlee0909: Yeah, I changed this behavior a little bit. If I didn't change this, and instead kept it as it was, then _build.mark_started() would get called multiple times, and lead to a FSM state transition error: 

"Error during build state transition. (FysomError: event BuildEvent.START_BUILDING inappropriate in current state BuildState.BUILDING)".

Alternatively, I could have added a build transition from the BUILDING state to itself upon multiple START_BUILDING events and left the self._build.mark_started() in the allocate_slave() method, but I thought it'd make more sense to not mark a build started multiple times. I'm open to keeping it the way it was.

test/functional/job_configs.py
tjlee0909: Thanks, done.

test/unit/master/test_build.py
tjlee0909: Yeah, I should have changed that if I were to keep the things the way they are in the PR.

tjlee0909: Done

app/master/build_scheduler.py
josephharrington: The only other thing I could think of was to have some flag set to false when the scheduler is instantiated, and then set it to true when a slave gets allocated. (Basically just using an explicit flag instead of the implicit emptiness of the `_slaves_allocated` list.)

josephharrington:
👍 Looks good. I feel like it would be ideal if the building timestamp was set as soon as a slave was allocated, but it's not too big of an issue. The main purpose of that is so that we could measure how long a build is prepared but just waiting on slaves -- with this change the difference between the `PREPARED` timestamp and the `BUILDING` timestamp would also include the setup time of the first allocated slave, which maybe is fine.

tjlee0909:
I see, I'll use the boolean flag for slave allocated.

app/master/build_scheduler.py
```diff
@@ -63,10 +66,9 @@ def allocate_slave(self, slave):
 
         :type slave: Slave
         """
-        if not self._slaves_allocated:
-            # If this is the first slave to be allocated, update the build state.
+        if not self._build_started:
+            self._build_started = True
```
tjlee0909: @josephharrington - Ok, this seems pretty simple.

josephharrington:
👍

PR #322, opened 2016-09-06
Fix broken functional tests on windows

josephharrington:
Something must have changed on the Appveyor side since this code
hasn't been touched recently. The issue was that a functional test
was referring to "/tmp" which is not valid on Windows. I'm surprised
this worked on Windows in the first place.


boxcla:

PR #325, opened 2016-12-30
Prevent immortal connection pools

josephharrington:
The master keeps a connection pool for each connected slave. The
problem is that even when the slave disconnects/is taken offline
those pooled connections stay open. This isn't that bad unless
you're constantly connecting and disconnecting slaves.

At Box we constantly connect and disconnect slaves (recycling the
entire cluster once a night). Over several weeks the immortal
connection pools eventually cause the master process to hit a system
limit on the number of open file descriptors. :(

This change makes sure we clear the connection pool when a slave
disconnects.

### DIFF
--- a/test/unit/master/test_slave.py
+++ b/test/unit/master/test_slave.py
@@ -1,13 +1,12 @@
-from unittest.mock import Mock, MagicMock
+from unittest.mock import Mock, MagicMock, ANY
 
-from app.master.atom import Atom, AtomState
 from app.master.build import Build
 from app.master.build_request import BuildRequest
 from app.master.slave import DeadSlaveError, SlaveMarkedForShutdownError, Slave
-from app.master.subjob import Subjob
 from app.util.secret import Secret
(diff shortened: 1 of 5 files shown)


boxcla:

test/unit/master/test_slave.py
```diff
@@ -140,10 +139,10 @@ def test_mark_as_idle_raises_when_executors_are_in_use(self):
     def test_mark_as_idle_raises_when_slave_is_in_shutdown_mode(self):
         slave = self._create_slave()
         slave._is_in_shutdown_mode = True
-        slave.kill = Mock()
 
         self.assertRaises(SlaveMarkedForShutdownError, slave.mark_as_idle)
-        slave.kill.assert_called_once_with()
+        self.mock_network.post_with_digest.assert_called_once_with(
+            AnyStringMatching('/v1/kill'), ANY, ANY)
```
josephharrington: Changes in existing tests are just random cleanup not related to my change.

PR #327, opened 2017-02-14
Add Prometheus instrumentation

gcurtis:
This change uses the Prometheus client SDK to add some instrumentation to the master. So far, only HTTP request latency is instrumented.

There is also a new endpoint at /v1/metrics which exposes the metrics in in Prometheus text exposition format. The metrics endpoint required a new method in ClusterBaseHandler so that a raw text string could be written.

### DIFF
--- a/app/web_framework/cluster_master_application.py
+++ b/app/web_framework/cluster_master_application.py
@@ -3,6 +3,8 @@ import os
 import urllib.parse
 
 import tornado.web
+import prometheus_client
+from prometheus_client import Histogram
 
 from app.util import analytics
 from app.util import log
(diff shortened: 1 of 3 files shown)


boxcla:
Verified that @gcurtis has signed the CLA. Thanks for the pull request!

app/web_framework/cluster_master_application.py
```diff
@@ -4,6 +4,9 @@
 
 import tornado.web
```
josephharrington: Plz remove blank line. Thx.

requirements.txt
```diff
@@ -16,3 +16,4 @@ PyYAML==3.11
 requests==2.3.0
 termcolor==1.1.0
 tornado==3.2.2
+prometheus_client==0.0.19
```
josephharrington: Plz alphabetize. Thx.

Issue #329, opened 2017-03-02
Console output logs are not available on the master


josephharrington:
The API to get console output for a build's atom currently goes through the flow:
1. Try to get the console output from a file on the master
2. If that file is not there, redirect to the slave's console output endpoint for the build.

The intention is that for finished builds we'll load console output from the master, and for in-progress builds we'll redirect to the relevant slave that the atom is running on.

We broke this in #298 when we started deleting build files after a build completed (to prevent inode exhaustion). Currently the master *always* redirects to the slave to get console output for finished builds since the console output file has been deleted on the master.

For Box specifically, this has become a more critical issue lately since we've started recycling slave nodes on a nightly schedule. This means that console output is not available for builds older than a day or so since the master tries to redirect to a slave that no longer exists.

josephharrington:
The suggested fix is to pull console output from the build artifact file on the master (instead of the current behavior of looking for the appropriate clusterrunner_console_output text file).

A catch here is that right now we're creating the build artifact as a tar file. Tar files have no index, so it's impossible to pull a single file out of them without reading the whole thing. This is not a problem with small build artifact files, but depending on what the job does and how many atoms it has this file can be arbitrarily large.

One workaround is to compress the artifact to a zip file instead of a tar. Zip files allow efficiently reading individual files from inside the artifact without extracting the whole thing. We'd probably want to create both the tar and the zip in the short term so that we don't break anything expecting a tar.

tjlee0909:
I recognize that this is an immediately pressing issue due to how much we want to get that automated recycling happen at Box, so my suggestion below is not something that we can immediately act on.

I was thinking that we should start breaking up the functionality of the cluster-master, and coming up with a new entity that is responsible for making sure that all of the services are running and orchestrating them.

Right now, the Cluster-Master does everything. I'd like to break it such that our ClusterRunner service topology looks something like this:

1). The front end service. This is the entry point to ClusterRunner, and the only way API users interact with ClusterRunner. This service should maintain the same V1 API interface that we currently provide (I don't see a reason why our API needs to change yet). This service doubly acts as a load balancer for the other services listed below (for HA).

2). A build result storage service. Responsible for persisting, and retrieving builds. Any service that needs to reference build results (e.g.: console output of completed builds) would go here. This service should be able to support keying by not just build id's, but also by session id's so that we can persist build histories across as many ClusterRunner restarts that we want. Additionally, any logic for how long we want to keep builds, and the cleanup of old ones, would live here.

[...]

PR #332, opened 2017-04-07
Rename archive methods to reference tar

josephharrington:
The current artifact archiving methods and variables only refer to
"artifact" and "compress". To address #329 we'd like to create a zip
archive file as well, so this change renames the current methods to
remove ambiguitiy between tar and (future) zip methods.

I also refactored `tar_directory` slightly so that it uses
`tar_directories`.


boxcla:

PR #335, opened 2017-04-26
Deprecate build result tar.gz in favor of zip

### LINKED ISSUE #329: Console output logs are not available on the master

josephharrington:
This adds an artifacts.zip with exactly the same contents as the existing artifacts.tar.gz. The tar.gz artifact format is now deprecated and clients should update to use the zip file instead.

The benefit of using a zip is that it will be easier to fix #329, an issue around loading console output from completed builds. By having a zip archive of the build results on the master, we'll be able to load console output directly from the master without redirecting to the slave that the build ran on. The fix for this issue is coming in a follow-up change.


boxcla:

app/master/build.py
```diff
@@ -81,7 +83,6 @@ def api_representation(self):
         return {
             'id': self._build_id,
             'status': build_state,
-            'artifacts': self._artifacts_archive_file,  # todo: this should probably be a url, not a file path
```
josephharrington: I removed this because it was an API field whose value was a local path. I think originally we might have planned to make it a url, but I'd rather do something like that as part of a larger effort to make the API more discoverable instead of only this one-off.

app/client/build_runner.py
```diff
@@ -164,7 +164,7 @@ def _download_and_extract_results(self, timeout=None):
                     for chunk in response.iter_content(chunk_size):
                         file.write(chunk)
 
-                app.util.fs.extract_tar(download_filepath, delete=True)
+                app.util.fs.unzip_directory(download_filepath, delete=True)
```
josephharrington: This is part of the CLI client. The type of archive it uses internally to download results doesn't *really* matter, but it's better to change it now since the plan is to eventually remove tar.gz result support.

app/master/build.py
```diff
@@ -452,8 +464,13 @@ def _create_build_artifact(self):
         self._build_artifact = BuildArtifact(self._build_results_dir())
         self._build_artifact.generate_failures_file()
         self._build_artifact.write_timing_data(self._timing_file_path, self._read_subjob_timings_from_results())
-        self._artifacts_archive_file = app.util.fs.tar_directory(self._build_results_dir(),
-                                                                 BuildArtifact.ARTIFACT_TARFILE_NAME)
+        self._artifacts_zip_file = app.util.fs.zip_directory(self._build_results_dir(),
+                                                             BuildArtifact.ARTIFACT_ZIPFILE_NAME)
+        # Temporarily move aside zip file so we can create a tar file, then move it back.
+        temp_zip_path = shutil.move(self._artifacts_zip_file, tempfile.mktemp())
```
josephharrington: This is a bit gross but I'm okay with it since we'll ideally remove the tar file at some point in the future.

test/framework/functional/base_functional_test_case.py
```diff
@@ -31,28 +31,6 @@ def setUp(self):
         self.cluster = FunctionalTestCluster(verbose=self._get_test_verbosity())
         self._network = Network()
 
-    def _create_test_config_file(self, conf_values_to_set=None):
```
josephharrington: Dead code

app/client/build_runner.py
```diff
@@ -146,8 +146,8 @@ def _download_and_extract_results(self, timeout=None):
         """
         timeout_time = time.time() + timeout if timeout else sys.maxsize
 
-        download_artifacts_url = self._master_api.url('build', self._build_id, 'result')
-        download_filepath = 'build_results/artifacts.tar.gz'
+        download_artifacts_url = self._master_api.url('build', self._build_id, 'artifacts.zip')
```
wjdhollow: Minor. Should we use a constant or variable for 'artifacts.zip'?

app/master/cluster_master.py
```diff
@@ -296,20 +296,19 @@ def get_build(self, build_id):
 
         return build
 
-    def get_path_for_build_results_archive(self, build_id):
+    def get_path_for_build_results_archive(self, build_id: int, is_tar_request: bool=False) -> str:
         """
         Given a build id, get the absolute file path for the archive file containing the build results.
 
+        :param is_tar_request: If true, download the tar.gz archive instead of a zip.
```
wjdhollow: Minor, but your param docs are out of order.

app/client/build_runner.py
josephharrington: Hmm maybe, although it's just an API route so by that argument 'build' should probably also be a constant. I could see arguments for making those constants but kind of want to keep that out of this change.

app/master/cluster_master.py
josephharrington: Thanks! I'll fix that.

app/client/build_runner.py
cmcginty: There are a number of places where `artifacts.zip` is referenced in the code. In addition, there is a new static define BuildArtifact.ARTIFACT_ZIPFILE_NAME defined as `results.zip`. The usage of the term "artifact" is getting pretty muddy.

PR #338, opened 2017-05-10
Add timeout to all outgoing http requests

josephharrington:
Currently if a slave becomes unresponsive the master service will
freeze up for a long time (or forever). The slave allocation loop
will hang trying to contact a slave and no builds will be able to
start.

This adds a timeout to all requests between the master and slaves
and corresponding error handling. I made the timeout on
slave->master calls longer because we don't currently have great
metrics and have had issues with request load on the master in the
past (which we should improve separately).

### DIFF
--- a/app/master/build_scheduler.py
+++ b/app/master/build_scheduler.py
@@ -1,7 +1,7 @@
 from queue import Empty
 from threading import Lock
 
-from app.master.slave import SlaveMarkedForShutdownError
+from app.master.slave import Slave, SlaveMarkedForShutdownError
 from app.util import analytics
 from app.util.log import get_logger
 
@@ -60,19 +60,24 @@ class BuildScheduler(object):
             return False
         return True
 
-    def allocate_slave(self, slave):
+    def allocate_slave(self, slave: Slave) -> bool:
         """
         Allocate a slave to this build. This tells the slave to execute setup commands for this build.
-
-        :type slave: Slave
+        :param slave: The slave to allocate
+        :return: Whether slave allocation was successful; this can fail if the slave is unresponsive
         """
         if not self._build_started:
             self._build_started = True
             self._build.mark_started()
-        self._slaves_allocated.append(slave)
-        slave.setup(self._build, executor_start_index=self._num_executors_allocated)
+
+        # Increment executors before triggering setup. This helps make sure the build won't take down
+        # every slave in the cluster if setup calls fail because of a problem with the build.
+        next_executor_index = self._num_executors_allocated
         self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
         analytics.record_event(analytics.BUILD_SETUP_START, build_id=self._build.build_id(), slave_id=slave.id)
+        self._slaves_allocated.append(slave)
+
+        return slave.setup(self._build, executor_start_index=next_executor_index)
 
     def begin_subjob_executions_on_slave(self, slave):
         """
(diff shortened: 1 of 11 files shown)


boxcla:

app/util/network.py
```diff
@@ -100,7 +101,17 @@ def encode_body(self, body_decoded):
         """
         return json.dumps(body_decoded)
 
-    def _request(self, method, url, data=None, should_encode_body=True, error_on_failure=False, *args, **kwargs):
```
wjdhollow: Super minor, but you are missing the python3 annotations.

josephharrington: @wjdhollow Ah yeah, I decided not to do that here because of the union types (e.g., `dict|str|bytes|None`) that would force me to mix between docstring and annotations. I want to add mypy/typing to this repo at some point so that we can do more complex type annotations.

app/master/build_scheduler.py
```diff
@@ -60,19 +60,24 @@ def needs_more_slaves(self):
...
-        slave.setup(self._build, executor_start_index=self._num_executors_allocated)
+
+        # Increment executors before triggering setup. This helps make sure the build won't take down
+        # every slave in the cluster if setup calls fail because of a problem with the build.
+        next_executor_index = self._num_executors_allocated
         self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
         analytics.record_event(analytics.BUILD_SETUP_START, build_id=self._build.build_id(), slave_id=slave.id)
+        self._slaves_allocated.append(slave)
+
+        return slave.setup(self._build, executor_start_index=next_executor_index)
```
cmcginty: If the the slave setup fails, shouldn't the `num_executors_allocated` and `slaves_allocated` list be corrected?

app/master/build_scheduler.py
```diff
@@ -60,19 +60,24 @@ def needs_more_slaves(self):
...
         if not self._build_started:
             self._build_started = True
             self._build.mark_started()
-        self._slaves_allocated.append(slave)
-        slave.setup(self._build, executor_start_index=self._num_executors_allocated)
+
+        # Increment executors before triggering setup. This helps make sure the build won't take down
+        # every slave in the cluster if setup calls fail because of a problem with the build.
+        next_executor_index = self._num_executors_allocated
         self._num_executors_allocated += min(slave.num_executors, self._max_executors_per_slave)
```
cmcginty: This line looks suspicious, mostly because I would never expect `slave.num_executors > self._max_executors_per_slave` to be true.

josephharrington: @cmcginty `max_executors_per_slave` is a user-configured parameter for each job (set in their clusterrunner.yaml), and often it is set to 1 to limit parallelism on a given host.

app/master/build_scheduler.py
josephharrington: Ideally yes (actually ideally we'd just attempt the setup call before adjusting those values, then return if it fails.) It's a bit weird right now but as I mentioned in the comment, if for some reason the call to start setup fails because of something weird with the build then we want to make sure that won't take down every slave in the cluster. `self._num_executors_allocated` is the only thing preventing that currently.

This should definitely be improved, although it's a pretty rare occurrence that I don't believe we've seen in the wild. It's also gross because if the call to start setup does fail on every slave assigned to the build then the build will just hang forever in the building state. I'd like to come back to this and improve it but I'm not too concerned about it being a common issue.

app/master/build_scheduler.py
josephharrington: Another aspect that might not be obvious is that there's a one-to-one correspondence between Build instances and BuildScheduler instances, so any instance attribute on this class essentially belongs to a specific build. I'm happy to explain this choice in more detail sometime.

PR #340, opened 2017-05-15
Add Prometheus metrics for slave request latency

josephharrington:
This mirrors the current metrics on the master that were added in #327.


boxcla:

tjlee0909:
Probably worth looking into seeing if we can move this into a base class somewhere.

👍

PR #341, opened 2017-05-15
Do not crash master when postbuild tasks fail

### LINKED ISSUE #323: Postbuild failures can take down master
Right now we run postbuild tasks (creating build artifact.tar.gz, deleting intermediate result files) in a `SafeThread`. This means if an exception is raised we'll take down the master. This is totally unnecessary -- we can just run postbuild tasks in a normal `Thread` and fail the build if an error occurs.

josephharrington:
Previously if an exception occurred during postbuild tasks (archiving artifacts, writing timing data, etc.) for any build then the master service would shut down. Nonsense!

This changes the behavior to merely fail the build whose postbuild tasks failed.

Fixes #323.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -4,7 +4,7 @@ import os
 from queue import Queue, Empty
 import shutil
 import tempfile
-from threading import Lock
+from threading import Lock, Thread
 import time
 import uuid
 
@@ -19,7 +19,6 @@ from app.util.exceptions import ItemNotFoundError
 import app.util.fs
 from app.util.log import get_logger
 from app.util.single_use_coin import SingleUseCoin
-from app.util.safe_thread import SafeThread
 
 
 class Build(object):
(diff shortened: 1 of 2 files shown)


boxcla:

test/unit/master/test_build.py
```diff
@@ -445,6 +445,13 @@ def test_delete_temporary_build_artifact_files_skips_results_tarball(self):
 
         mock_shutil.rmtree.assert_called_once_with(expected_async_delete_call_path, ignore_errors=True)
 
+    def test_exception_during_postbuild_tasks_fails_build(self):
+        self.mock_util.fs.zip_directory.side_effect = FileNotFoundError('Where my files at?')
+
+        build = self._create_test_build(BuildStatus.FINISHED)
+
+        self.assertEqual(build._status(), BuildStatus.ERROR)
```
josephharrington: You will say, "Whoa, why are you accessing a private method?!" And I will say, "Yes, I do not love that but all the tests in here do it and fixing it is outside the scope of this change."

app/master/build.py
```diff
@@ -454,12 +454,18 @@ def _result(self):
...
-        self._delete_temporary_build_artifact_files()
-        self._postbuild_tasks_are_finished = True
-        self._state_machine.trigger(BuildEvent.POSTBUILD_TASKS_COMPLETE)
+        try:
+            self._create_build_artifact()
+            self._delete_temporary_build_artifact_files()
+            self._postbuild_tasks_are_finished = True
+            self._state_machine.trigger(BuildEvent.POSTBUILD_TASKS_COMPLETE)
+
+        except Exception as ex:  # pylint: disable=broad-except
```
ethomas2: This is happening on the master right? Can we get some prometheus metrics to tell us when this occurs? Even if it's just a `num_errors` counter like the one we have in forge?

josephharrington: Yes, I want to add more Prometheus metrics in general but will do that in a separate change.

Issue #339, opened 2017-05-15
FileNotFoundError when creating zip archive


josephharrington:
This is a new bug (introduced with the recent zip archive change) ~that triggers #323, bringing down the master service~. (Edit: #343 was merged to prevent this from taking down the master.)
```
[2017-05-15 02:57:53.006] 9852 INFO    MasterTornadoTh build           All results received for build 4!
[2017-05-15 02:57:54.673] 9852 DEBUG   PostBuild4      build_artifact  Overwrote existing timing file in /home/jenkins/.clusterrunner/timings/master/gitosis@devscmro.dev.box.net/appconf/PHPUnit.timing.json
[2017-05-15 02:58:04.139] 9852 ERROR   PostBuild4      unhandled_excep Unhandled exception handler caught exception.
Traceback (most recent call last):
  File "/home/jenkins/ClusterRunnerBuild/app/util/safe_thread.py", line 18, in run
    super().run()
  File "/usr/local/lib/python3.4/threading.py", line 868, in run
    self._target(*self._args, **self._kwargs)
  File "/home/jenkins/ClusterRunnerBuild/app/master/build.py", line 459, in _perform_async_postbuild_tasks
    self._create_build_artifact()
  File "/home/jenkins/ClusterRunnerBuild/app/master/build.py", line 469, in _create_build_artifact
    BuildArtifact.ARTIFACT_ZIPFILE_NAME)
  File "/home/jenkins/ClusterRunnerBuild/app/util/fs.py", line 134, in zip_directory
    tmp_path = shutil.make_archive(tempfile.mktemp(), 'zip', target_dir)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/shutil.py", line 785, in make_archive
    filename = func(base_name, base_dir, **kwargs)
  File "/home/jenkins/.virtualenvs/clusterrunner/lib/python3.4/shutil.py", line 690, in _make_zipfile
    zf.write(path, path)
  File "/usr/local/lib/python3.4/zipfile.py", line 1366, in write
    with open(filename, "rb") as fp:
FileNotFoundError: [Errno 2] No such file or directory: 'artifact_531_11/result.xml'
```

josephharrington:
This is pretty weird because this is coming from inside the stdlib, and the line write before `zf.write()` in shutil checks for file existence:
```python
for dirpath, dirnames, filenames in os.walk(base_dir):
    for name in filenames:
[...]

PR #343, opened 2017-05-16
Ignore zipfile artifact creation failures

josephharrington:
We've seen an issue (#339) on our ClusterRunner production instance that will cause zip creation to fail with a FileNotFoundError. Since we're not yet depending on zip files, this just logs and ignores any failures that occur.

This also moves zip creation to occur after tar.gz creation. Hopefully we'll be able to use the successfully created tar.gz artifact to debug issues with creating the zip artifact.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -471,13 +471,21 @@ class Build(object):
         self._build_artifact = BuildArtifact(self._build_results_dir())
         self._build_artifact.generate_failures_file()
         self._build_artifact.write_timing_data(self._timing_file_path, self._read_subjob_timings_from_results())
-        self._artifacts_zip_file = app.util.fs.zip_directory(self._build_results_dir(),
-                                                             BuildArtifact.ARTIFACT_ZIPFILE_NAME)
-        # Temporarily move aside zip file so we can create a tar file, then move it back.
-        temp_zip_path = shutil.move(self._artifacts_zip_file, tempfile.mktemp())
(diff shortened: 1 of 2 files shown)


boxcla:

josephharrington:
The happy path of this change should be covered by current functional/unit tests.

app/master/build.py
```diff
@@ -465,13 +465,21 @@ def _create_build_artifact(self):
...
+        temp_tar_path = None
+        try:
+            # Temporarily move aside tar file so we can create a zip file, then move it back.
+            # This juggling can be removed once we're no longer creating tar artifacts.
+            temp_tar_path = shutil.move(self._artifacts_tar_file, tempfile.mktemp())
+            self._artifacts_zip_file = app.util.fs.zip_directory(self._build_results_dir(),
+                                                                 BuildArtifact.ARTIFACT_ZIPFILE_NAME)
+        except Exception:  # pylint: disable=broad-except
+            # Due to issue #339 we are ignoring exceptions in the zip file creation for now.
+            self._logger.exception('Zipping of artifacts failed. This error will be ignored.')
```
gcurtis: Would it be straightforward to add a Prometheus metric as well?

josephharrington: You [sound like](https://github.com/box/ClusterRunner/pull/341#discussion_r116619079) @ethomas2. ;)

Yes, I will do that in a follow-up change.

Issue #345, opened 2017-05-19
Functional test broken on AppVeyor


josephharrington:
This is probably just a race condition in the test -- I think tests run pretty slowly on AppVeyor and so might be more likely to fail when tests rely on timing. A quick glance at the test code makes me think it is susceptible to races in at least two different places.
```
======================================================================
FAIL: test_build_completes_after_allocating_deallocating_and_reallocating_slaves_to_build (test.functional.master.test_deallocation_and_allocation_of_slaves_mid_build.TestDeallocationAndAllocationOfSlavesMidBuild)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\projects\clusterrunner\test\functional\master\test_deallocation_and_allocation_of_slaves_mid_build.py", line 30, in test_build_completes_after_allocating_deallocating_and_reallocating_slaves_to_build
    self.assert_build_has_successful_status(build_id)
  File "C:\projects\clusterrunner\test\framework\functional\base_functional_test_case.py", line 106, in assert_build_has_successful_status
    self.assert_build_status_contains_expected_data(build_id, expected_successful_build_params)
  File "C:\projects\clusterrunner\test\framework\functional\base_functional_test_case.py", line 92, in assert_build_status_contains_expected_data
    'Build status API response should contain the expected status data.')
nose.proxy.AssertionError: Mismatched values: 'result', expected: 'NO_FAILURES', actual: None,'status', expected: 'FINISHED', actual: 'BUILDING' : Build status API response should contain the expected status data.
-------------------- >> begin captured stdout << ---------------------
[2017-05-19 07:12:41.699] 2076 NOTICE  MainThread      functional_test Hard killing process with pid 2844...
[2017-05-19 07:13:02.944] 2076 NOTICE  MainThread      functional_test Gracefully killing process with pid 2152...
[2017-05-19 07:13:02.944] 2076 NOTICE  MainThread      functional_test Gracefully killing process with pid 2908...
--------------------- >> end captured stdout << ----------------------
```

PR #350, opened 2017-05-25
Skip test that fails on Windows

josephharrington:
See issue #345.

### DIFF
--- a/test/functional/master/test_deallocation_and_allocation_of_slaves_mid_build.py
+++ b/test/functional/master/test_deallocation_and_allocation_of_slaves_mid_build.py
@@ -1,11 +1,14 @@
 import os
 import tempfile
+from unittest import skipIf
 import yaml
 
+from app.util.process_utils import is_windows
 from test.framework.functional.base_functional_test_case import BaseFunctionalTestCase
 from test.functional.job_configs import JOB_WITH_SLEEPS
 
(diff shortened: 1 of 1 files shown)


boxcla:

test/functional/master/test_deallocation_and_allocation_of_slaves_mid_build.py
```diff
@@ -1,11 +1,14 @@
...
 import tempfile
+from unittest import skipIf
 import yaml
 
+from app.util.process_utils import is_windows
 from test.framework.functional.base_functional_test_case import BaseFunctionalTestCase
 from test.functional.job_configs import JOB_WITH_SLEEPS
 
 
+@skipIf(is_windows(), 'Fails on AppVeyor; see issue #345')
```
tjlee0909: Can we just increase the timeout for master.block_until_build_finished() to 20 seconds or something instead of skipping this on Windows?

josephharrington: Skipping is intended as a short term fix until someone can actually fix the test, perhaps by doing what you suggest.

tjlee0909: K

Issue #348, opened 2017-05-24
ClusterRunner should use hardcoded version string


shriganeshs:
The version should be hardcoded string. If ClusterRunner could not find version in  cx_Freeze package, it assumes you're developing it locally and tries to calculate the version from git commits.
```bash
# clusterrunner -V
fatal: Not a git repository (or any of the parent directories): .git
ClusterRunner 0.5.???
```

Below is the code
```python
def get_version():
    """
    Get the version of the application. This method should return the correct version in both the frozen and unfrozen
    (running from cloned source) cases.
    :return: The version of the application
    :rtype: str
    """
    if getattr(sys, 'frozen', False):
        return _get_frozen_package_version()  # frozen/packaged

    return _calculate_source_version()  # unfrozen/running from source```

**Comment by josephharrington** [comment]
There are two tasks:
1. Execute logic that hardcodes the version before building the rpm.
2. Change the logic in `get_version` so that it doesn't rely on the `sys.frozen` attribute.

See setup.py for the logic that we'd need to execute before building the rpm:
```
version = autoversioning.get_version()
autoversioning.write_package_version_file(version)
```
This will create a file that essentially hardcodes the version number.

Just for background, the `getattr(sys, 'frozen', False)` pattern is not specific to cx_Freeze -- that is also used in other bundling/freezing tools.

But it sounds like bundling into an rpm does not set the `frozen` attribute, so we should decide on another way to detect it. Maybe it can switch on if it's running from a git repo or not. (If running in git repo then calculate the version, otherwise use hardcoded version.)

josephharrington:
In case it was not clear from my above comment, my vote is to keep auto-versioning (not switch to manually-updated, hardcoded version numbers).

shriganeshs:
Fixed in #352

PR #351, opened 2017-05-26
Create cr rpm

shriganeshs:
This PR adds

1. Makefile target 'release'
2. .flute file to build an rpm
3. ability to auto generate version


boxcla:
Hi @shriganeshs, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

shriganeshs:
CLA signed

boxcla:
Verified that @shriganeshs has just signed the CLA. Thanks, and we look forward to your contribution.

Makefile
```diff
@@ -43,3 +49,21 @@ test-functional:
...
+	rm $(NAME).spec
+
+rpm: rpmbuild/RPMS/x86_64/$(NAME)-$(RPM_VERSION)-$(RELEASE).x86_64.rpm
+
+release: rpm
+	REPO_NAME="productivity" \
+	REPO_PATH="com/box" \
+	SERVICE_KIND="$(NAME)" \
+	SERVICE_VERSION="$(RPM_VERSION)-$(RELEASE)" \
+	rpm-to-artifactory rpmbuild/RPMS/x86_64/$(NAME)-$(RPM_VERSION)-$(RELEASE).x86_64.rpm
```
josephharrington: I don't think this stuff should go in the main open source repo. We should probably have a private supporting repo with the Box-specific scripts. That probably applies to the .flute file as well since Flute is not (yet) open source.

Makefile
```diff
@@ -1,5 +1,11 @@
 .PHONY: all lint test init pylint pep8 test-unit test-unit-via-clusterrunner test-functional freeze
 
+NAME=$(shell grep '^Name' .flute | grep -Eo '[[:alnum:]_-]+$$')
+RPM_VERSION_PREFIX=0.5
+COMMIT_COUNT=$(shell git rev-list --count HEAD)
```
josephharrington: I'd rather we not reimplement the autoversioning logic in the Makefile. It should just use the ClusterRunner code. Can it not just execute `./main.py --version` to get the version?

josephharrington:
Also, @shriganeshs I don't think this would solve the problem of `clusterrunner --version` not working after building the rpm, would it? For that to work `package_version.py` must be written.

Feel free to grab some time with me to chat about the right approach to this.

Makefile
tjlee0909: Agreed

shriganeshs: Yes, it will not. We still need to address #348.

shriganeshs:
Yes, this PR does not address #348.

Makefile
shriganeshs: To run ClusterRunner (i.e. main.py) don't we need to run `make init` first to install requirements?
Why the builder machine should have capability to run the CR executable? It should just package the CR.

josephharrington: Fair point. We could work around that however by doing something like converting autoversioning.py to a requirement-less script. So running ./app/util/autoversioning.py would just spit out the version. I think that's better than having two different ways to calculate the version.

shriganeshs:
This PR is split into #352 and a new private repo (for rpm building).

PR #352, opened 2017-05-31
Enable autoversioning to output correct version

shriganeshs:
This is an enhancement to enable rpm building with auto-generated version. Addresses #348.
1. Now `autoversioning.py` can be used by Makefile to generate a version string for building ClusterRunner rpm.
2. ClusterRunner executable
    1. First looks for version in hardcoded (generated while rpm building) `package_version.py` file.
    2. Else it calculates the version using git commit count (assuming its running within git repo).
    3. If 1 and 2 fails ClusterRunner will print '0.0.0' as default version.

### DIFF
--- a/app/util/autoversioning.py
+++ b/app/util/autoversioning.py
@@ -20,18 +20,24 @@ def get_version():
     :return: The version of the application
     :rtype: str
     """
-    if getattr(sys, 'frozen', False):
-        return _get_frozen_package_version()  # frozen/packaged
 
-    return _calculate_source_version()  # unfrozen/running from source
+    return _get_frozen_package_version() or _calculate_source_version() or '0.0.0'
 
 
 def _try_rename(src, dst):
     try:
         os.rename(src, dst)
-    except FileExistsError:
-        # Skip backing up the original package_version.py if a FileExistsError happened.
-        # This might happen on Windows as NTFS doesn't support writing to a file while the file is opened in python.
+    except (FileExistsError, FileNotFoundError):
+        # Skip backing up the original package_version.py if a FileExistsError or FileNotFoundError happened.
+        # FileExistsError might happen on Windows as NTFS doesn't support writing to a file while the file
+        # is opened in python.
+        pass
+
+
+def _try_remove(src):
+    try:
+        os.remove(src)
+    except OSError:
         pass
 
 
@@ -70,8 +76,11 @@ def _get_frozen_package_version():
 
     # only import package_version when needed as on Windows once imported, the actual package_version.py can't be
     # edited anymore
-    from app.util import package_version
-    return package_version.version
+    try:
+        from app.util import package_version
+        return package_version.version
+    except ImportError:
+        return None
 
 
 def _calculate_source_version():
@@ -79,15 +88,12 @@ def _calculate_source_version():
     Calculate the version using a scheme based off of git repo info. Note that since this depends on the git history,
     this will *not* work from a frozen package (which does not include the git repo data). This will only work in the
     context of running the application from the cloned git repo.
-
-    If one of the git commands used to calculate the version fails unexpectedly, the patch in the version string will
-    be set to "???".
+    If this is running outside of a git repo, it will handle the CalledProcessError exception and return None.
 
     :return: The version of the (source) application
     :rtype: str
     """
     global _calculated_version
-
     if _calculated_version is None:
         try:
             head_commit_hash = _get_commit_hash_from_revision_param('HEAD')
@@ -97,9 +103,8 @@ def _calculate_source_version():
             hash_extension = '' if head_commit_is_on_trunk else '-{}'.format(head_commit_hash[:7])
             mod_extension = '' if not _repo_has_uncommited_changes() else '-mod'
             _calculated_version = '{}.{}{}{}'.format(_MAJOR_MINOR_VERSION, commit_count, hash_extension, mod_extension)
-
         except subprocess.CalledProcessError:
-            _calculated_version = '{}.???'.format(_MAJOR_MINOR_VERSION)
+            _calculated_version = None
 
     return _calculated_version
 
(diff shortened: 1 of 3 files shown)


boxcla:

app/util/autoversioning.py
```diff
@@ -80,26 +92,19 @@ def _calculate_source_version():
     this will *not* work from a frozen package (which does not include the git repo data). This will only work in the
     context of running the application from the cloned git repo.
 
-    If one of the git commands used to calculate the version fails unexpectedly, the patch in the version string will
-    be set to "???".
```
josephharrington: Can you add a comment that if this is run outside of a git repo that it will raise `subprocess.CalledProcessError`?

Alternatively you could keep the try/except and just make this method return `None` if it fails. That would reduce coupling between the calling code and the implementation of this method.

app/util/autoversioning.py
```diff
@@ -20,18 +20,30 @@ def get_version():
     :return: The version of the application
     :rtype: str
     """
-    if getattr(sys, 'frozen', False):
-        return _get_frozen_package_version()  # frozen/packaged
 
-    return _calculate_source_version()  # unfrozen/running from source
+    # Initialize default version to '0.0.0'. So in case of failure the version will be shown as '0.0.0'.
+    version = '0.0.0'
```
josephharrington: I think '0.0.0' should be a constant if we're using it as a comparison. Alternatively you can initialize version to `None` (both here and in package_version.py) and then do a `return version or '0.0.0'` at the end of this method.

app/util/autoversioning.py
```diff
@@ -170,3 +175,12 @@ def _execute_local_git_command(*args):
...
     return command_output.decode()
+
+
+if __name__ == '__main__':
+    """
+    Print version string.
+    """
+    version = get_version()
+    write_package_version_file(version)
+    print(version)
```
josephharrington: You might want to call `restore_original_package_version_file` after printing the version so it doesn't leave the git repo dirty. It doesn't matter too much but is a nice-to-have. (See setup.py for how it's being called there.)

app/util/autoversioning.py
josephharrington: I've also been wondering if we should remove app/util/package_version.py from the repo completely. It seems like having a file already exist there with an invalid version adds more complexity than if we just wrote that file as needed. That would also simplify the code that backs up and restores the original package_version file.

Basically the code in that case would look like:
```python
def _get_frozen_package_version():
    try:
        from app.util import package_version
        return package_version.version
    except ImportError:
        return None


def _calculate_source_version():
    try:
        ...  # existing logic
    except subprocess.CalledProcessError:
        _calculated_version = None
    return _calculated_version


def get_version():
    return _get_frozen_package_version() or _calculate_source_version() or '0.0.0'
```

Thoughts?

app/util/autoversioning.py
shriganeshs: The `rpmbuild` needs to include the file in package. So I think we can not remove the file in the `autoprovisioning` script. I think we need to do it from Makefile.
Let me know if I am missing something.

app/util/autoversioning.py
shriganeshs: It is very clean approached. I liked it and I will make changes accordingly.

shriganeshs: Done.

app/util/autoversioning.py
shriganeshs: Done.

app/util/autoversioning.py
```diff
@@ -170,3 +175,12 @@ def _execute_local_git_command(*args):
         cwd=os.path.dirname(__file__),
     )
     return command_output.decode()
+
+
+if __name__ == '__main__':
+    """
+    Print version string.
```
shriganeshs: Done.

app/util/autoversioning.py
josephharrington: Ah, yeah that makes sense.

test/unit/util/test_autoversioning.py
```diff
@@ -86,13 +86,18 @@ def test_calculate_source_version_caches_computed_version(self):
...
 
-    def test_unexpected_failure_in_git_command_sets_patch_version_to_unknown(self):
+    def test_package_version_file_absent_and_unexpected_failure_in_git_command_sets_patch_version_to_default(self):
         self.check_output_mock.side_effect = [subprocess.CalledProcessError(1, 'fake')]  # make all git commands fail
 
+        # Return False for a 'isfile' check.
+        def fake_isfile(filename):
+            return False
+
+        self.patch('app.util.autoversioning.os').path.isfile.side_effect = fake_isfile
```
josephharrington: This seems unneeded now?

app/util/autoversioning.py
```diff
@@ -20,18 +20,24 @@ def get_version():
...
     :rtype: str
     """
-    if getattr(sys, 'frozen', False):
-        return _get_frozen_package_version()  # frozen/packaged
 
-    return _calculate_source_version()  # unfrozen/running from source
+    return _get_frozen_package_version() or _calculate_source_version() or '0.0.0'
 
 
 def _try_rename(src, dst):
```
josephharrington: We don't really even need this method anymore, right? `write_package_version_file` could just remove any existing file now.

We can do this cleanup in a separate change if you'd rather.

shriganeshs: Yes. Thanks for pointing out.

test/unit/util/test_autoversioning.py
shriganeshs: Yes. Thanks for pointing out.

app/util/autoversioning.py
shriganeshs: I would like to keep the rename logic for now to keep the backup and see the version change/difference.
But in future, I agree that we can clean this and make `write_package_version_file` to remove old file.

josephharrington:
@shriganeshs Looks like you merged a lint error. Can you fix that when you get a chance?

PR #353, opened 2017-06-01
Add mypy to requirements

josephharrington:
Mypy includes `typing` which enables smarter type hints. It would
also be cool to eventually start running the mypy linter on PRs.


boxcla:

PR #355, opened 2017-06-01
Add typing to requirements

josephharrington:
This enables smarter type hints in annotations. We should start
moving hints out of docstrings and into function annotations.

This is an alternative for #353.


boxcla:

PR #357, opened 2017-06-02
Pull logs from zip archive if it exists

josephharrington:
Currently logs are only available on the slave once a build has completed because the console output files are moved into an archive file on the master. This means that if a slave is taken down that logs for builds that ran on it will not be available via the console output API.

This updates the logic to pull logs from the zip archive if it exists on the master which will allow slaves to be taken offline without losing log functionality.


boxcla:

app/common/console_output.py
```diff
@@ -68,34 +85,33 @@ def _parse_from_offset(self, max_lines, offset_line):
...
-            # Figure out how many lines there are, so that we know which offset line to start appending
-            # console output from.
+        with self._file as f:
             for line in f:
-                if line.endswith("\n"):
-                    total_lines += 1
+                if not line.endswith(b'\n'):
+                    break  # last line; file still (probably) being written to
 
-            f.seek(0)
```
josephharrington: Zipfile does not support `seek` so this method had to be rewritten to use a different approach.

app/common/build_artifact.py
```diff
@@ -140,8 +144,49 @@ def _update_timing_file(self, timing_file_path, new_timing_data):
...
+        """
+        console_output = None
+
+        # Read directly from output file if it exists (while build is in progress).
+        artifact_dir = cls.atom_artifact_directory(build_id, subjob_id, atom_id, result_root=result_root)
+        output_file_path = os.path.join(artifact_dir, cls.OUTPUT_FILE)
+        if os.path.isfile(output_file_path):
+            console_output = ConsoleOutput.from_plaintext(output_file_path)
+
+        # Read from build artifact archive if it exists (after build is finished).
```
josephharrington: This part should be skipped if console_output exists from the previous lines.

app/common/console_output.py
```diff
@@ -1,64 +1,81 @@
...
+    @classmethod
+    def from_plaintext(cls, path: str) -> 'ConsoleOutput':
+        """
+        :param path: Path to a plaintext file containing output
+        """
+        file = open(path, 'rb')
+        return cls(file=file)
+
+    @classmethod
+    def from_zipfile(cls, zip_path, path_in_archive) -> 'ConsoleOutput':
```
josephharrington: Add type hints.

app/client/cluster_api_client.py
```diff
@@ -238,6 +240,11 @@ def _graceful_shutdown_slaves(self, body):
         )
         return response
 
+    def get_console_output(self, build_id: int, subjob_id: int, atom_id: int, max_lines: int=50, offset: int=0):
```
tjlee0909: Can we add docstrings explaining what the return type is and what the content is?

app/common/build_artifact.py
```diff
@@ -140,8 +144,49 @@ def _update_timing_file(self, timing_file_path, new_timing_data):
...
+        :return: The console output if it exists in the specified result_root, None if it does not exist
+        """
+        console_output = None
+
+        artifact_dir = cls.atom_artifact_directory(build_id, subjob_id, atom_id, result_root=result_root)
+        output_file_path = os.path.join(artifact_dir, cls.OUTPUT_FILE)
+        if os.path.isfile(output_file_path):
+            # Read directly from output file if it exists (while build is in progress).
+            console_output = ConsoleOutput.from_plaintext(output_file_path)
+        else:
```
tjlee0909: Having this distinction between build artifacts that have been zipped, and build artifacts that haven't been zipped, makes me think that eventually we'll want to redefine, and split up, this "BuildArtifact" class into two different classes.

Definitely outside the scope of this PR.

app/common/console_output.py
```diff
@@ -1,64 +1,81 @@
...
-        Return a segment of the console output.
+        Return a segment of the console output. Note this closes the file handle so can only be called once
+        per ConsoleOutput object.
 
-        :type max_lines: int
+        :param max_lines: The maximum number of lines of output to return
         :param offset_line: The starting line number in the console output from which the segment should
             return content for. If set to None, then reads content starting from the end.
-        :type offset_line: int | None
-        :rtype: ConsoleOutputSegment
```
tjlee0909: Why remove this rtype docstring? Did you mean to add some other docstring in place of it?

app/common/console_output.py
```diff
@@ -1,64 +1,81 @@
...
             for i in range(offset_line, offset_line + max_lines):
                 line = f.readline()
 
                 # We have reached the end of the file, or a line that has not finished being written to.
-                if line == '' or not line.endswith("\n"):
+                if not line.endswith(b'\n'):
                     break
 
-                console_output.append(line)
+                console_output.append(line.decode(encoding='utf-8', errors='replace'))
```
tjlee0909: Is this line due to some bug we saw?

app/common/console_output.py
```diff
@@ -1,64 +1,81 @@
...
             for i in range(0, offset_line):
-                # This is an error, meaning that there aren't even offset_line+1 lines in self.path.
-                if f.readline() == '':
-                    raise ValueError('offset: {} is higher than the total number of lines in file {}'.format(
-                        offset_line, self.path))
-
+                # This is an error, meaning that there aren't even offset_line+1 lines in the file.
+                if len(f.readline()) == 0:
+                    raise BadConsoleOutputRequestError(
+                        'offset {} is higher than the total number of lines: {}'.format(offset_line, total_lines))
```
tjlee0909: Cool!

app/common/console_output.py
```diff
@@ -68,34 +85,33 @@ def _parse_from_offset(self, max_lines, offset_line):
...
-                    total_lines += 1
+                if not line.endswith(b'\n'):
+                    break  # last line; file still (probably) being written to
 
-            f.seek(0)
-            offset_line = max(total_lines - max_lines, 0)
+                total_num_lines += 1
+                console_output_lines.append(line.decode(encoding='utf-8', errors='replace'))
+                if len(console_output_lines) > max_lines:
+                    console_output_lines.popleft()
```
tjlee0909: So pro Joey, using dequeues.

app/common/console_output.py
josephharrington: Yes! Good catch.

app/common/console_output.py
josephharrington: This is actually equivalent to what we were doing before. Previously this class would open the file in text mode so these params were passed with the `open()` call. Now it's opened in binary mode (to be compatible with how zip files are read) so the decoding happens here.

I don't remember if we added errors=replace in response to some bug, but I'd guess that was the case. We could probably dig it up in the git history.

app/common/build_artifact.py
josephharrington: Yeah, this class probably needs some love. It also feels pretty leaky to me with all its class methods to provide paths. I feel like ideally the actual paths to artifact directories would be completely internal to this class.

PR #358, opened 2017-06-05
Add randomness to secret

josephharrington:
Previously if using an autogenerated secret (the default behavior) then there was no randomness and the same one would be generated every time. Obviously bad.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 
 import argparse
-import hashlib
+import random
+import string
+
 import os
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


boxcla:

main.py
```diff
@@ -244,7 +246,9 @@ def _set_secret(config_filename):
     if 'secret' in Configuration and Configuration['secret'] is not None:
         secret = Configuration['secret']
     else:  # No secret found, generate one and persist it
-        secret = hashlib.sha512().hexdigest()
+        secret_length = 128
+        secret = ''.join(random.SystemRandom().choice(string.ascii_lowercase + string.digits)
+                         for _ in range(secret_length))
```
josephharrington: https://stackoverflow.com/a/23728630

tjlee0909: This indentation is a little bit odd to me, can we just set some variables so that we can get this statement in one line?

josephharrington: This passes linting just fine and is the default in IntelliJ but just for you TJ I can add a few more lines to reduce the line length.

PR #359, opened 2017-06-06
Reject the bad secret

josephharrington:
Because of the bug fixed in #358, there is a risk of config files containing an insecure secret. This makes ClusterRunner fail to start up with that specific secret.


boxcla:

josephharrington:
@gcurtis and I talked offline about whether this is a breaking change or not. We agree that critical security fixes, while potentially incompatible with existing deployments, should not be considered breaking as far as version numbering is concerned because users should be notified of and should address them immediately.

app/util/secret.py
```diff
@@ -23,6 +27,8 @@ def set(cls, secret):
             raise RuntimeError('Empty secret is not allowed!')
         if len(secret) < 8:
             raise RuntimeError('Secret must be at least 8 characters long!')
+        if secret == cls._BAD_SECRET:
+            raise RuntimeError('The secret in your clusterrunner.conf is insecure! Please generate a new one.')
```
cmcginty: Should there be more info about how to generate a new secret?

josephharrington: Sure!

PR #361, opened 2017-06-08
Adding .python-version file

nickzuber:
Running the `pyenv local cr` command will generate a `.python-version` file. We might as well commit this file to the repository since it should be there for virtualenv to automatically use the correct version of python (I think that's how pyenv works).


boxcla:
Hi @nickzuber, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

nickzuber:
CLA signed!

boxcla:
Verified that @nickzuber has just signed the CLA. Thanks, and we look forward to your contribution.

josephharrington:
@nickzuber For future reference, it would be great if you would squash your commits before merging (unless they are legitimately distinct, standalone changes). We don't want to pollute git history with a bunch of noise -- there is no benefit to having commits like "Address code review feedback" preserved for eternity.

Also please follow this commit message styling guide (specifically the one about "imperative mood"):
https://chris.beams.io/posts/git-commit/

Issue #344, opened 2017-05-18
List of metrics to add


josephharrington:
Here's a list of some metrics we could add that would be useful. Feel free to add to this list if you think of any others.

- Total build hours (summed atom time), potentially labeled with repo
    - Could have this on both the master and slave
- Counters for internal errors
    - setup_build failure
    - atomizer failure
    - postbuild tasks failure
        - zip file creation failure
    - API 500s
    - subjob payload write failure
    - build request handler fails to prepare build
    - failed outgoing network requests (connection refused and timeout)
- Some aggregate measure of expected vs actual atom time
- Gauge for number of idle vs. working slaves

PR #362, opened 2017-06-09
Add prometheus metric for time spent in each build state

ethomas2:


boxcla:

ethomas2:
First in a series of pr's I hope to put up adding metrics from #344

app/common/metrics.py
```diff
@@ -4,3 +4,8 @@
 request_latency = Histogram(  # pylint: disable=no-value-for-parameter
     'request_latency_seconds',
     'Latency of HTTP requests in seconds')
+
+build_state_duration_seconds = Histogram(  # pylint: disable=no-value-for-parameter
```
ethomas2: I played with the idea of adding a separate metric `api_queued_state` that's just the sum of the amount of time we spend in Queued, Preparing and Prepared times. If we want a separate metric I can do that, but we can always just sum them in prometheus.

app/master/build.py
```diff
@@ -61,12 +62,15 @@ def __init__(self, build_request):
         self._postbuild_tasks_are_finished = False  # WIP(joey): Remove and use build state.
         self._timing_file_path = None
 
+        leave_state_callbacks = {build_state: self._on_leave_state
+                                 for build_state in BuildState}
```
josephharrington: It might be simpler to roll this into `onchangestate` instead. (That's where the absolute timestamp values are already being set so it would make sense to record deltas there as well.)

From [Fysom docs](https://github.com/mriehl/fysom#callbacks):
> In addition, a generic onchangestate() callback can be used to call a single function for all state changes.

It looks like if you do that you also might not need to do an additional `time.time()` call in the callback since at that point the `transition_timestamps` value for the new state would already be set. That's not a huge deal though.

app/master/build.py
```diff
@@ -291,6 +295,15 @@ def _on_enter_error_state(self, event):
         self._error_message = getattr(event, 'error_msg', default_error_msg)
         self._logger.warning('Build {} failed: {}', self.build_id(), self._error_message)
 
+    def _on_leave_state(self, event):
+        start_time = self._state_machine.transition_timestamps.get(event.src)
```
josephharrington: How do you feel about making this data internal to the BuildFSM object? It's already recording the timestamp values internally so it wouldn't be that weird for it also to be recording durations/deltas. Right now the timestamps are the "entered" time, so adding deltas would be equivalent to recording both "entered" and "exited" times.

Issue #342, opened 2017-05-16
Unnecessary exception printed in master logs


josephharrington:
When someone makes a request for a build artifact that is not ready yet, we put the following error in the master logs:
```
[2017-05-15 17:02:25.845] 33228 ERROR   MasterTornadoTh cluster_base_ha Exception occurred during request to /v1/build/1/artifacts.zip.
Traceback (most recent call last):
  File "/Users/jharrington/.pyenv/versions/cr/lib/python3.4/site-packages/tornado/web.py", line 1346, in _when_complete
    callback()
  File "/Users/jharrington/.pyenv/versions/cr/lib/python3.4/site-packages/tornado/web.py", line 1367, in _execute_method
    self._when_complete(method(*self.path_args, **self.path_kwargs),
  File "<decorator-gen-17>", line 2, in get
  File "/Users/jharrington/.pyenv/versions/cr/lib/python3.4/site-packages/prometheus_client/core.py", line 879, in wrapped
    return func(*args, **kwargs)
  File "/Users/jharrington/code/ClusterRunner/app/web_framework/cluster_master_application.py", line 297, in get
    artifact_file_path = self.get_result_file_download_path(int(build_id))
  File "/Users/jharrington/code/ClusterRunner/app/web_framework/cluster_master_application.py", line 317, in get_result_file_download_path
    return self._cluster_master.get_path_for_build_results_archive(build_id)
  File "/Users/jharrington/code/ClusterRunner/app/master/cluster_master.py", line 313, in get_path_for_build_results_archive
    raise ItemNotReadyError('Build artifact file is not yet ready. Try again later.')
app.util.exceptions.ItemNotReadyError: Build artifact file is not yet ready. Try again later.
```

This is unnecessary logging for what is essentially a 404. I don't think we need to log anything on the master in this case.

This is compounded by the fact that the `clusterrunner build` command will retry forever until the artifact is ready, so if there's a problem with the build and the artifact is not created then this will fill up the logs on the master until the client is explicitly stopped.

PR #363, opened 2017-06-13
Remove exception logging for ItemNotReadyError

nickzuber:
Attempting to reach an item that isn't ready is basically a 404 and there isn't really a need to log anything to the master in this case.

Also when attempting to reach an artifact that isn't ready, the `clusterrunner build` command will continuously try to request it until the client is explicitly stopped -- this will spam the master logs.

Resolves: #342


boxcla:

app/web_framework/cluster_base_handler.py
```diff
@@ -38,8 +38,10 @@ def _handle_request_exception(self, ex):
         :param ex: The exception that was caught
         :type ex: Exception
         """
-        # _handle_request_exception() is called in the exception handler, so we can still use logger.exception.
-        self._logger.exception('Exception occurred during request to {}.', self.request.uri)
+        # Requests to access an item that isn't ready shouldn't log anything
+        if type(ex) is not ItemNotReadyError:
```
josephharrington: While you're here, do you think there are any other exception types we should remove from logging? 404s seem like they'd be another good one to omit. Thoughts?

app/web_framework/cluster_base_handler.py
```diff
@@ -38,8 +38,10 @@ def _handle_request_exception(self, ex):
         :param ex: The exception that was caught
         :type ex: Exception
         """
-        # _handle_request_exception() is called in the exception handler, so we can still use logger.exception.
-        self._logger.exception('Exception occurred during request to {}.', self.request.uri)
+        # Requests to access an item that isn't ready shouldn't log anything
+        if not isinstance(ex, ItemNotReadyError):
```
josephharrington: While you're here, do you think there are any other exception types we should remove from logging? 404s seem like they'd be another good one to omit. Thoughts?

nickzuber: Good idea, that could be here as well. The other types of exceptions seem like they'd be important to have in the logs, so do we just want to `ItemNotReadyError` and `ItemNotFoundError`?

josephharrington: Sounds good to me. :)

PR #364, opened 2017-06-13
Add developer setup link to readme

### LINKED ISSUE #131: Developer Environment setup is not clear
From looking at the README file, the setup instructions are not clear. I found in the README, the [Contributing guidelines](http://www.clusterrunner.com/docs/development-guide/) but there is no mention about what it takes to setup the environment there. I was confused on which version of Python was necessary for tests to work. I used virtualenv to setup my environment because Python3.4 (required to run tests) is not setup as default.

I am running on Mac OS X, and I had to run the following steps to get started.

```
# Out of band, I needed to update my Homebrew Python install to include Python3.4
# The latest version of Python I had was only Python3.3, and it did not work.

virtualenv -p `which python3.4` venv
source venv/bin/activate
pip install -r requirements.txt
```

After the install make test does _mostly_ work. I am still having trouble running the functional tests out of the box. Maybe I am missing some software, but I haven't dug into the tests to see if I am missing some software yet.

Would be nice to have it documented someplace. Would you like me to open up a PR for this? Let me know.

josephharrington:
Resolves #131.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,18 +18,20 @@ The entire process of installing, initializing, and executing tests through Clus
 
 Give it a shot by starting at our [documentation](http://www.clusterrunner.com) site.
 
-
 ## Read the Documentation
 
 Check out our [documentation](http://www.clusterrunner.com) site for everything you need to know.
 
 - [Getting Started](http://www.clusterrunner.com/docs/home/)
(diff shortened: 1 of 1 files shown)


boxcla:

PR #365, opened 2017-06-13
Fix AppVeyor badge

josephharrington:
We changed the account that AppVeyor tests run under but didn't
change the badge.


boxcla:

PR #367, opened 2017-06-14
Hide stderr in autoversioning

josephharrington:
If autoversioning-related git commands fail, the app just reports its version as 0.0.0. But currently the stderr from the failing git commands is not suppressed. This can be seen in TravisCI output as a bunch of scary "fatal: Needed a single revision" lines, but we should just hide those.


boxcla:

Issue #324, opened 2016-11-07
Url with missing scheme can create ambiguous clone directories


josephharrington:
When the master gets a git-type build request with a url that looks something like `ssh://scm.dev.box.net:12345/awesome-project`, it clones it to the directory `~/.clusterrunner/repos/master/scm.dev.box.net12345/awesome-project/`. That's correct and unambiguous across hosts.

However if we get something like `git.dev.box.net:Productivity/ClusterRunnerHealthCheck`, the clone directory will be `~/.clusterrunner/repos/master/Productivity/ClusterRunnerHealthCheck/` which does not include the hostname, so will be ambiguous across hosts and can lead to unpredictable behavior.

For example, if we send a build request for a different host with the same repo name ("Productivity/ClusterRunnerHealthCheck") the clusterrunner master will reuse a repo clone that has its remotes pointed at a different host.

This is an issue with the `_generate_path_from_repo_url` method in app/project_type/git.py, and ultimately an unexpected behavior of stdlib `urllib.parse.urlparse`:

```python
urlparse('git.dev.box.net:Productivity/ClusterRunnerHealthCheck')
# ParseResult(scheme='git.dev.box.net', netloc='', path='Productivity/ClusterRunnerHealthCheck', params='', query='', fragment='')
```
Note above that the `scheme` field, not the `netloc` field, contains the hostname, which is surprising.

Also note that the clone itself *does* work with that url, so instead of requiring a scheme on urls we probably just need to be smarter in dealing with various input formats.

nickzuber:
@josephharrington  Is the goal here with `_generate_path_from_repo_url` to just generate a deterministic path for a repository url or is there some type of convention we want to stick with? Like is there a particular reason we don't want to include a scheme _and_ netloc when generating the url?

Also what other input formats should we expect besides the two mentioned here?

josephharrington:
@nickzuber I don't think the convention is too important as long as it's unique. That said, I don't love including scheme as the *only* fix for this issue -- it is surprising that the input url is not parsed correctly and could lead to other bugs in the future if someone changes this code and expects it to act sane.

Another consideration is that we probably don't want to change the generated path for most repos if we can help it. If the path changes then that will cause repos to be recloned the next time clusterrunner is upgraded which depending on the repo can be undesirable.

@tjlee0909 @nadeemahmad  Do you know if we still do prepopulation of repos on our slave images? If so, then we'd want to be extra careful with this change to make sure CR is still looking for those repos in the right place.

[...]

PR #369, opened 2017-06-14
Generate repo url with scheme if netloc is empty

nickzuber:
For certain urls, the `urlparse` function won't behave as expected, and the url's netloc field will be empty. In this case, we can just use the scheme field to mitigate ambiguitiy across hosts.

Resolves: #324

### DIFF
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -41,7 +41,9 @@ class Git(ProjectType):
         url_folder_path_parts = url_full_path_parts[:-1]
         repo_directory = os.path.join(
             base_sys_path,
-            url_components.netloc.replace(':', ''),  # remove colons from netloc (e.g. turn example:8000 to example8000)
+            # Remove colons from netloc/scheme (e.g. turn example:8000 to example8000).
+            # If netloc is empty the url is malformed -- use scheme instead.
+            (url_components.netloc or url_components.scheme).replace(':', ''),
             *url_folder_path_parts
         )
         # remove '-'s as PHP Intl extension doesn't appear to work if the repo is in a directory with a dash in the path
--- a/app/project_type/git.py
+++ b/app/project_type/git.py
@@ -43,7 +43,7 @@ class Git(ProjectType):
         url_host_path_part = url_components.netloc or url_components.scheme
         repo_directory = os.path.join(
             base_sys_path,
-            url_host_path_part.replace(':', ''), # Remove colons (e.g. turn example:8000 to example8000).
+            url_host_path_part.replace(':', ''),  # Remove colons (e.g. turn example:8000 to example8000).
             *url_folder_path_parts
         )
         # remove '-'s as PHP Intl extension doesn't appear to work if the repo is in a directory with a dash in the path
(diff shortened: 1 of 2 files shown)


boxcla:

app/project_type/git.py
```diff
@@ -41,7 +41,9 @@ def _generate_path_from_repo_url(base_sys_path, url):
         url_folder_path_parts = url_full_path_parts[:-1]
         repo_directory = os.path.join(
             base_sys_path,
-            url_components.netloc.replace(':', ''),  # remove colons from netloc (e.g. turn example:8000 to example8000)
+            # Remove colons from netloc/scheme (e.g. turn example:8000 to example8000).
+            # If netloc is empty the url is malformed -- use scheme instead.
+            (url_components.netloc or url_components.scheme).replace(':', ''),
```
josephharrington: Maybe create an intermediate variable that holds the result of (netloc or scheme)? It might make this a bit easier to read. I think the comment is good though.

test/unit/project_type/test_git.py
```diff
@@ -286,6 +286,31 @@ def fake_popen_constructor(command, stdout, stderr, *args, **kwargs):
...
+        Configuration['repo_directory'] = join(expanduser('~'), '.clusterrunner', 'repos')
+        url = 'ssh://scm.dev.box.net:12345/awesome-project'
+
+        actual_repo_directory = Git.get_full_repo_directory(url)
+        expected_repo_directory = join(
+            Configuration['repo_directory'],
+            'scm.dev.box.net12345',
+            'awesomeproject',
+        )
+        self.assertEqual(expected_repo_directory, actual_repo_directory)
```
josephharrington: These look like good tests to combine into a single genty-driven test. You could also probably combine it with the `test_get_full_repo_directory` above.

If you search the test directory for genty you'll find examples. [Here's one.](https://github.com/box/ClusterRunner/blob/0d2f5b54/test/integration/common/test_console_output.py#L38-L56)

I like to combine similar tests like that because it reduces duplication, but more importantly it makes it easier to see that these are really the same test just with different inputs.

app/project_type/git.py
nickzuber: Good idea, do you have an opinion on what we should call it?

josephharrington: No strong opinions. Maybe something like `url_host_path_part`? :woman_shrugging:

test/unit/project_type/test_git.py
```diff
@@ -79,20 +79,41 @@ def test_execute_command_in_project_type_specifies_cwd_if_doesnt_exist(self):
...
+        ),
+        no_netloc=(
+            'git.dev.box.net:Productivity/ClusterRunnerHealthCheck',
+            join(
+                'git.dev.box.net',
+                'Productivity',
+                'ClusterRunnerHealthCheck',
+            )
+        ),
+    )
```
nickzuber: If there's a better or cleaner way to format this I'd be happy to change it

test/unit/project_type/test_git.py
```diff
@@ -79,20 +79,41 @@ def test_execute_command_in_project_type_specifies_cwd_if_doesnt_exist(self):
...
-
-        expected_repo_sys_path = join(
-            Configuration['repo_directory'],
-            'scm.example.com',
-            'path',
-            'to',
-            'project',
+        expected_repo_path = join(
+            Configuration['repo_directory'], 
+            expected_repo_path_without_base
```
cmcginty: It's good to leave the trailing comma on the last element when you split the args over multiple lines.

nickzuber: Good idea, makes for cleaner diffs

Issue #287, opened 2016-01-12
When slave host goes down (but no shutdown command is issued), slave should be marked as is_alive: false


tjlee0909:

josephharrington:
Slaves *do* get marked as `is_alive: false`, but not until the master tries to allocate the slave to a build. And then it takes 15 minutes due to #317.

We should have the master occasionally check slave connectivity so that we know slaves' online/offline status before the cluster is under load.

josephharrington:
The fix for this may (or may not) be related to #214. Both are related to the master and slaves checking  that each other are still up and running while idle.

wjdhollow:
@nadeemahmad

wjdhollow:
@josephharrington mentioned that to reduce the number of messages the master needs to send, it could keep track of the last time it last heard from a slave to determine which nodes to ping.

josephharrington:
On the master side, we can keep track of the last time it heard from each slave. Then we'd add a thread on the master that iterates through the list of slaves that it hasn't heard from in awhile and does an is_alive check on each one.

This approach would take advantage of the fact that we don't need to dumbly ping slaves that are actively reporting results. Additionally, addressing #214 would ensure that even when all slaves are idle that the master is only actively pinging slaves that are misbehaving (haven't phoned home recently). In that case, the threshold on the master for when to ping a slave it thinks has died should be greater than whatever interval the slave is configured to phone home.

PR #370, opened 2017-06-14
More metrics

ethomas2:
Adds 3 of the 4 metrics from #344 
* serialized build time
* metrics for number of active/idle slaves
* aggregate measure of actual vs expected atom time

### DIFF
--- a/app/common/metrics.py
+++ b/app/common/metrics.py
@@ -1,4 +1,7 @@
-from prometheus_client import Histogram
+from typing import Callable, Iterator, List
+
+from prometheus_client import Histogram, REGISTRY
+from prometheus_client.core import GaugeMetricFamily
 
 
 request_latency = Histogram(  # pylint: disable=no-value-for-parameter
@@ -9,3 +12,43 @@ build_state_duration_seconds = Histogram(  # pylint: disable=no-value-for-parame
     'build_state_duration_seconds',
     'Total amount of time a build spends in each build state',
     ['state'])
+
+serialized_build_time_seconds = Histogram(  # pylint: disable=no-value-for-parameter
+    'serialized_build_time_seconds',
+    'Total amount of time that would have been consumed by builds if all work was done serially')
+
+
+class SlavesCollector:
+    """
+    Prometheus collector to collect the total number of alive/dead/idle slaves connected to the master.
+    collect() is called once each time prometheus scrapes the /metrics endpoint. This class ensures that
+    1. The list of slaves only gets iterated through once per scrape
+    2. A single slave is is not double counted in 2 states
+    """
+
+    _slaves_collector_is_registered = False
+
+    def __init__(self, get_slaves: Callable[[], List['app.master.slave.Slave']]):
+        self._get_slaves = get_slaves
+
+    def collect(self) -> Iterator[GaugeMetricFamily]:
+        active, idle, dead = 0, 0, 0
+        for slave in self._get_slaves():
+            if not slave.is_alive(use_cached=True):
+                dead += 1
+            elif slave.current_build_id is not None:
+                active += 1
+            else:
+                idle += 1
+
+        slaves_gauge = GaugeMetricFamily('slaves', 'Total number of slaves', labels=['state'])
+        slaves_gauge.add_metric(['active'], active)
+        slaves_gauge.add_metric(['idle'], idle)
+        slaves_gauge.add_metric(['dead'], dead)
+        yield slaves_gauge
+
+    @classmethod
+    def register_slaves_metrics_collector(cls, get_slaves: Callable[[], List['app.master.slave.Slave']]):
+        if not cls._slaves_collector_is_registered:
+            REGISTRY.register(SlavesCollector(get_slaves))
+            cls._slaves_collector_is_registered = True
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -8,6 +8,8 @@ from threading import Lock, Thread
 import time
 import uuid
 
+from typing import Dict
+
 from app.common.build_artifact import BuildArtifact
 from app.master.build_fsm import BuildFsm, BuildEvent, BuildState
 from app.master.build_request import BuildRequest
@@ -19,7 +21,7 @@ from app.util.exceptions import ItemNotFoundError
 import app.util.fs
 from app.util.log import get_logger
 from app.util.single_use_coin import SingleUseCoin
-from app.common.metrics import build_state_duration_seconds
+from app.common.metrics import build_state_duration_seconds, serialized_build_time_seconds
 
 
 class Build(object):
@@ -471,7 +473,9 @@ class Build(object):
         data. This method also transitions the FSM to finished after the postbuild tasks are complete.
         """
         try:
-            self._create_build_artifact()
+            timing_data = self._read_subjob_timings_from_results()
+            self._create_build_artifact(timing_data)
+            serialized_build_time_seconds.observe(sum(timing_data.values()))
             self._delete_temporary_build_artifact_files()
             self._postbuild_tasks_are_finished = True
             self._state_machine.trigger(BuildEvent.POSTBUILD_TASKS_COMPLETE)
@@ -480,10 +484,10 @@ class Build(object):
             self._logger.exception('Postbuild tasks failed for build {}.'.format(self._build_id))
             self.mark_failed('Postbuild tasks failed due to an internal error: "{}"'.format(ex))
 
-    def _create_build_artifact(self):
+    def _create_build_artifact(self, timing_data: Dict[str, float]):  # pylint: disable=unsubscriptable-object
         self._build_artifact = BuildArtifact(self._build_results_dir())
         self._build_artifact.generate_failures_file()
-        self._build_artifact.write_timing_data(self._timing_file_path, self._read_subjob_timings_from_results())
+        self._build_artifact.write_timing_data(self._timing_file_path, timing_data)
         self._artifacts_tar_file = app.util.fs.tar_directory(self._build_results_dir(),
                                                              BuildArtifact.ARTIFACT_TARFILE_NAME)
         temp_tar_path = None
--- a/app/master/cluster_master.py
+++ b/app/master/cluster_master.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 import os
 
 from app.common.cluster_service import ClusterService
+from app.common.metrics import SlavesCollector
 from app.master.build import Build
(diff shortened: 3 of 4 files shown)


boxcla:

ethomas2:
I can split these up into seperate pr's, but I thought they belonged together. I want to add another commit on this pr for the last metric from #344 before I merge.

app/common/metrics.py
```diff
@@ -9,3 +12,48 @@
...
+                idle += 1
+
+        slaves_gauge = GaugeMetricFamily('slaves', 'Total number of slaves', labels=['state'])
+        slaves_gauge.add_metric(['active'], active)
+        slaves_gauge.add_metric(['idle'], idle)
+        slaves_gauge.add_metric(['dead'], dead)
+        yield slaves_gauge
+
+
+slaves_collector_is_registered = False
```
josephharrington: Private? Alternatively this could go in the SlavesCollector class itself and the register method could be a class method.

app/master/build.py
```diff
@@ -471,19 +473,21 @@ def _perform_async_postbuild_tasks(self):
         data. This method also transitions the FSM to finished after the postbuild tasks are complete.
         """
         try:
-            self._create_build_artifact()
+            timing_data = self._read_subjob_timings_from_results()
+            self._create_build_artifact(timing_data)
```
josephharrington: It seems like there's not much point to moving the `self._read_subjob_timings_from_results()` call out of `self._create_build_artifact()`.

app/master/build.py
```diff
@@ -523,6 +527,22 @@ def _delete_temporary_build_artifact_files(self):
         end_time = time.time() - start_time
         self._logger.info('Completed deleting artifact files for {}, took {:.1f} seconds.', self._build_id, end_time)
 
+    def _update_prometheus_metrics(self):
+        serialized_time_val, estimation_error_val = 0, 0
+        for subjob in self._all_subjobs_by_id.values():
+            for atom in subjob.atoms:
+                if atom.actual_time is None:
+                    self._logger.warning('Could not load atom.actual_time for atom {} build {}', atom.command_string, self._build_id)
```
josephharrington: Is this warning necessary? It seems like this would have already warned when the subjob result is processed [here](https://github.com/box/ClusterRunner/blob/0d2f5b5/app/master/subjob.py#L141).

app/master/build.py
```diff
@@ -523,6 +527,22 @@ def _delete_temporary_build_artifact_files(self):
...
+                    return
+                if atom.expected_time is None:
+                    self._logger.warning('Could not load atom.expected_time for atom {} build {}', atom.command_string, self._build_id)
```
josephharrington: I think this warning will create a ton of noise the first time a job is run since there are no expected times in that case. How do you feel about removing this warning as well? I feel like it would be reasonable for `atom.expected_time` to be set to `None` in the case of a new atom, even if that's not the current behavior.

Also, does it make sense to return if expected time is none? It seems like even if we don't have any expected times set for a build that we could still record the serialized build time.

app/master/build.py
```diff
@@ -523,6 +527,22 @@ def _delete_temporary_build_artifact_files(self):
...
+                    return
+                serialized_time_val += atom.actual_time
+                estimation_error_val += abs(atom.actual_time - atom.expected_time)
+
+        serialized_build_time_seconds.observe(serialized_time_val)
+        build_estimation_error.observe(estimation_error_val)
```
josephharrington: I'm starting to wonder if build_estimation_error is very useful to have as a metric. I think it would definitely be useful on a per-build level (so it would be cool to add it to the API response), but I'm not sure if it's actionable on a global level. What do you think?

app/common/metrics.py
```diff
@@ -9,3 +12,48 @@
...
+    1. The list of slaves only gets iterated through once per scrape
+    2. A single slave is is not double counted in 2 states
+    """
+    def __init__(self, get_slaves: Callable[[], List['app.master.slave.Slave']]):
+        self._get_slaves = get_slaves
+
+    def collect(self):
+        active, idle, dead = 0, 0, 0
+        for slave in self._get_slaves():
+            if not slave.is_alive(use_cached=True):
```
tjlee0909: I'd want to set use_cached=False here.

The slave's dead/alive status is only updated whenever the master makes network calls to the slave. If a slave is working on a subjob, then the master will never contact that slave again until the slave makes an API call to the master indicating that it completed the subjob--which could be never, if the slave were to die while running a subjob.

app/master/build.py
tjlee0909: 👍 to what Joey said.

I don't think this merits a warning.

app/common/metrics.py
josephharrington: TJ, it sounds like you might be hinting at #287, which I still think we should do but probably not like this. If we set `use_cached=False` here I think every time Prometheus went to scrape metrics (every 15 seconds?) we'd synchronously send a request to every slave. I think this would not be the most efficient way to do heartbeats to slaves and we definitely wouldn't want to do it while we were handling an API request.

@ethomas2 I was assuming that `collect()` gets called when Prometheus scrapes the metrics endpoint, but is that correct?

tjlee0909: Having it being down on one thread, every 15 seconds, would be pretty bad, I agree. I'd imagine it'd take as long as 15 seconds to even do this for all 200 slaves.

That said, I don't think efficiency is much of an issue here. On the slave side, responding to one very light API request (that does nothing on the slave) every 15 seconds is nothing--though I'd be more comfortable if this were more like a minute.

I am concerned about this running on the master every 15 seconds though for all 200 slaves.

tjlee0909: Ok, I'm fine with leaving it as is.

app/common/metrics.py
```diff
@@ -9,3 +12,48 @@
...
+serialized_build_time_seconds = Histogram(  # pylint: disable=no-value-for-parameter
+    'serialized_build_time_seconds',
+    'Total amount of time that would have been consumed by builds if all work was done serially')
+
+build_estimation_error = Histogram(  # pylint: disable=no-value-for-parameter
+    'build_estimation_error',
+    'For each build, the sum of the absolute value of the difference between the estimated and actual time for each atom to complete')
+
+
+class SlavesCollector(object):
```
ethomas2: Done

app/common/metrics.py
ethomas2: Done. Did the class method approach.

app/master/build.py
```diff
@@ -484,7 +484,7 @@ def _perform_async_postbuild_tasks(self):
             self._logger.exception('Postbuild tasks failed for build {}.'.format(self._build_id))
             self.mark_failed('Postbuild tasks failed due to an internal error: "{}"'.format(ex))
 
-    def _create_build_artifact(self, timing_data: Dict[str, float]):
+    def _create_build_artifact(self, timing_data: Dict[str, float]):  # pylint: disable=unsubscriptable-object
```
ethomas2: Don't remember why I put that in there. I took it out.

app/master/build.py
ethomas2: Yeah. I originally took it out because I thought timing_data had both `atom.actual_time` and `atom.expected_time`. I was gonna use timing_data to calculate the build_estimation_error metric, but
1.  timing_data doesn't have actual_time
2. I'm getting rid of build_estimation_error anyway

app/master/build.py
ethomas2: Killed the whole method.

app/master/build.py
ethomas2: Killed it

app/master/build.py
ethomas2: Killed the whole method. Calculating serialized_build_time every time. Not doing any warnings.

app/common/metrics.py
```diff
@@ -9,3 +12,48 @@
...
+    """
+    Prometheus collector to collect the total number of alive/dead/idle slaves connected to the master.
+    collect() is called once each time prometheus scrapes the /metrics endpoint. This class ensures that
+    1. The list of slaves only gets iterated through once per scrape
+    2. A single slave is is not double counted in 2 states
+    """
+    def __init__(self, get_slaves: Callable[[], List['app.master.slave.Slave']]):
+        self._get_slaves = get_slaves
+
+    def collect(self):
```
ethomas2: Done.

app/master/build.py
```diff
@@ -471,19 +473,21 @@ def _perform_async_postbuild_tasks(self):
         data. This method also transitions the FSM to finished after the postbuild tasks are complete.
         """
         try:
-            self._create_build_artifact()
+            timing_data = self._read_subjob_timings_from_results()
+            self._create_build_artifact(timing_data)
             self._delete_temporary_build_artifact_files()
+            self._update_prometheus_metrics()
```
ethomas2: No reason. I don't want it to be after the state_machine.trigger() event. That event puts the build in the finished state, and as far as I can tell right now nothing happens to a build after it finishes. I'd like to keep it that way if I can.

app/master/build.py
```diff
@@ -523,6 +527,22 @@ def _delete_temporary_build_artifact_files(self):
...
+                if atom.actual_time is None:
+                    self._logger.warning('Could not load atom.actual_time for atom {} build {}', atom.command_string, self._build_id)
+                    return
+                if atom.expected_time is None:
+                    self._logger.warning('Could not load atom.expected_time for atom {} build {}', atom.command_string, self._build_id)
+                    return
+                serialized_time_val += atom.actual_time
+                estimation_error_val += abs(atom.actual_time - atom.expected_time)
+
+        serialized_build_time_seconds.observe(serialized_time_val)
```
ethomas2: I agree, and maybe i'll do that in a follow up change, but request_latency was already a variable before I got here and it was lowercase. For consistency I want to keep mine lowercase as well.  If request_latency was only used in a couple places then i'd just change it here, but it's used in 40 places. I don't want to clutter up my diff with 40 lines that don't belong in it.

app/common/metrics.py
ethomas2: Yeah, collect gets called once per scrape.

app/master/build.py
ethomas2: Actually, I just put it back. The logic looks a little different now. I think it's warranted now.

josephharrington: Do you want to move the timing data call back into `self._create_build_artifact`?

josephharrington: Ah, nevermind.

PR #373, opened 2017-06-16
Fail functional tests if builds are slow

josephharrington:
We see some flakiness in CI related to builds not being in an expected state at the end of a test. There are calls to block a test until a build reaches a certain state, but if the build doesn't reach that state within the timeout we currently just ignore that and continue on. This change adds assertions so that tests will fail immediately if a build is taking too long.

This is a step towards making sources of flakiness more obvious in the functional tests.

### DIFF
--- a/test/functional/test_cluster_basic.py
+++ b/test/functional/test_cluster_basic.py
@@ -30,7 +30,8 @@ class TestClusterBasic(BaseFunctionalTestCase):
             'project_directory': project_dir.name,
         })
         build_id = build_resp['build_id']
-        master.block_until_build_finished(build_id, timeout=30)
+        self.assertTrue(master.block_until_build_finished(build_id, timeout=30),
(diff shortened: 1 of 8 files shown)


boxcla:

test/functional/test_cluster_basic.py
```diff
@@ -93,8 +95,10 @@ def test_slave_reconnection_does_not_take_down_master(self):
             'project_directory': project_dir.name,
         })
 
-        master.block_until_build_finished(build_1['build_id'], timeout=30)
-        master.block_until_build_finished(build_2['build_id'], timeout=30)
+        self.assertTrue(master.block_until_build_finished(build_1['build_id'], timeout=45),
```
josephharrington: This timeout seems long but 30 seconds is slightly too slow for AppVeyor.

PR #374, opened 2017-06-20
Add metrics for internal errors

ethomas2:


boxcla:

app/web_framework/cluster_base_handler.py
```diff
@@ -46,6 +47,11 @@ def _handle_request_exception(self, ex):
         self.set_status(status_code)
         self.finish()
 
+    def on_finish(self):
+        http_request_duration_seconds.labels(self.request.method,  # pylint: disable=no-member
+                                             self.route,  # pylint: disable=no-member
```
josephharrington: `self.route` is not a thing, is it? I imagine this would be failing for every request.

app/master/build.py
```diff
@@ -11,17 +11,17 @@
...
 from app.project_type.project_type import ProjectType
 from app.util import util
 from app.util.conf.configuration import Configuration
 from app.util.counter import Counter
 from app.util.exceptions import ItemNotFoundError
-import app.util.fs
 from app.util.log import get_logger
 from app.util.single_use_coin import SingleUseCoin
-from app.common.metrics import build_state_duration_seconds, serialized_build_time_seconds
+import app.util.fs
```
ethomas2: I sorted this with vim by piping the imports through the sort bash command. The sort bash command put it at the end because it starts with `import` and all the other lines start with `from`. I put it back (intellij wins this one)

app/master/build_request_handler.py
```diff
@@ -100,6 +102,8 @@ def _prepare_build_async(self, build, project_lock):
                         self._scheduler_pool.add_build_waiting_for_slaves(build)
 
             except Exception as ex:  # pylint: disable=broad-except
+                if isinstance(ex, AtomizerError):
+                    internal_errors.labels(ErrorType.AtomizerFailure).inc()  # pylint: disable=no-member
```
ethomas2: Done.

app/web_framework/cluster_base_handler.py
ethomas2: Yeah, this was completely broken. Thought I tested it, apparently I didn't. Fixed.

app/common/metrics.py
```diff
@@ -17,6 +20,20 @@
     'serialized_build_time_seconds',
     'Total amount of time that would have been consumed by builds if all work was done serially')
 
+internal_errors = Counter(
+    'internal_errors',
+    'Total number of internal errors',
+    ['type'])
+
+
+class ErrorType(str, Enum):  # WIP(joey): Remove this class.
```
ethomas2: whoops. Done.

app/web_framework/cluster_base_handler.py
```diff
@@ -30,6 +32,14 @@ def __init__(self, *args, **kwargs):
         PreconditionFailedError: http.client.PRECONDITION_FAILED,
     }
 
+    def initialize(self, route_node=None, **kwargs):
+        """
+        :param route_node: Each handler is associated with a RouteNode
+        :type route_node: RouteNode
+        """
+        self._route_node = route_node
+        super().initialize(**kwargs)
```
josephharrington: Do you need `initialize()` at all if you're setting route_node in `__init__`?

app/web_framework/cluster_base_handler.py
```diff
@@ -20,6 +21,7 @@ class ClusterBaseHandler(tornado.web.RequestHandler):
 
     def __init__(self, *args, **kwargs):
         self._logger = log.get_logger(__name__)
+        self._route_node = kwargs.get('route_node')
```
josephharrington: You could `kwargs.pop` here maybe?

app/web_framework/cluster_base_handler.py
```diff
@@ -46,6 +56,12 @@ def _handle_request_exception(self, ex):
         self.set_status(status_code)
         self.finish()
 
+    def on_finish(self):
+        if self._route_node is not None:
+            http_request_duration_seconds.labels(self.request.method,  # pylint: disable=no-member
+                                                 self._route_node.regex(),  # pylint: disable=no-member
```
josephharrington: You might not need the `disable=no-member` now. I wonder if you were seeing an error on `self.route` before because that legitimately did not exist.

app/common/metrics.py
```diff
@@ -17,6 +20,23 @@
...
+class ErrorType(str, Enum):
+    AtomizerFailure = 'AtomizerFailure'
+    NetworkRequestFailure = 'NetworkRequestFailure'
+    PostBuildFailure = 'PostBuildFailure'
+    SetupBuildFailure = 'SetupBuildFailure'
+    SubjobWriteFailure = 'SubjobWriteFailure'
+    ZipFileCreationFailure = 'ZipFileCreationFailure'
+
+    def __str__(self):
+        return self.value
```
josephharrington: Do we need this? If so, mind adding a comment about why? Inheriting from `str` should give this to us for free but it depends on what consuming code is doing. I guess it's needed if consuming code does `str(...)` anywhere:
```python
In [64]: class Test(str, Enum):
    ...:     apple = 'apple'
    ...:     black = 'black'
    ...:     crust = 'crust'

In [65]: 'hey {}'.format(Test.apple)
Out[65]: 'hey apple'

In [66]: str(Test.apple)
Out[66]: 'Test.apple'

In [67]: format(Test.apple)
Out[67]: 'apple'
```

app/web_framework/cluster_base_handler.py
ethomas2: Didn't mean to leave this here. Deleting it and doing it in `initialize()`

app/web_framework/cluster_base_handler.py
ethomas2: See above comment.

app/common/metrics.py
ethomas2: It's so that the /metrics endpoint has metrics that look like
`internal_errors{type="PostBuildFailure"} 1.0`
instead of 
`internal_errors{type="ErrorType.PostBuildFailure"} 1.0`

I don't think I need to inherit from `str`, i'm just doing it because all the other enums in cluster runner do it.

I'll add a comment.

josephharrington: Yeah, I figured that was it. `prometheus_client` must be using `str()` instead of `format()` internally. The reason the other enums inherit from `str` is because `json.dumps` must be using `format()` internally so we get the correct formatting for free.

app/web_framework/cluster_base_handler.py
ethomas2: Yup. Done.

PR #377, opened 2017-06-26
Stop killing slaves when setup_build fails

### LINKED ISSUE #116: When setup fails on Slave, the slave should not be deliberately taken down
Right now, we have the slave be taken down when build_setup fails. I believe we had originally deliberately designed it this way because we feared that individual slaves might become defective, and that if a slave fails setup on a certain build, that it will likely fail setup on all subsequent builds.

After running ClusterRunner for a while, we've realized that this assumption is incorrect. Perhaps the master should keep track of how many of the slaves that it gave subjobs to (for this particular build) have failed setup--and if most/all of them have failed, fail the build instead of taking slaves offline.

ethomas2:
Fixes #116

### DIFF
--- a/app/master/cluster_master.py
+++ b/app/master/cluster_master.py
@@ -3,7 +3,7 @@ import os
 
 from app.common.cluster_service import ClusterService
 from app.common.metrics import ErrorType, SlavesCollector, internal_errors
-from app.master.build import Build
+from app.master.build import Build, MAX_SETUP_FAILURES
 from app.master.build_request import BuildRequest
 from app.master.build_request_handler import BuildRequestHandler
 from app.master.build_scheduler_pool import BuildSchedulerPool
@@ -219,7 +219,13 @@ class ClusterMaster(ClusterService):
         :type slave: Slave
         """
         internal_errors.labels(ErrorType.SetupBuildFailure).inc()  # pylint: disable=no-member
-        raise BadRequestError('Setup failure handling on the master is not yet implemented.')
+        build = self.get_build(slave.current_build_id)
+        build.setup_failures += 1
+        if build.setup_failures >= MAX_SETUP_FAILURES:
+            build.cancel()
+            build.mark_failed('Setup failed on this build more than {} times. Failing the build.'
+                              .format(MAX_SETUP_FAILURES))
+        slave.teardown()
 
     def handle_request_for_new_build(self, build_params):
         """
@@ -279,7 +285,7 @@ class ClusterMaster(ClusterService):
         build = self._all_builds_by_id[int(build_id)]
         slave = self._all_slaves_by_url[slave_url]
         # If the build has been canceled, don't work on the next subjob.
-        if not build.is_finished:  # WIP(joey): This check should be internal to the Build object.
+        if not build.is_finished or build.has_error:  # WIP(joey): This check should be internal to the Build object.
             try:
                 build.complete_subjob(subjob_id, payload)
             finally:
--- a/test/functional/job_configs.py
+++ b/test/functional/job_configs.py
@@ -175,3 +175,29 @@ BasicSleepingJob:
     expected_num_subjobs=5,
     expected_num_atoms=5,
 )
+
+FAILING_SETUP_JOB = FunctionalTestJobConfig(
+    config={
+        'posix': """
+FailingSetupJob:
+    setup_build:
+        - exit 1
+    commands:
+        - echo $TOKEN
+    atomizers:
+        - TOKEN: seq 0 4
+""",
+        'nt': """
+FailingSetupJob:
+    setup_build:
+        - exit /b 1
+    commands:
+        - echo !TOKEN!
+    atomizers:
+        - TOKEN: seq 0 4
+"""
+    },
+    expected_to_fail=True,
+    expected_num_subjobs=5,
+    expected_num_atoms=5
+)
--- a/test/functional/test_cluster_basic.py
+++ b/test/functional/test_cluster_basic.py
@@ -6,9 +6,10 @@ import yaml
 from genty import genty, genty_dataset
 
 from app.common.build_artifact import BuildArtifact
+from app.master.build import BuildStatus
 from test.framework.functional.base_functional_test_case import BaseFunctionalTestCase
 from test.framework.functional.fs_item import Directory, File
-from test.functional.job_configs import BASIC_FAILING_JOB, BASIC_JOB, JOB_WITH_SETUP_AND_TEARDOWN
+from test.functional.job_configs import BASIC_FAILING_JOB, BASIC_JOB, FAILING_SETUP_JOB, JOB_WITH_SETUP_AND_TEARDOWN
 
 
 @genty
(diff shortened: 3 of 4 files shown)


boxcla:

ethomas2:
Code copied from https://github.com/Medium/ClusterRunner/commit/fd38c99155eb0373815b75230f90f09295861770

app/master/cluster_master.py
```diff
@@ -279,7 +282,7 @@ def handle_result_reported_from_slave(self, slave_url, build_id, subjob_id, payl
         build = self._all_builds_by_id[int(build_id)]
         slave = self._all_slaves_by_url[slave_url]
         # If the build has been canceled, don't work on the next subjob.
-        if not build.is_finished:  # WIP(joey): This check should be internal to the Build object.
+        if not build.is_finished or build.has_error:  # WIP(joey): This check should be internal to the Build object.
```
ethomas2: I included this `or build.has_error` logic because it was in the commit from https://github.com/Medium/ClusterRunner/commit/fd38c99155eb0373815b75230f90f09295861770, but I don't see why it's necessary. I tested the change without this line and it still seemed to work fined. It doesn't even seem related to this change.

Do you know why this is here @josephharrington, @tjlee0909? I'm gonna dig for a bit and then get rid of it if I can't find a reason to keep it here.

josephharrington: This seems like it would make builds that have failed due to an error keep executing, which I think is undesirable. I kind of suspect they added this to better deal with some other weird behavior around build cancellation (maybe builds getting stuck that are canceled before any subjobs start). I think we should leave this out, especially because we are also going to be fixing the issues with build cancellation.

tjlee0909: (Evan and I just talked about this in person)

My guess is that without it, not all of the slaves (especially the ones that didn't fail build_setup) will not be properly de-allocated from the build that they are working on unless this extra conditional statement is there.

It's important that ```scheduler.execute_next_subjob_or_free_executor(slave)``` gets called eventually on each slave so that the slave knows that it is free to work on new builds (and be added back to the free slave pool). I believe that in the event of a cancellation, those slaves will not be placed back into the free slave pool without this ```or build.has_error``` line.

tjlee0909: If we wanted to test it, we could run a build with multiple slaves (and several subjobs) where the build setup will only fail on one slave, and verify that all of the slaves end up being de-allocated properly.

josephharrington: @tjlee0909, yeah your comment sounds plausible. If that's the case I'm not opposed to leaving the has_error check in, even though it is not obvious why it is there. (I think my note about making it internal to the build object is still valid.)

Actually I suspect that `if` check is not even necessary. I bet we could just delete the `if not build.is_finished` line entirely. The point that TJ brought up about slaves not necessarily being torn down/reassigned is a good one and I bet there are some cases where this `if not build.is_finished` check prevents that.

This is definitely tied in with the larger problems with build.cancel that I'm working on, so if you prefer to hand wave it for now I'm fine with that.

test/functional/test_cluster_basic.py
```diff
@@ -104,6 +104,23 @@ def test_slave_reconnection_does_not_take_down_master(self):
...
+            'type': 'directory',
+            'config': yaml.safe_load(FAILING_SETUP_JOB.config[os.name])['FailingSetupJob'],
+            'project_directory': project_dir.name,
+        })
+
+        build_id = build_resp['build_id']
+        self.assertTrue(master.block_until_build_finished(build_id, timeout=30),
+                        'The build should finish building within the timeout.')
+        self.assertTrue(slave.get_slave_status()['slave']['is_alive'],
+                        'Slave should not die even though setup failed')
```
josephharrington: Do you want to assert that the build has an error status?

Testing the API response for the slave is not bad, but the real assertion would be that the slave is correctly marked idle on the master and is able to be assigned to other builds. We could add that to this test by asserting a subsequent build runs successfully.

ethomas2: Done

app/master/cluster_master.py
ethomas2: I wasn't able to confirm tj's theory.

I launched a master and 2 slaves on my machine, both slaves had only 1 executor. I made a setup_build that immediately failed the first time it was run, but the second time it waited 5 seconds and then exited 0. If tj's theory was correct, then I would've expected slave 2 to not be freed, but it was.

I asked huckphin why he put this in there on the medium fork https://github.com/Medium/ClusterRunner/commit/fd38c99155eb0373815b75230f90f09295861770#commitcomment-22775741. I'm just gonna merge this tonight without the `or error`. If huckphin gives me a valid reason to leave this in after I merge, i'll do a follow up change.

https://github.com/Medium/ClusterRunner/commit/fd38c99155eb0373815b75230f90f09295861770#commitcomment-22775741

tjlee0909: Ok, then something else must be performing the slave de-allocation action.

And just to make sure--when you say that the slaves have been free'd, did you mean the if you were to go to the /v1/slave endpoint, that the current_build_id of each of the slaves is null?

Also, I'm curious how you went about doing this: ``` I made a setup_build that immediately failed the first time it was run, but the second time it waited 5 seconds and then exited 0. ```

ethomas2: yeah. Visited the /v1/slave endpoint and current_build_id was null for both.

I made the following setup_build file
```python
import time
from os import system
import sys

if system('mkdir lockdir') == 0:
    sys.exit(1)
else:
    time.sleep(5)
    sys.exit(0)
```

The first time it's run the `mkdir` exits 0. The second time the `mkdir` fails.

josephharrington: @ethomas2 The bug TJ was referring to I think is a little more involved:
1. Slave A is allocated to build. Setup succeeds. First subjobs are sent to slave A.
2. Slave B is allocated to build. Setup fails. Build is marked failed (`build.has_error == True`).
3. Subjob results are reported from slave A. `handle_result_reported_from_slave()` is a no-op since `build.is_finished` is false. Slave A is stuck since `execute_next_subjob_or_free_executor` is never called.

josephharrington:
Before you merge this we should look at our internal logs and see how many setup_build failures we've been seeing. This will make our lives easier as maintainers of the cluster since slave nodes won't be going offline all the time, but since this change fails builds we're basically pushing that inconvenience down to users of the system.

test/functional/test_cluster_basic.py
```diff
@@ -104,6 +105,28 @@ def test_slave_reconnection_does_not_take_down_master(self):
...
+        self.assertTrue(master.block_until_build_finished(build_id, timeout=30),
+                        'The build should finish building within the timeout.')
+
+        slaves = master.get_slaves()['slaves']
+        build_status = master.get_build_status(build_id)['build']['status']
+        self.assertTrue(all(slave['is_alive'] for slave in slaves),
+                        'Slave should not die even though setup failed.')
+        self.assertEqual(build_status, BuildStatus.ERROR)
+
+
```
cmcginty: 😳 isn't this a pylint bug?

ethomas2: oh good catch. Not sure why pylint didn't catch it.

app/master/cluster_master.py
```diff
@@ -219,7 +219,13 @@ def _handle_setup_failure_on_slave(self, slave):
         :type slave: Slave
         """
         internal_errors.labels(ErrorType.SetupBuildFailure).inc()  # pylint: disable=no-member
-        raise BadRequestError('Setup failure handling on the master is not yet implemented.')
+        build = self.get_build(slave.current_build_id)
+        build.setup_failures += 1
```
ethomas2: I'm pretty sure this method is only ever called from tornado's thread, so I don't have to worry about locking on setup failures before incrementing.

I saw the Counter class in `app/util`. I could use that instead.

test/functional/job_configs.py
```diff
@@ -175,3 +175,30 @@ def __init__(self, config, expected_to_fail, expected_num_subjobs, expected_num_
...
+        - echo !TOKEN!
+    atomizers:
+        - TOKEN: seq 0 4
+"""
+    },
+    expected_to_fail=True,
+    expected_num_subjobs=5,
+    expected_num_atoms=5
+)
+
```
cmcginty: 😳

ethomas2: Done. Also added a snippet to my vimrc to do this automatically on save.

cmcginty: Nice! I just did that too ;-)

test/functional/test_cluster_basic.py
```diff
@@ -104,6 +105,27 @@ def test_slave_reconnection_does_not_take_down_master(self):
...
+        master = self.cluster.start_master()
+        slave = self.cluster.start_slave()
+
+        project_dir = tempfile.TemporaryDirectory()
+        build_resp = master.post_new_build({
+            'type': 'directory',
+            'config': yaml.safe_load(FAILING_SETUP_JOB.config[os.name])['FailingSetupJob'],
+            'project_directory': project_dir.name,
+        })
+
```
josephharrington: Will this test still pass if atomization fails? Should we also assert that num_atoms on the build is greater than 0?

ethomas2: Done

PR #378, opened 2017-06-28
API Version Identification

nickzuber:
The request handlers now know which version of the API to serve whenever they receive a request. 

The `version` endpoint for both the master and slave nodes have been updated to include the version of the API being served as well as the current version of ClusterRunner.

```diff
{
    "version": "0.5.512-dd16247-mod",
+   "api_version": 1,
    "child_routes": {}
}
```


boxcla:

test/functional/master/test_api_version_requests.py
```diff
@@ -0,0 +1,25 @@
...
+    @genty_dataset(
+        no_accept_header=(None, 1),
+        v1_accept_header=(1, 1),
+        v2_accept_header=(2, 2),
+        invalid_version_accept_header=(999, 1),
+    )
+    def test_api_version_with_accept_header(self, version: int, exp_version: int):
+        # TODO: Add tests for no `v1` in URI when it gets deprecated.
+        master = self.cluster.start_master()
+        version_url = master._api.url('version')
```
nickzuber: If there's a better way to write these tests just let me know, I kind of just tried to make things work based off how we set up the testing infrastructure

app/util/api_version_handler.py
```diff
@@ -0,0 +1,78 @@
...
+    @classmethod
+    def _get_default(cls, uri: str):
+        """
+        Get's the default version of the API if none was specified. This takes into account
+        the URI of the request.
+        :param uri: The URI from the request being checked.
+        :type uri: str
+        :rtype int:
+        """
+        if 'v1' in uri.split('/'):
```
josephharrington: Should we be more specific here, like 'v1' has to be the first element in the uri path?

nickzuber: The tricky part about this is it depends on the path structure being parsed looks like.

For example:

`/v1/something/path` gets split to `'', 'v1', 'something', 'path'`
and
`v1/something/path` gets split to `'v1', 'something', 'path'`

So there's some inconsistency (unless we know that the path will look a certain way every time). Splitting it then checking for `v1` makes sure we find `/v1/` somewhere in the uri instead of something like `/hasv1init/`

app/web_framework/cluster_base_handler.py
```diff
@@ -86,6 +89,10 @@ def prepare(self):
         """
         self._check_expected_session_id()
 
+        accept_header = self.request.headers.get('Accept')
+        uri = self.request.uri
+        self.api_version = APIVersionHandler.resolve_version(accept_header, uri)
```
nickzuber: That's a good idea -- do you think we should just include a header with the API version or rather try to mimic what GitHub does here? So like

```
HTTP/1.1 200 OK
Api-Version: 2
...
```

or

```
HTTP/1.1 200 OK
Clusterrunner-Media-Type: clusterrunner.v2; format=json
...
```

I personally like the first idea of just having an `Api-Version` header since the version being served is the only real unique information we're providing. Also we already do something like this with the `Session-Id` response header.

nickzuber: > Should this be in the superclass (ClusterBaseHandler)?

Yeah maybe, I wasn't too sure where exactly this should live. I was looking at the [tornado request handling docs](http://www.tornadoweb.org/en/stable/guide/structure.html#overriding-requesthandler-methods) and trying to figure out _when_ is the best part to parse the headers. It looks like we can (and should) add it to the `initialize` function of the `ClusterBaseHandler`, is this right?

app/util/api_version_handler.py
```diff
@@ -0,0 +1,78 @@
+import re
+
+
+class APIVersionHandler(object):
```
nickzuber: Sounds good 👍

app/util/api_version_handler.py
```diff
@@ -0,0 +1,78 @@
...
+        return cls._versions[0]
+
+    @classmethod
+    def get_latest(cls):
+        """
+        Returns the latest version of the API. This assumes that the latest version is the
+        version with the highest value. To ensure we get this value, the list is sorted beforehand.
+        :rtype int:
+        """
+        cls._versions.sort()
```
ethomas2: same comment. `return max(cls._versions)`

app/util/api_version_handler.py
```diff
@@ -0,0 +1,78 @@
...
+            return cls.get_latest()
+
+    @classmethod
+    def get_first(cls):
+        """
+        Returns the first version of the API. This assumes that the first version is the
+        version with the lowest value. To ensure we get this value, the list is sorted beforehand.
+        :rtype int:
+        """
+        cls._versions.sort()
```
ethomas2: I don't like the idea of mutating the `_versions` array.
You could just `return min(cls._versions)` instead.

app/util/api_version_handler.py
ethomas2:
```python
first_uri_path_component = next(part for part in uri.split('/') if part != '', None)
if 'v1' == first_uri_path_component:
  ...
```

That might be a bad variable name, but you get the idea.

app/util/api_version_handler.py
```diff
@@ -0,0 +1,78 @@
...
+    def resolve_version(cls, header: str, uri: str):
+        """
+        Gets the respective version of the API relative to the request header and URI.
+        :param header: The value of the header which to search for version type (Content-Type/Accept).
+        :type header: str
+        :param uri: The URI from the request being checked.
+        :type uri: str
+        :rtype: int
+        """
+        matches = re.match(r'.*(?:application/vnd.clusterrunner.v(\d+)\+json).*', header, re.I)
```
ethomas2: I find this logic hard to read since it's split across 5 methods (`resolve_versions`, `_get`, `_get_default`, `get_first`, `get_last`). How do you feel about condensing the first 3 of those into 1 method? (Ideally i'd like to condense the last 2 also, but they're public methods)
```python
def resolve_version(header, uri):
  header_version = None
  try:
    match =  re.match(r'.*(?:application/vnd.clusterrunner.v(\d+)\+json).*', 
                       header, re.I)
    header_version = int(match.group(0))
  except (AttributeError, ValueError):
    pass

  uri_version = next(part for part in uri.split('/') if part != '', None)
  requested_version = header_version or uri_version
  return requested_version if requested_version in cls._versions else DEFAULT_VERSION 
```

I _think_ that's the same logic as what you've laid out

app/util/api_version_handler.py
```diff
@@ -0,0 +1,78 @@
...
+
+class APIVersionHandler(object):
+    # List of all versions of the API.
+    _versions = [
+        1,
+        2
+    ]
+
+    @classmethod
+    def resolve_version(cls, header: str, uri: str):
```
ethomas2: Minor: wanna rename `header` to `accept_header` so that
1. It matches the name of the variable passed into this method
2. It's clearer which header this is

nickzuber: Maybe even `accept_header_value` since header technically refers to the "Accept" part as well and this is just the value

app/util/api_version_handler.py
nickzuber: Having `get_first` and `get_last` can be useful in other places where we want to know the first and last version of the API (might not be as relevant for `get_first`, but I know that we definitely want to use `get_last` in other spots later down the line). I don't mind the idea of condensing some of these functions though, especially the internal methods

app/util/api_version_handler.py
nickzuber: Good idea, I like this much better

nickzuber:
@ethomas2 @josephharrington Updated with feedback addressed

ethomas2:
lgtm. I'd rather joey approve though.

app/web_framework/api_version_handler.py
```diff
@@ -0,0 +1,62 @@
+import re
+
+
+class APIVersionHandler():
```
josephharrington: Omit parens. (`class APIVersionHandler:`)

app/web_framework/api_version_handler.py
```diff
@@ -0,0 +1,62 @@
...
+    OLD_VERSIONED_URI_COMPONENT = 'v1'
+    API_VERSION_HEADER_KEY = 'Api-Version'
+
+    _versions = [
+        1,
+        2
+    ]
+
+    @classmethod
+    def resolve_version(cls, accept_header_value: str, uri: str) -> int:
```
josephharrington: Looks like `accept_header_value` is optional. (`accept_header_value: Optional[str]`)

app/web_framework/api_version_handler.py
```diff
@@ -0,0 +1,62 @@
...
+        Get the respective version of the API relative to the request header and URI.
+        :param accept_header_value: The value of the header which to search for version type (Content-Type/Accept).
+        :param uri: The URI from the request being checked.
+        """
+        if accept_header_value is None:
+            return cls._get_default(uri)
+
+        version = cls._get_default(uri)
+        try:
+            matches = re.match(r'.*(?:application/vnd.clusterrunner.v(\d+)\+json).*',
```
josephharrington: Instead of `re.match` with a leading and trailing `.*`, you can just use `re.search`.
https://docs.python.org/3.4/library/re.html#search-vs-match

Also maybe use `re.IGNORECASE` instead of `re.I` for readability.

app/web_framework/cluster_base_handler.py
```diff
@@ -37,6 +38,12 @@ def initialize(self, route_node=None, **kwargs):
         :type route_node: RouteNode
         """
         self._route_node = route_node
+
+        accept_header = self.request.headers.get('Accept')
+        uri = self.request.uri
+        self.api_version = APIVersionHandler.resolve_version(accept_header, uri)
+
+        super().set_header(APIVersionHandler.API_VERSION_HEADER_KEY, self.api_version)
```
josephharrington: Is `self.set_header()` a better choice here? Usually `super()` is only needed when calling the superclass from the overriding implementation of the same method in the subclass.

test/functional/master/test_api_version_requests.py
```diff
@@ -0,0 +1,25 @@
...
+        return {header: value}
+
+    @genty_dataset(
+        no_accept_header=(None, 1),
+        v1_accept_header=(1, 1),
+        v2_accept_header=(2, 2),
+        invalid_version_accept_header=(999, 1),
+    )
+    def test_api_version_with_accept_header(self, version: int, exp_version: int):
+        # TODO: Add tests for no `v1` in URI when it gets deprecated.
```
josephharrington: I think you can omit this todo since it's a bit speculative. Notes like this have a way of sticking around forever even after they're obsolete. :)

test/functional/master/test_api_version_requests.py
```diff
@@ -0,0 +1,25 @@
...
+@genty
+class TestMasterAPIVersionRequests(BaseFunctionalTestCase):
+
+    def _build_accept_header_with_api_version(self, version: int):
+        header = 'Accept'
+        value = 'application/vnd.clusterrunner.v{}+json'.format(version)
+        return {header: value}
+
+    @genty_dataset(
+        no_accept_header=(None, 1),
```
josephharrington: I would expect the case where no header is passed in to return the *latest* API version. Is that right?

I noticed that you didn't touch the `UrlBuilder` class which looks like it embeds the `/v1/` automatically. (That's what `master._api.url('version')` is calling into below.) Was that intentional?

nickzuber: Since `v1` is in the header in this particular example we'd want to return the first version for backwards compatibility.

I adjust the `UrlBuilder` class and add more tests without having the `v1` in the URL in my other branch (which I'm gonna make a PR with after this merges and I rebase). The reason I changed the `UrlBuilder` in the next branch instead of this one is because the next branch is makes the `v1` in the URL optional -- right now in this PR we still need to have the `v1` in the URL

Issue #375, opened 2017-06-23
Proposal for versioning the API


nickzuber:
ClusterRunner API Versioning
============================

Overview
--------

The ClusterRunner API currently doesn't implement any kind of versioning, which limits us from making any kinds of enhancements or improvements to any existing API features in the future. For example, one of the changes we'd like to make is to implement pagination by default. This kind of a change would cause issues for anyone currently consuming the API, which is why we would need to release this kind of update on a new version.

There are two main approaches for handling API versioning, which are basically just two different places we can specify the version: 
 - URI
 - Media Types

In general, having the API version in the URI is a simpler approach to implement but using [media types is more academically accepted](https://en.wikipedia.org/wiki/HATEOAS).

Approach: URI
-------------

Using the URI to specify the API version is really simple and it's something that ClusterRunner currently simulates:

```
/v1/build/1/subjob/1/atom/0/console
```

With the current way we build routes in our API, and considering that we already simulate a version in our URI, making this change would be really straightforward. 

Our routes are structured like this: 

```python
RouteNode(r'parent', _SomeParentHandler, 'atom').add_children([
    RouteNode(r'child', _SomeChildHandler)
])
```

The regex that corresponds to an endpoint is constructed from the parent routes' regex. Each time we have a capturing groups in the regex, the values captured will be passed in as parameters for the handler's HTTP method. This means if we wanted to add the version to the URI, we could capture the version in the route and handle it that way.

```python
RouteNode(r'v(\d+)', _APIVersionHandler, 'atom').add_children([
    RouteNode(r'example', _ExampleHandler)
])
```
```python
class _SomeChildHandler(_ClusterMasterBaseAPIHandler):
    def get(self, api_version):
        # Do things based on api_version
        response = { ... }
        self.write(response)
```

#### Pros
 - Easy & simple to implement in our current API
 - Obvious which version is being used
 - Simple to test

#### Cons
 - Same resources can be selected with different URIs
 - Goes against HTTP specifications
 - Upgrading API can be tedious

Approach: Media Types
-----------------

Putting the API version in a media type in request headers is generally the more accepted way to handle versioning since a request URI is described as [a Uniform Resource Identifier that **identifies the resource** upon which to apply the request](https://tools.ietf.org/html/rfc2616#section-5.1.2).

Requests/reponses generally are structured something like this:

```
===>
GET /v1/build/ HTTP/1.1
Accept: application/vnd.clusterrunner.v1+json

<===
[...]

PR #379, opened 2017-06-30
Enable https for CR master and slave

shriganeshs:
This change enables https for master slave when a SSL cert and key files are provided through config file.
When SSL cert and/or key file is not provided in config file it defaults to http.
Note: The HTTPS enabled CR master does not support slaves with HTTP scheme. So when using HTTPS both master and slave should be running with HTTPS.

### DIFF
--- a/app/client/cluster_api_client.py
+++ b/app/client/cluster_api_client.py
@@ -7,6 +7,7 @@ from app.util import log, poll
 from app.util.network import Network
 from app.util.secret import Secret
 from app.util.url_builder import UrlBuilder
+from app.util.conf.configuration import Configuration
 
 
 class ClusterAPIClient(object):
@@ -15,7 +16,7 @@ class ClusterAPIClient(object):
     """
     def __init__(self, base_api_url):
         """
-        :param base_api_url: The base API url of the service (e.g., 'http://localhost:43000')
+        :param base_api_url: The base API url of the service (e.g., 'http(s)://localhost:43000')
         :type base_api_url: str
         """
         self._api = UrlBuilder(self._ensure_url_has_scheme(base_api_url))
@@ -24,13 +25,13 @@ class ClusterAPIClient(object):
 
     def _ensure_url_has_scheme(self, url):
         """
-        If url does not start with 'http' or 'https', add 'http://' to the beginning.
+        If url does not start with 'http' or 'https', add 'http://' or 'https://' at the beginning.
         :type url: str
         :rtype: str
         """
         url = url.strip()
         if not url.startswith('http'):
-            url = 'http://' + url
+            url = '{}://'.format(Configuration['protocol_scheme']) + url
         return url
 
 
--- a/app/client/cluster_api_client.py
+++ b/app/client/cluster_api_client.py
@@ -3,11 +3,11 @@ from urllib import parse
 from typing import Any, Callable, Dict, List, Optional
 
 from app.master.build import BuildStatus
+from app.util.conf.configuration import Configuration
 from app.util import log, poll
 from app.util.network import Network
 from app.util.secret import Secret
 from app.util.url_builder import UrlBuilder
-from app.util.conf.configuration import Configuration
 
 
 class ClusterAPIClient(object):
@@ -31,7 +31,7 @@ class ClusterAPIClient(object):
         """
         url = url.strip()
         if not url.startswith('http'):
-            url = '{}://'.format(Configuration['protocol_scheme']) + url
+            url = '{}://{}'.format(Configuration['protocol_scheme'], url)
         return url
 
 
--- a/app/subcommands/service_subcommand.py
+++ b/app/subcommands/service_subcommand.py
@@ -2,11 +2,13 @@ import functools
 import os
 import sys
 import tornado.ioloop
+from tornado.httpserver import HTTPServer
 
 from app.subcommands.subcommand import Subcommand
 from app.util import fs
 from app.util.safe_thread import SafeThread
 from app.util.unhandled_exception_handler import UnhandledExceptionHandler
+from app.util.conf.configuration import Configuration
 
 
 class ServiceSubcommand(Subcommand):
@@ -33,7 +35,10 @@ class ServiceSubcommand(Subcommand):
         # will raise an exception if another process is using the same port. We rely on this exception to force us to
         # exit if there are any port conflicts.
         try:
-            application.listen(port, '0.0.0.0')
+            # If SSL cert and key files are provided in configuration, ClusterRunner wil start with HTTPS protocol.
+            # Otherwise ClusterRunner will start with HTTP protocol.
+            server = HTTPServer(application, ssl_options=self._get_https_options())
+            server.listen(port, '0.0.0.0')
         except OSError:
             self._logger.error('Could not start application on port {}. Is port already in use?'.format(port))
             sys.exit(1)
@@ -54,3 +59,14 @@ class ServiceSubcommand(Subcommand):
             except OSError:
                 pass
         UnhandledExceptionHandler.singleton().add_teardown_callback(remove_pid_file)
+
+    def _get_https_options(self):
+        https_cert_file = Configuration['https_cert_file'] if 'https_cert_file' in Configuration else None
+        https_key_file = Configuration['https_key_file'] if 'https_key_file' in Configuration else None
+
+        if https_cert_file and https_key_file:
+            return {
+                'certfile': https_cert_file,
+                'keyfile': https_key_file,
+            }
+        return None
--- a/app/subcommands/service_subcommand.py
+++ b/app/subcommands/service_subcommand.py
@@ -4,11 +4,11 @@ import sys
 import tornado.ioloop
 from tornado.httpserver import HTTPServer
 
+from app.util.conf.configuration import Configuration
 from app.subcommands.subcommand import Subcommand
 from app.util import fs
 from app.util.safe_thread import SafeThread
 from app.util.unhandled_exception_handler import UnhandledExceptionHandler
-from app.util.conf.configuration import Configuration
 
 
 class ServiceSubcommand(Subcommand):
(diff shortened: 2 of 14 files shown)


boxcla:

shriganeshs:
I am working on adding functional test cases for `https`. Any thoughts on how do I store (encrypted) the cert and key files?

ethomas2:
You can generate a [self signed cert](https://www.digitalocean.com/community/tutorials/openssl-essentials-working-with-ssl-certificates-private-keys-and-csrs) and check it into the repo (with lots of warnings not to use the cert in production deployments).

It would be nice if functional tests could generate a new cert every time. But I don't know how to do that off the top of my head, and I couldn't figure it out after a minute of googling.

test/unit/util/conf/test_master_config_loader.py
```diff
@@ -0,0 +1,46 @@
...
+                             key, type(expected_stored_protocol_scheme_value), expected_stored_protocol_scheme_value,
+                             type(actual_stored_protocol_scheme_value), actual_stored_protocol_scheme_value))
+
+    def test_configure_postload_sets_protocol_scheme_to_https(self):
+        mock_config_file = self.patch('app.util.conf.base_config_loader.ConfigFile').return_value
+        mock_config_file.read_config_from_disk.return_value = {'general': {'https_cert_file': '/path/to/cert',
+                                                                           'https_key_file': '/path/to/key'},
+                                                               'master': {}
+                                                              }
+        # Unpached the patched method in BaseUnitTestCase for only this test case
```
cmcginty: You could improve this comment if you explain why the test needs to unpatch the method.

test/unit/util/conf/test_master_config_loader.py
```diff
@@ -0,0 +1,46 @@
...
+        config = Configuration.singleton()
+        config_loader = MasterConfigLoader()
+        config_loader.configure_defaults(config)
+
+        key = 'protocol_scheme'
+        expected_stored_protocol_scheme_value = 'http'
+        actual_stored_protocol_scheme_value = Configuration[key]
+
+        self.assertEqual(expected_stored_protocol_scheme_value, actual_stored_protocol_scheme_value,
+                         'The configuration value for the key "{}" was expected to be {}:{}, but was {}:{}.'.format(
```
cmcginty: I'm not sure how much value we get out of providing a custom error message. Does something bad happen if the assert fails and this message is not provided?

test/unit/util/conf/test_slave_config_loader.py
```diff
@@ -0,0 +1,46 @@
...
+                             type(actual_stored_protocol_scheme_value), actual_stored_protocol_scheme_value))
+
+    def test_configure_postload_sets_protocol_scheme_to_https(self):
+        mock_config_file = self.patch('app.util.conf.base_config_loader.ConfigFile').return_value
+        mock_config_file.read_config_from_disk.return_value = {'general': {'https_cert_file': '/path/to/cert',
+                                                                           'https_key_file': '/path/to/key'},
+                                                               'slave': {}
+                                                              }
+        # Unpached the patched method in BaseUnitTestCase for only this test case
+        # self.unpatch('app.util.conf.master_config_loader.MasterConfigLoader.load_from_config_file')
```
cmcginty: Did you mean to disable this line. If so, you can remove it.

shriganeshs: Done.

test/unit/util/conf/test_master_config_loader.py
shriganeshs: Done.

test/unit/util/conf/test_master_config_loader.py
shriganeshs: This message is definitely redundant information.
But I think the above message is providing more info regarding type of the variables. Also it describes the assertEquals failure message in more readable form.

app/client/cluster_api_client.py
```diff
@@ -24,13 +25,13 @@ def __init__(self, base_api_url):
...
         """
-        If url does not start with 'http' or 'https', add 'http://' to the beginning.
+        If url does not start with 'http' or 'https', add 'http://' or 'https://' at the beginning.
         :type url: str
         :rtype: str
         """
         url = url.strip()
         if not url.startswith('http'):
-            url = 'http://' + url
+            url = '{}://'.format(Configuration['protocol_scheme']) + url
```
tjlee0909: Super minor: you might as well use string.format here for the url as well:

```
url = '{}://{}'.format(Configuration['protocol_scheme'], url)
```

app/subcommands/service_subcommand.py
```diff
@@ -54,3 +59,14 @@ def remove_pid_file():
             except OSError:
                 pass
         UnhandledExceptionHandler.singleton().add_teardown_callback(remove_pid_file)
+
+    def _get_https_options(self):
+        https_cert_file = Configuration['https_cert_file'] if 'https_cert_file' in Configuration else None
+        https_key_file = Configuration['https_key_file'] if 'https_key_file' in Configuration else None
```
tjlee0909: Minor:

I think you could probably just remove these if 'key' in configuration checks, because you always default it to None.

app/util/conf/master_config_loader.py
```diff
@@ -38,3 +39,9 @@ def configure_postload(self, conf):
         # where to store results on the master
         conf.set('results_directory', join(base_directory, 'results', 'master'))
         conf.set('timings_directory', join(base_directory, 'timings', 'master'))  # timing data
+
+        # Set protocol scheme (by default its set to 'http')
+        https_cert_file = Configuration['https_cert_file']
+        https_key_file = Configuration['https_key_file']
+        if https_cert_file and https_key_file:
+            conf.set('protocol_scheme', 'https')
```
josephharrington: I think you could do this in the `configure_postload` section of base_config_loader. That method contains conf values derived from other conf values. It might allow you to eliminate a little bit of duplication.

app/util/conf/master_config_loader.py
```diff
@@ -1,6 +1,7 @@
 from os.path import join
 
 from app.util.conf.base_config_loader import BaseConfigLoader
+from app.util.conf.configuration import Configuration
```
nickzuber: Do we need to import this if it isn't used within the file?

test/functional/master/test_api_version_requests.py
```diff
@@ -11,10 +11,10 @@ def _build_accept_header_with_api_version(self, version: int):
         return {header: value}
 
     @genty_dataset(
-        no_accept_header=(None, 1),
+        no_accept_header=(None, 2),
         v1_accept_header=(1, 1),
         v2_accept_header=(2, 2),
-        invalid_version_accept_header=(999, 1),
+        invalid_version_accept_header=(999, 2),
```
nickzuber: How come these are changed from `1` to `2`? If there's a `v1` in the URI (which all these tests have at the moment since it's hard coded into our `UrlBuilder`, this is changed in #381 ), then we default to the first version of the API if no other valid version is specified in the request header. See [the design doc](https://github.com/box/ClusterRunner/issues/375) for reference

app/util/conf/slave_config_loader.py
```diff
@@ -1,6 +1,7 @@
 from os.path import join
 
 from app.util.conf.base_config_loader import BaseConfigLoader
+from app.util.conf.configuration import Configuration
```
nickzuber: Same as https://github.com/box/ClusterRunner/pull/379#discussion_r125797187

app/util/network.py
```diff
@@ -38,7 +38,8 @@ def reset_session(self):
         if self._session:
             self._session.close()  # Close any pooled connections held by the previous session.
         self._session = requests.Session()
-        self._session.mount('http://', HTTPAdapter(pool_connections=self._poolsize, pool_maxsize=self._poolsize))
+        self._session.mount('{}://'.format(Configuration['protocol_scheme']),
```
wjdhollow: Should `protocol_scheme` be a constant? It's used in several places across multiple files.

app/util/conf/master_config_loader.py
shriganeshs: @nickzuber (for all 3 comments) - I am/was working on the changes. I have fixed these and just pushed the changes.

app/util/conf/master_config_loader.py
shriganeshs: Done.

app/util/network.py
josephharrington: You could make that same point about all configuration value reads. If we were to do something like that I'd rather it be consistent across the app and not just for one value.

app/util/conf/base_config_loader.py
```diff
@@ -116,6 +117,10 @@ def configure_postload(self, conf):
         conf.set('log_file', join(log_dir, conf.get('log_filename')))
         conf.set('eventlog_file', join(log_dir, conf.get('eventlog_filename')))
 
+        # Set protocol scheme (by default its set to 'http')
```
josephharrington: `its` -> `it's`
:neckbeard:

shriganeshs:
We have decided not to write functional tests for HTTPS (i.e. starting master and slave in HTTPS mode).
I tried to create a functional test for starting master in HTTPS mode using programmatically created self signed certificate (e.g. [link](https://skippylovesmalorie.wordpress.com/2010/02/12/how-to-generate-a-self-signed-certificate-using-pyopenssl/)). For testing the HTTPS end-points the self signed certificate needs to be verified. The certificate verification through functional tests will need changes the the CR code (especially around `request` calls in `app/util/network.py`) to either provide a certificate to verify or ignore the verification (by passing `verify=False` to `get / post` etc. requests). The issue here is that this modification to ignore the verification will not really complete the functional testing. Also it's not worth to verify the actual SSL encrypted traffic. So we decided to skip functional tests for HTTPS.

Issue #380, opened 2017-07-05
POC for handling different versions in API


nickzuber:
This is a proof of concept for an idea I had for how we want to adjust our current api to allow for different versions. My main goal here is to make adding new http methods that align with different api versions to be easy and clean, and to have it so we need to change how we write http methods as little as necessary.

Design
-------

The idea is that we have a few decorators that assign the appropriate supported versions to their respective http method. 

### RequestHandler HTTP Methods Decorator

There is another decorator `versioned_method` which wraps a `RequestHandler` http method. This decorator takes three arguments: `method`, which describes the http method that the decorator is wrapping, `start` which specifies the earliest version of the API that this method supports (defaults to the first version), and `end` which specifies the last version of the API that this method supports (defaults to the latest version).

For each http method wrapper, we set two attributes to the function: a range of versions that this particular method supports that will be used if the api is using one of those versions, and a tag describing which type of http method this particular function is used for (`post`, `get`, `put`, etc.).

```python
def versioned_method(method, start=FIRST_API_VERSION, end=LATEST_API_VERSION):
    """
    Decorate HTTP method with a range of API versions that this method is supported in.
    :param method: Type of HTTP method that this function is describing.
    :param start: First supported version in API for this method (inclusive)
    :param end: Last supported version in API for this method (inclusive)
    """
    def versioning_decorator(func):
        def wrapper(self, *args, **kwargs):
            print('\tCalling wrapped function from `versioned_method`')
            return func(self, *args, **kwargs)

        versions = [v for v in API_VERSIONS if v >= start and v <= end]
        setattr(wrapper, SUPPORTED_API_VERSIONS_METHOD_KEY, versions)
        setattr(wrapper, SUPPORTED_API_VERSIONS_HTTP_TYPE_KEY, HTTP_METHODS[method])
        return wrapper

    return versioning_decorator
```

### RequestHandler Class Decorator

There is a decorator `versioned_handler` which wraps a `RequestHandler`. Its job is to figure out which http method to use based on the api version whenever a request is made. We identify the api version that is requested from the `RequestHander` and also identify all the methods in the `RequestHandler` that are marked as versioned from the `versioned_method` decorator.

[...]

PR #381, opened 2017-07-06
Make the v1 in URIs optional

nickzuber:
The `v1` in the URIs is now flagged as an optional route. This means that it may or may not be included in any given endpoint. For example:

```
http://localhost:43000/build

↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ 

{
    "builds": [],
    "child_routes": {
        "build": "/build/[build_id]"
    }
}
```
and
```
http://localhost:43000/v1/build

↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ 

{
    "builds": [],
    "child_routes": {
        "build": "/v1/build/[build_id]"
    }
}
```

What's new
------------

### Getting child routes

Also there's a new method in the `RouteNode` class called `get_child_routes` which returns the children of a given `RouteNode`. The difference from using this method and simply calling `route_node.children` is in `get_child_routes`, if a child route is flagged as optional and it _does not_ show up in the request URI, then we consider its children to be part of the current route's children. 

```python
def get_child_routes(self, uri=None):
    """
    Get a list of all children routes for this route. If one of the child routes is marked optional
    and it doesn't appear within the requested URI, we return its children instead of it.
    :param uri: The URI of the request.
    """
    if uri is None:
        return self.children

    uri_parts = uri.split('/')
    child_routes = list()
    for child in self.children:
        if child.optional and child.name() not in uri_parts:
            child_routes += child.get_child_routes(uri)
        else:
            child_routes.append(child)
    return child_routes
```

Example of this affecting the `/v1/` endpoint:

```
http://localhost:43000/v1

↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ 

{
    "child_routes": {
        "slaves": "/v1/slave",
        "eventlog": "/v1/eventlog",
        "version": "/v1/version",
        "metrics": "/v1/metrics",
        "queue": "/v1/queue",
        "builds": "/v1/build"
    },
    "master": {
        "slaves": [],
        "status": "Master service is up."
    }
}
```

and without `v1` in the URI

```
http://localhost:43000/

↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ 

{
    "child_routes": {
        "slaves": "/slave",
        "eventlog": "/eventlog",
        "version": "/version",
        "metrics": "/metrics",
        "queue": "/queue",
        "builds": "/build"
    }
}
```

### Route templates

Route templates now handle routes that are marked as optional. If a route is marked as optional and _does not_ appear in the request URI, then it is not considered part of the route template.

```python
def route_template(self, uri=None):
    """
    The generic form of this route, for display in the API 'child routes'
    :rtype: str
    """
    uri_parts = uri.split('/') if uri else []
    ancestor_names = []
    for ancestor in list(reversed(self.ancestors())):
        # If a route is marked optional, only include it in route template if its already in URI
        if ancestor.optional and ancestor.name() not in uri_parts:
            continue
        ancestor_names.append(ancestor.name().rstrip('/'))

    return '/'.join(ancestor_names + [self.name()])
```

### Building URLs

The `UrlBuilder` now has an optional parameter that dictates whether or not the URL being build will include the `v1`. By default it _does_ include `v1` for backwards compatibility.

```python
def url(self, *args, use_versioned_url=True)
```

Notes
------

The `RouteHandler` associated with the `v1` endpoint writes some information about the master service:

```python
def get(self):
    response = {
        'master': self._cluster_master.api_representation(),
    }
    self.write(response)
```

Since this route is optional, we should think about include this behavior to the root `RequestHandler` as well or something. Right now the only way to get this information is to hit the `/v1/` endpoint, which shouldn't be the case if we don't necessarily want people going out of their way to use `v1`.


boxcla:

josephharrington:
I made some comments on your #380 that might affect this change. 😬

nickzuber:
This approach has been abandoned. Now we don't want to make the v1 in URI optional, but rather have routes with v1 return the first API version and routes without the most recent unless specified.

Routes with v1 in the URI are unique to _only_ the v1 version of the API -- a request for a different version of the API while having v1 in the URI will result in a 404

PR #384, opened 2017-07-11
Convert subjob start SafeThread to standard Thread

### LINKED ISSUE #337: Failure to start a subjob on a slave can take down the master
The code to send a subjob to a slave uses a SafeThread, which is intended to only be used for application critical threads. This means if the request fails (for example due to a connection error or network timeout) then the master service could go down.

This is relatively unlikely since a subjob is only sent when the build first starts (in which case we've very recently confirmed that the slave is responsive), or when the slave itself sends results for a previous subjob (in which case it's unlikely that the slave goes down in the small window between sending a result and receiving new work).

josephharrington:
Sending a subjob to a slave is done asynchronously. Even though it's bad if it fails, it shouldn't bring down the entire master service.

This is a quick fix because we've run into this issue multiple times. I'll follow up with a change to clean this up and improve the behavior in the error case.

Fixes #337.

### DIFF
--- a/app/master/slave.py
+++ b/app/master/slave.py
@@ -1,10 +1,11 @@
+from threading import Thread
+
 import requests
 
 from app.master.build import Build
 from app.util import analytics, log
 from app.util.counter import Counter
 from app.util.network import Network
-from app.util.safe_thread import SafeThread
 from app.util.secret import Secret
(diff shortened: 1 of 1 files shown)


boxcla:

PR #386, opened 2017-07-11
Refactor subjob start threading

josephharrington:
This is a follow-up to my previous change to move subjob start off of `SafeThread` to avoid taking down the master service. This change further improves the behavior when a subjob start call fails.

- Instead of the `Slave` class starting a new thread to kick off a subjob, `ClusterMaster` starts the thread. This makes more sense since threads are only started to perform asynchronous actions in response to API requests and the `Slave` class shouldn't need to worry about that.
- In the case where a new subjob is not started and the slave is instead torn down, previously the teardown API call happened synchronously with the slave's API call to report a result (which is bad). The teardown API call is now asynchronous.
- Record error metric when subjob fails to start.
- Requeue subjobs when subjob fails to start because of a dead slave or an error on the slave side. Previously subjobs were only requeued if the slave was explicitly put in shutdown mode.

### DIFF
--- a/app/common/metrics.py
+++ b/app/common/metrics.py
@@ -31,6 +31,7 @@ class ErrorType(str, Enum):
     NetworkRequestFailure = 'NetworkRequestFailure'
     PostBuildFailure = 'PostBuildFailure'
     SetupBuildFailure = 'SetupBuildFailure'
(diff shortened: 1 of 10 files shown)


boxcla:

Issue #282, opened 2015-12-30
Console output API does not match style of other APIs


josephharrington:
Here's an example of the console output API response:

```json
GET "/v1/build/1/subjob/0/atom/0/console"
{
	"total_num_lines": 1,
	"child_routes": {},
	"content": "hello dummy\n",
	"offset_line": 0,
	"num_lines": 1
}
```

Note that the "child_routes" entry is combined with the resource data. The pattern set by all the other APIs is to nest the resource data one level deeper. The main benefit of this is that response metadata ("child_routes" is the only example of this right now) can be separated from the data of the requested resource.

Examples of other API responses that show the proper pattern:

```json
GET "/v1/build/1/subjob/0/atom/0"
{
	"atom": {
		"subjob_id": 0,
		"id": 0,
		"exit_code": 0,
		"expected_time": 0.01,
		"command_string": "export ATOM=\"hai\";",
		"state": "COMPLETED",
		"actual_time": 0.03
	},
	"child_routes": {
		"console": "/v1/build/[build_id]/subjob/[subjob_id]/atom/[atom_id]/console"
	}
}
```

```json
GET "/v1/slave/0"
{
	"slave": {
		"url": "jharrington.local:43001",
		"num_executors_in_use": 0,
		"is_in_shutdown_mode": false,
		"current_build_id": null,
		"id": 1,
		"num_executors": 1,
		"is_alive": true
	},
	"child_routes": {
		"shutdown": "/v1/slave/[slave_id]/shutdown"
	}
}
```

This would be an easy change to make if we ignored backwards compatibility, but since we haven't yet done any API versioning there will be some overhead in fixing this the right way.

cmcginty:
The level of effort depends on how close we are to supporting API versioning.

josephharrington:
API versioning is in now. Effort for this would be pretty small.

PR #388, opened 2017-07-12
API Version Handling

nickzuber:
As discussed [here](https://github.com/box/ClusterRunner/issues/380#issuecomment-314570421), the ClusterRunner API can now handle different versioned requests.

The first bit to realize is that we can have situations where multiple routes have the same URI path, but different RequestHandlers based on the requested API version. Something like

```python
api_v1 = RouteNode(r'version', _VersionHandler)
api_v2 = RouteNode(r'version', _VersionHandlerForV2)
```

There are a few steps we take to handle this:

## Separate the routes by their version

We separate the routes by their version then add them as children under the root route. The `add_children` has been adjusted to take an option parameter that assigns all of the children an API version. So something like

```python
api_v1 = [RouteNode( ... ).add_children( ... )]
api_v2 = [RouteNode( ... ).add_children( ... )]

root_route = RouteNode(r'/', _RootHandler)
root_route.add_children(api_v1, version=1)
root_route.add_children(api_v2, version=2)
```

## Combine routes with same regex

Next we create a dictionary that maps unique regex patterns to all of the routes that have that share that same regex pattern.

```python
route_map = {}
for route in all_route_nodes:
    route_path = route.regex()
    route_map.setdefault(route_path, []).append(route)
```

## Join routes under RequestHandlerWrapper 

Now we want to create a dictionary that maps a unique regex pattern to a RouteNodeWrapper and RequestHandlerWrapper tuple. The RouteNodeWrapper takes a list of version to RouteNode tuples, and the RequestHandlerWrapper takes a list of version to RequestHandler tuples. Each of these wrappers create a map linking the version to the respective route or handler, so when a request is made to one of the URIs (one of the regex paths), we know which route and which handler to use.

We want to separate these two (routes and handlers) because of the way Tornado deals with them. We need to instantiate a given request with the correct handler as the request is being made, whereas we just need to select the correct route in ClusterBaseHandler for whenever we want to build child routes. 

```python
route_handler_map = {}
for regex in route_map:
    routes = route_map[regex]
    # Create two lists of tuples of version to handler and version to route for each unique regex pattern
    version_handler_tuples = []
    version_route_tuples = []
    for route in routes:
        version_handler_tuples.append((route.version, route.handler))
        version_route_tuples.append((route.version, route))
    route_handler_map[regex] = (RouteNodeWrapper(version_route_tuples), RequestHandlerWrapper(version_handler_tuples))
```

## Pass off handlers to Tornado

Tornado handlers take the form of a `tuple(regex, handler, parameters)`. So since we have a mapping of regex to RouteNodeWrapper and RequestHandlerWrapper tuples, this is pretty easy to construct.

```python
prepared_routes = []
for regex in route_handler_map:
    route_wrapper, handler_wrapper = route_handler_map[regex]
    prepared_routes.append((regex, handler_wrapper, dict(default_params, route_node_wrapper=route_wrapper)))

return prepared_routes
```

## The RequestHandlerWrapper class

The RequestHandlerWrapper class is a callable class that, when invoked, parses through the Tornado arguments to find the requested URI and headers, figures out what API version was requested, and instantiates the proper handler based on that requested API version. 

```python
class RequestHandlerWrapper:
    """
    Wrapper class around a collection of RequestHandlers that returns the correct handler relative
    to the version of the API that was requested. This class stores a map of RequestHandlers using
    their assigned API versions as keys.
    :param handlers_map: Dictionary of mappings for API versions to handlers
    """
    def __init__(self, version_handlers_map):
        self.version_handlers_map = dict(version_handlers_map)

    def __call__(self, *args, **kwargs):
        """
        This method gets called when RequestHandlerWrapper is instantiated by Tornado.
        Parse the request arguments for the accept header and URI and derive the requested
        API version. Return the handler which corresponds to that API version so then that
        handler is instantiated by Tornado.
        """
        _, request = args
        accept_header = request.headers['Accept']
        uri = request.uri
        version = APIVersionHandler.resolve_version(accept_header, uri)
        return self.get_handler(version)(*args, **kwargs)

    def _get_default_handler(self):
        """
        If there is a handler which has no assigned API version, return that handler, if not
        then return the ItemNotFoundHandler to signify that the route requested is not compatible
        with the API version requested.
        """
        route = get_item_not_found_route_with_message('Requested route is not compatible with requested API version.')
        return self.version_handlers_map.get(None, route.handler)

    def get_handler(self, version: int):
        return self.version_handlers_map.get(version, self._get_default_handler())
``` 

If the requested API version doesn't match any of the handlers' respective API versions, we throw a 404 by returning a handler that's only purpose in life is to raise an `ItemNotFoundError` exception.

## The RouteNodeWrapper class

This class is much simpler. Given a collection of versions and their respective routes, you can select any route based on a requested version. If a version is selected that does not have a route associated with it (an example would be trying to hit the `/v1/build` endpoint while requesting version 2 of the API) then we return a default RouteNode that simply throws a 404 when instantiated. Every handler needs to be associated with a RouteNode, so that's why we want to return a ItemNotFound type of RouteNode here.

```python
class RouteNodeWrapper:
    def __init__(self, version_routes_map):
        self.version_routes_map = dict(version_routes_map)

    def get_route(self, version: int):
        return self.version_routes_map.get(version, self._get_default_route())

    def _get_default_route(self):
        return self.version_routes_map.get(None, get_item_not_found_route_with_message(
            'Requested route is not compatible with requested API version.'))
```

### DIFF
--- a/app/util/fs.py
+++ b/app/util/fs.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import tarfile
 import tempfile
+import zipfile
 
 from app.util.process_utils import Popen_with_delayed_expansion
 
@@ -131,9 +132,17 @@ def zip_directory(target_dir: str, archive_filename: str) -> str:
     """
     # Create the archive in a temp location and then move it to the target dir.
     # (Otherwise the resulting archive will include an extra zero-byte file.)
-    tmp_path = shutil.make_archive(tempfile.mktemp(), 'zip', target_dir)
     target_path = os.path.join(target_dir, archive_filename)
-    shutil.move(tmp_path, target_path)
+    with tempfile.TemporaryDirectory() as temp_dirpath:
+        tmp_zip_filename = os.path.join(temp_dirpath, 'clusterrunner_tmp__' + archive_filename)
+        with zipfile.ZipFile(tmp_zip_filename, 'w', compression=zipfile.ZIP_DEFLATED) as zf:
+            for dirpath, dirnames, filenames in os.walk(target_dir):
+                for filename in filenames:
+                    path = os.path.normpath(os.path.join(dirpath, filename))
+                    if os.path.isfile(path):
+                        relpath = os.path.relpath(path, target_dir)
+                        zf.write(path, relpath)
+        shutil.move(tmp_zip_filename, target_path)
     return target_path
 
 


boxcla:

app/web_framework/cluster_application.py
```diff
@@ -1,5 +1,29 @@
 import tornado.web
 
+from app.web_framework.api_version_handler import APIVersionHandler
+from app.web_framework.cluster_base_handler import Throws404Handler
+
+
+class RequestHandlerWrapper:
```
nickzuber: This class probably shouldn't live here.

app/web_framework/cluster_base_handler.py
```diff
@@ -192,3 +192,12 @@ def _is_request_origin_allowed(self, request_origin):
...
         return False
+
+
+class Throws404Handler(ClusterBaseAPIHandler):
+    def initialize(self, route_node=None, **kwargs):
+        self._logger = log.get_logger(__name__)
+        super().initialize(route_node)
+
+    def get(self):
+        raise ItemNotFoundError
```
nickzuber: This returns a response that looks like:

```json
{
    "error": "",
    "child_routes": {}
}
```

How do we want this to look / is there an proper way to customize this?

app/web_framework/cluster_master_application.py
```diff
@@ -30,38 +30,71 @@ def __init__(self, cluster_master):
...
-                        ]),
-                        RouteNode(r'shutdown', _SlavesShutdownHandler, 'shutdown')
-                    ]),
-                    RouteNode(r'eventlog', _EventlogHandler)
+                    RouteNode(r'shutdown', _SlavesShutdownHandler, 'shutdown')
+                ]),
+                RouteNode(r'eventlog', _EventlogHandler)
+            ])]
+
+        api_v2 = [
```
nickzuber: Like mentioned above, this is mainly for testing purposes right now to prove that it works. We'd probably want to remove this from this PR if we want to merge it at some point

app/web_framework/cluster_master_application.py
```diff
@@ -30,38 +30,71 @@ def __init__(self, cluster_master):
...
+            RouteNode(r'slave', _SlavesHandler, 'slaves').add_children([
+                RouteNode(r'(\d+)', _SlaveHandler, 'slave').add_children([
+                    RouteNode(r'shutdown', _SlaveShutdownHandler, 'shutdown')
+                ]),
+                RouteNode(r'shutdown', _SlavesShutdownHandler, 'shutdown')
+            ]),
+            RouteNode(r'eventlog', _EventlogHandler)]
+
+        root_route = RouteNode(r'/', _RootHandler)
+        root_route.add_children(api_v1, version=1)
```
josephharrington: What did you decide to do about adding v1 routes under both '/' and '/v1/'? That seems like the main immediate use case for the RequestHandlerWrapper idea. (From what I can tell there are no colliding route regexes in the current RouteNodes.)

app/web_framework/route_node.py
```diff
@@ -60,8 +61,26 @@ def add_children(self, child_nodes):
...
+        """
+        return [child for child in self.children if child.version == version]
+
+    def assign_version_to_all_children(self, version: int):
+        """
+        Recursively assigns an API version to the current child and all of its direct children.
+        """
+        for node in self.children:
+            node.version = version
+            node.assign_version_to_all_children(version)
```
josephharrington: Super minor, but I think if you make this 
```python
self.version = version
for node in self.children:
    node.assign_version_to_all_children(version)
```
then you won't need the `node.version = version` line 65 above. :neckbeard:

app/web_framework/cluster_application.py
```diff
@@ -15,6 +52,33 @@ def get_all_handlers(root_route, default_params):
...
+        # Map regex to a list of route and RequestHandlerWrapper tuples.
+        # The RequestHandlerWrapper knows all the RequstHandlers and what API version those handlers work
+        # with for a given route, so when a route is hit by the API, we know which RequestHandler to use.
+        route_handler_map = {}
+        for regex in route_map:
+            for route in route_map[regex]:
+                if regex not in route_handler_map:
+                    route_handler_map[regex] = []
+                # Create a dict that maps version to handler
+                handlers_map = dict([(route.version, route.handler) for route in route_map[regex]])
```
josephharrington: I have been staring at this for awhile so I might be looking at this wrong, but I think there might be a bug here. This dict is created inside this `for route in route_map[regex]` loop, but it is doing the same loop as a comprehension. It seems like this would create copies of the same handlers_map. Am I seeing that right?

I have a feeling the logic in this method can be simplified. Maybe the behavior would be more obvious with some test cases.

app/web_framework/cluster_master_application.py
nickzuber: That's a good point, it doesn't really get used unless we explicitly make a request for v2 so it won't hurt to leave it in. Good opportunity to use this to try out #282 and changing `/builds/[build_id]` to `/build/[build_id]` with this too

app/web_framework/cluster_application.py
```diff
@@ -15,6 +52,33 @@ def get_all_handlers(root_route, default_params):
         :rtype: list [tuple (str, tornado.web.RequestHandler, dict)]
         """
         all_route_nodes = [root_route] + root_route.descendants()
+
+        # Map regex to all the routes that have the same regex pattern.
+        route_map = {}
+        for route in all_route_nodes:
+            route_path = route.regex()
+            if route_path not in route_map:
+                route_map[route_path] = []
```
nickzuber: What kind of issues? The regex mappings are used to resolve any ambiguity for any routes that has the same path, so if the regex for v1 and v2 are different (`'artifacts.zip'` and `'artifacts.zip$'`) then the routes should just be able to handle that normally, right?

app/web_framework/cluster_application.py
nickzuber: Ah yeah nice catch, we can move this outside of this inner loop to make things a little more efficient. 

So it's a little tricky what's going on here (and if you see any areas for improvement let me know), but what's happening is we're taking our map of regex to list of routes that share the same regex, and we want to create a map of regex to list of route & RequestHandlerWrapper tuples.

Each RequestHandlerWrapper takes a mapping of version to handler, where each version/handler item comes from the the routes which belong to that regex pattern.

So it looks like:

```javascript
{
    regex: [
        (RouteNode1, RequestHandlerWrapper({
            1: _HandlerForV1 // RequestHandler from RouteNode1
            2: _HandlerForV2 // RequestHandler from RouteNode2
        }),
        (RouteNode2, RequestHandlerWrapper({
            1: _HandlerForV1 // RequestHandler from RouteNode1
            2: _HandlerForV2 // RequestHandler from RouteNode2
        }),
    ],
    regex: [ /* ... */ ],
}
```

So the logic of creating this would look like

```python
route_handler_map = {}
for regex in route_map:
    # Create a dict that maps version to handler for each unique regex pattern
    handlers_map = dict([(route.version, route.handler) for route in route_map[regex]])
    for route in route_map[regex]:
        if regex not in route_handler_map:
            route_handler_map[regex] = []
        route_handler_map[regex].append((route, RequestHandlerWrapper(handlers_map)))
```

So we kind of loop through the route list in the `route_map` -- the first time being to create the RequestHandlerWrapper for each of the RouteNodes for that regex pattern, and again to assign each of those RouteNodes that RequestHandlerWrapper.

app/web_framework/cluster_base_handler.py
```diff
@@ -192,3 +192,14 @@ def _is_request_origin_allowed(self, request_origin):
...
                            'Not setting Access-Control-Allow-Origin header.', request_origin, allowed_origins_regex)
         return False
+
+
+class ItemNotFoundHandler(ClusterBaseAPIHandler):
+    """
+    ClusterBaseAPIHandler that raises an ItemNotFoundError when it's instantiated.
+    """
+    def initialize(self, route_node=None, **kwargs):
+        super().initialize(route_node)
```
nickzuber: Yeah so it's because without defining `initialize` here, we get passed in `cluster_master` or `cluster_slave` as a keyword argument since this handler is being instantiated by the master/slave service. 

Then we call `super().initialize(route_node)` because all of our handlers are expected to be assigned to a route node (without it we get this error):

```
AttributeError: 'ItemNotFoundHandler' object has no attribute '_route_node'
```

app/web_framework/cluster_base_handler.py
```diff
@@ -192,3 +192,14 @@ def _is_request_origin_allowed(self, request_origin):
...
+
+    def get(self):
+        raise ItemNotFoundError
```
nickzuber: Since users will only get this message when they try to make an API call to an endpoint that doesn't support the requested API version, should we give a more detailed error message explaining that's the case? Something like

```
raise ItemNotFoundError('Requested route is not compatible with requested API version.')
```

nickzuber: Alternatively we can have a wrapper function that takes an error message and returns the `ItemNotFoundHandler` with that error.

```python
def get_ItemNotFoundHandler_with_message(msg):
    class ItemNotFoundHandler(ClusterBaseAPIHandler):
        """
        ClusterBaseAPIHandler that raises an ItemNotFoundError when it's instantiated.
        """
        def initialize(self, route_node=None, **kwargs):
            """
            
            """
            super().initialize(route_node)

        def get(self):
            raise ItemNotFoundError(msg)

    return ItemNotFoundHandler
```

app/web_framework/cluster_master_application.py
nickzuber: Do we want to add v1 routes under `/`? And yeah as of now there are no collisions but there would be in the future (or if we want to add v1 routes under `/` as well)

app/util/url_builder.py
```diff
@@ -19,13 +19,17 @@ def __init__(self, service_address, api_version=API_VERSION_1):
         self._api_version = api_version
         self._scheme = '{}://'.format(Configuration['protocol_scheme'])
 
-    def url(self, *args):
+    def url(self, *args, use_versioned_url=True):
```
nickzuber: We could do that and I think that would be nice to have, but won't it make your https://github.com/box/ClusterRunner/pull/388#discussion_r127317485 here moot?

app/web_framework/route_node.py
nickzuber: Nice catch 😎

app/web_framework/cluster_application.py
nickzuber: Also important note: I found a flaw in this logic (the handlers need to be assigned to api version but so do the routes). Adjusting this now and will push an update soon

app/web_framework/cluster_master_application.py
```diff
@@ -30,58 +31,100 @@ def __init__(self, cluster_master):
...
-    pass
+    def get(self):
+        response = {
+            'master': self._cluster_master.api_representation(),
+        }
+        self.write(response)
 
 
-class _APIVersionOneHandler(_ClusterMasterBaseAPIHandler):
+class _APIVersionOneHandler(_RootHandler):
```
nickzuber: I kind of assumed we wanted to do this -- basically the master info was only available on the `v1` endpoint so I moved it to the root endpoint and I have the `v1` endpoint mimic it for backwards compatibility

josephharrington:
I already mentioned this to you, but just to record it here: after looking through this again I think this might be getting a bit more complex than we currently need. Let's chat a bit today about if we can simplify this or split it up into more digestible chunks. :)

PR #389, opened 2017-07-12
Rewrite zip archive creation

### LINKED ISSUE #383: Build result zips sometimes contain a different build's results
Sometimes the artifacts.zip for a build will be completely incorrect, containing the results of a entirely different build.

Since console output logs for completed subjobs are pulled from this zip archive, sometimes the API will also return console output for a different build. Depending on what a build outputs this can be very hard to detect. Often the only sign something is wrong is that an atom is recorded as failing but the console output looks successful.

The root cause of this is a race condition in Python's `shutil.make_archive`. That method is actually not at all thread-safe, even though that caveat is undocumented. There is [an issue](https://bugs.python.org/issue30511) filed against Python related to this.

We should be able to work around this by using the `zipfile` module directly instead of going through `shutil`.

josephharrington:
This reimplements the functionality of `shutil.make_archive` without the need for calls to `os.chdir`. This is thread-safe, and fixes the issues we've been seeing with finished build showing console output for other builds.

It will also hopefully fix the occasional zip file creation issues we've been seeing.

Fixes #383, and potentially #339.

### DIFF
--- a/app/util/fs.py
+++ b/app/util/fs.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import tarfile
 import tempfile
+import zipfile
 
 from app.util.process_utils import Popen_with_delayed_expansion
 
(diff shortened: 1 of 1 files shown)


boxcla:

app/util/fs.py
```diff
@@ -131,9 +132,17 @@ def zip_directory(target_dir: str, archive_filename: str) -> str:
     """
     # Create the archive in a temp location and then move it to the target dir.
     # (Otherwise the resulting archive will include an extra zero-byte file.)
-    tmp_path = shutil.make_archive(tempfile.mktemp(), 'zip', target_dir)
     target_path = os.path.join(target_dir, archive_filename)
-    shutil.move(tmp_path, target_path)
+    with tempfile.TemporaryDirectory() as temp_dirpath:
```
gcurtis: You could also use [`tempfile.NamedTemporaryFile`][1] with `delete=False` to create the zip file directly and skip creating the temp directory.

[1]: https://docs.python.org/3.4/library/tempfile.html#tempfile.NamedTemporaryFile

josephharrington: @gcurtis Yeah, definitely. I thought about doing that but went this route instead just because it felt slightly more awkward to create a NamedTemporaryFile, immediately close it, and then reopen it with zipfile.

PR #390, opened 2017-07-12
Refactor shutdown functional test to be more stable

josephharrington:
This test has been flaky on Windows because of parallel subjobs all trying to copy the same file. This refactors the test config to be simpler and not rely on a copy. The test asserts on both the project directory file list and file contents at the end of the test, so the copy is not really necessary.


boxcla:

PR #392, opened 2017-07-18
Add v2 API routes

nickzuber:
Add v2 API routes to both the slave and master services. These routes are assuming that there are no collisions with the route URIs, which is safe to assume for now since all v1 API paths are under `/v1/...` and all v2 API paths are just under `/...`

## What's changed

The `RouteNode` class now has a method that returns all child routes that have the same version requested. So when we get children now we don't want to use `self._route_node.children` but instead use `self._route_node.get_children(self.api_version)` to make sure we only get the children routes that have the correct version.

We want to use this for the root route, since it has both v1 and v2 routes as children. This way, when a request to v1 of the API is made on the route now, we will only show the child routes that are also from v1, etc.

### DIFF
--- a/app/web_framework/cluster_slave_application.py
+++ b/app/web_framework/cluster_slave_application.py
@@ -18,33 +18,59 @@ class ClusterSlaveApplication(ClusterApplication):
         }
         # The routes are described using a tree structure.  This is a better representation of a path than a flat list
         #  of strings and allows us to inspect children/parents of a node to generate 'child routes'
-        root_route = \
-            RouteNode(r'/', _RootHandler).add_children([
-                RouteNode(r'v1', _APIVersionOneHandler).add_children([
-                    RouteNode(r'version', _VersionHandler),
-                    RouteNode(r'build', _BuildsHandler, 'builds').add_children([
-                        RouteNode(r'(\d+)', _BuildHandler, 'build').add_children([
-                            RouteNode(r'setup', _BuildSetupHandler),
-                            RouteNode(r'teardown', _TeardownHandler),
-                            RouteNode(r'subjob', _SubjobsHandler, 'subjobs').add_children([
-                                RouteNode(r'(\d+)', _SubjobHandler, 'subjob').add_children([
-                                    RouteNode(r'atom', _AtomsHandler, 'atoms').add_children([
-                                        RouteNode(r'(\d+)', _AtomHandler).add_children([
-                                            RouteNode(r'console', _AtomConsoleHandler)
-                                        ])
+        api_v1 = [
+            RouteNode(r'v1', _APIVersionOneHandler).add_children([
+                RouteNode(r'version', _VersionHandler),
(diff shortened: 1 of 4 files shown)


boxcla:

app/web_framework/cluster_slave_application.py
```diff
@@ -18,33 +18,59 @@ def __init__(self, cluster_slave):
...
+                RouteNode(r'executor', _ExecutorsHandler, 'executors').add_children([
+                    RouteNode(r'(\d+)', _ExecutorHandler, 'executor')
+                ]),
+                RouteNode(r'eventlog', _EventlogHandler),
+                RouteNode(r'kill', _KillHandler)
+            ])]
+
+        api_v2 = [
+            RouteNode(r'version', _VersionHandler),
+            RouteNode(r'build', _BuildsHandler, 'builds').add_children([
```
josephharrington: Want to change these to plural too?

app/web_framework/route_node.py
```diff
@@ -5,21 +5,23 @@ class RouteNode(object):
...
-    def __init__(self, regex_part, handler, label=None):
+    def __init__(self, regex_part, handler, label=None, version=None):
         """
         :param regex_part: To generate the regex the web framework will use to match this route, we combine a set of
         regex_parts: The regex_part in this node and the regex_parts in all its ancestor nodes.
         :type regex_part: str
         :param handler: The handler class that will be instantiated by the web framework when this route is hit.
         :type handler: type
         :param label: A human-friendly label for the objects returned by this route, ie "builds"
+        :param version: The API version assigned to this route
```
josephharrington: Can you add a type hint (might as well convert the docstring hints to annotation style while you're here).

app/web_framework/route_node.py
```diff
@@ -51,17 +53,35 @@ def name(self):
                     return '[{}]'.format(get_params[-1])
         return self.regex_part
 
-    def add_children(self, child_nodes):
+    def add_children(self, child_nodes, version=None):
         """
         Build the tree structure by adding child RouteNodes to this RouteNode.  Can be chained.
         :type child_nodes: list[RouteNode]
+        :param version: The API version assigned to all children routes
```
josephharrington: Same comment about type hints. :kissing_heart:

app/web_framework/route_node.py
```diff
@@ -51,17 +51,35 @@ def name(self):
                     return '[{}]'.format(get_params[-1])
         return self.regex_part
 
-    def add_children(self, child_nodes):
+    def add_children(self, child_nodes: list, version: Optional[int]=None):
```
josephharrington: This can be `child_nodes: List['RouteNode']` (with RouteNode in quotes).
https://www.python.org/dev/peps/pep-0484/#forward-references

Also, you can remove the docstring type hints for any annotation types you've added.

josephharrington: Oh and same for the :rtype: hint.

Issue #393, opened 2017-07-19
Pagination Proposal


nickzuber:
The first part that we need to figure out for pagination is how to get the requested range of results. This depends on how we want to paginate -- by actual pages or just limiting results.

We can do this with query parameters `start` and `amount`, or alternatively `page` and `per_page`. If we go through the route of having actual pages then we'd want a way to show `next` and `prev` pages probably just through some field in the response. 

If we went with the `start` and `amount` approach, we could provide an HTTP handler method with a decorator to introduce the respective `start` and `amount` values.

```python
DEFAULT_PAGINATION_START = 0
DEFAULT_PAGINATION_AMOUNT = 50

def paginated(func):
    def function_with_pagination(self, *args, **kwargs):
        start = int(self.get_query_argument('start', DEFAULT_PAGINATION_START, True))
        amount = int(self.get_query_argument('amount', DEFAULT_PAGINATION_AMOUNT, True))
        pagination_range = { 'start': start, 'amount': amount }
        return func(self, *args, pagination_range=pagination_range, **kwargs)

    return function_with_pagination
```

Then we could use this like

```python
@paginated
def get(self, pagination_range):
    start, amount = pagination_range.values()
    response = {
        'builds': [build.api_representation() for build in self._cluster_master.builds(start, amount)]
    }
    self.write(response)
```

And for example, to limit the builds we could do

```python
def builds(self, start=None, amount=None):
    """
    Returns a list of all builds
    :rtype: list[Build]
    """
    num_builds = len(self._all_builds_by_id)
    start = start or 0
    amount = amount or num_builds
    return [self._all_builds_by_id[key] for key in islice(self._all_builds_by_id, start, min((start + amount), num_builds))]
```

ethomas2:
I like the `start` and `amount` way you suggested. 

I don't understand the other suggestion. The "actual pages" way where there are `prev` and `next`. Can you expand on that? Ie what would this look like to the user, and what will implementation look like.

nickzuber:
@ethomas2 Yeah so an example of this is github's API. I think they do things a bit differently but the idea is that you split the results into pages with x amount per page. 

So you could have something like `builds?per_page=10&page=1` and the response could maybe some additional field linking to the next and previous pages. This could give you something like

```javascript
{
    "child_routes": { 
        // ...
    },
    "builds": {
        // ...
    },
    "navigation": {
        "prev": null,
        "next": "builds?per_page=10&page=2"
    }
}
```

[...]

PR #394, opened 2017-07-20
Pagination params to base handler

nickzuber:
Add pagination parameters to the ClusterBaseHandler. This allows for any RequestHandler to get query params called `start` and `amount`.

In the v2 API, we also paginate the `builds` endpoint for the master service. 

The current defaults are `start=0` and `amount=50`, do we want to adjust these? Also do we want to adjust the naming of `start` and `amount`?

Closes https://github.com/box/ClusterRunner/issues/393


boxcla:

app/master/cluster_master.py
```diff
@@ -62,12 +63,15 @@ def api_representation(self):
             'slaves': slaves_representation,
         }
 
-    def builds(self):
+    def builds(self, start=None, amount=None):
```
josephharrington: I mentioned this in your proposal, but just to push a bit harder 😛: I think I prefer "limit/offset" to "amount/start" just because there is precedent in other contexts, so might help understanding. I'm not sure they're objectively better words, but since "limit/offset" is used frequently in database contexts I feel like someone reading this code is more likely to be able to interpret those words quickly. 

Also, want to add type hints throughout? :smile_cat:

app/web_framework/cluster_master_application.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, cluster_master):
         api_v2 = [
             RouteNode(r'metrics', _MetricsHandler),
             RouteNode(r'version', _VersionHandler),
-            RouteNode(r'builds', _BuildsHandler).add_children([
+            RouteNode(r'builds', _BuildsPaginatedHandler).add_children([
```
josephharrington: Let's call this something like `_v2BuildsHandler`. Just because pagination is the only difference now doesn't mean it will always be the only difference.

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +66,11 @@ def _handle_request_exception(self, ex):
         self.set_status(status_code)
         self.finish()
 
+    def _get_pagination_params(self):
+        start = int(self.get_query_argument('start', self.DEFAULT_PAGINATION_START, True))
+        amount = int(self.get_query_argument('amount', self.DEFAULT_PAGINATION_AMOUNT, True))
```
josephharrington: We should probably reject negative amounts.

app/master/cluster_master.py
```diff
@@ -62,12 +63,15 @@ def api_representation(self):
...
         """
```
nadeemahmad: The docs need to be updated.

app/master/cluster_master.py
```diff
@@ -62,12 +63,15 @@ def api_representation(self):
...
         Returns a list of all builds
         :rtype: list[Build]
         """
-        return self._all_builds_by_id.values()
+        num_builds = len(self._all_builds_by_id)
+        start = start or 0
+        amount = amount or num_builds
+        return [self._all_builds_by_id[key] for key in islice(self._all_builds_by_id, start, min((start + amount), num_builds))]
```
nadeemahmad: This is slightly difficult to read, I think this should separated out into a few lines, maybe something like this (untested):
```python
total_builds = len(self._all_builds_by_id)
page = page or 0
size = size or total_builds
stop_index = min((page + size), total_builds))
builds_on_request_page = islice(self._all_builds_by_id, start=page, stop=stop_index)
return builds_on_request_page.values()
```

nadeemahmad:
I actually like @josephharrington's suggestion around `limit` and `offset`, so ignore my comments around `page` and `size`.

app/web_framework/cluster_master_application.py
nickzuber: Good point -- is `V2[OldHandlerName]` the convention we want to go with?

josephharrington: I think that's fine. It would help make it clear just from looking at the routenodes which endpoints had changed for v2 and which were still falling back to old logic.

app/master/cluster_master.py
```diff
@@ -62,12 +65,24 @@ def api_representation(self):
             'slaves': slaves_representation,
         }
 
-    def builds(self):
+    def builds(self, offset: int=None, limit: int=None) -> List['Build']:
         """
-        Returns a list of all builds
-        :rtype: list[Build]
+        Returns a list of all builds. Both offset and limit are assumed to be nonnegative at this point
+        since the corrections or negative offsets and limits are done when the query parameters are first initialized.
```
nickzuber: Should we do another check here for negative entries? Technically this method should only ever get limit and offset arguments from the pagination method, which will not return negative numbers, so another check here might be overkill

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +66,14 @@ def _handle_request_exception(self, ex):
         self.set_status(status_code)
         self.finish()
 
+    def _get_pagination_params(self):
```
nickzuber: There isn't really a good way to write tests for this method unless we have this data stored in some response. The problem is that this method gets the offset and limit from the query arguments in a request, and then does things internally with them, so there isn't really a way for us to intercept those values without explicitly having a handler that does that.

One idea is we could split this up into two methods like:

```python
def get_pagination_params(self) -> Tuple[int, int]:
    offset = self.get_query_argument('offset', self.DEFAULT_OFFSET, True)
    limit = self.get_query_argument('limit', self.DEFAULT_LIMIT, True)
    return self._validate_arguments(int(offset), int(limit))

def _validate_arguments(self, offset: int, limit: int) -> Tuple[int, int]:
    # No negative offsets or limits are accepted
    non_negative_offset = max(offset, 0)
    non_negative_limit = max(offset, 0)
    return non_negative_limit, min(non_negative_limit, self.MAX_LIMIT)
```

Then we can easily unit test `_validate_arguments`. I think this would be a good idea for testing and also wouldn't hurt to separate the methods anyways -- any opinions before I do this on whether or not we should?

ethomas2:
Is it possible for a single build to ever be such a large json blob that we'd want to somehow limit how much of that single build is returned? Ie I ask for `offset=X&limit=1` but the response is still very large? It seems to me that
1. That's not possible
2. Even if it was there wouldn't really be anything we could do about it

but I just wanted to make sure that's not something we're overlooking.

app/master/cluster_master.py
ethomas2: 1. I like the idea of adding an extra check, especially since it's so easy to do.
`offset = max(offset or 0, 0)`
In my perfect world every method would be callable with every combination of arguments that its type signature permits. Since this method's type signature allows for negative arguments, it should behave sanely when negative arguments are passed in. (Sometimes obeying this philosophy is more trouble than it's worth, but in this case it's a one liner)

2. iirc @josephharrington suggested allowing you to pass in negative offsets in order to get the most recent builds. So if I passed in `offset=0&limit=-1` it would give me the most recent build. I liked that suggestion. Could we rename the variables to something like `start` and `range` and then do
```python
start = start % num_builds
end = min(start + range,  num_builds)
end = end % num_builds
start, end = min(start, end), max(start, end)
return [self._all_builds_by_id[key] for key in islice(self._all_build_by_id, start, end)]
```

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +66,14 @@ def _handle_request_exception(self, ex):
...
+        offset = self.get_query_argument('offset', self.DEFAULT_OFFSET, True)
+        limit = self.get_query_argument('limit', self.DEFAULT_LIMIT, True)
+        # No negative offsets or limits are accepted
+        non_negative_offset = max(int(offset), 0)
```
ethomas2: 1. This is sort've a stupid edge case, but `offset` and `limit` are user provided strings, which means `int` might raise a value error and return a 500 to the user. Catch the error and return a 400 instead.

2. Might as well just rename `non_negative_offset` and `non_negative_limit` to `offset` and `limit`

3. If you go with my earlier suggestion (shamelessly stolen from Joey) about allowing `offset` and `limit` to be negative, you'll want to get rid of `max` logic from this method (allow negative numbers to be passed through to the `builds` method.

app/web_framework/cluster_base_handler.py
ethomas2: Imo this method shouldn't be tested in isolation in the first place. I don't really care if this method is behaving properly, I care if the pagination functionality is behaving properly. This method is just a means to an end, so don't test it. (It's also private, which is an indication that it's not a good thing to test).

A better test would be a functional test that passes various `offset`/`limit` combinations to cluster runner and verifies that we get back the set of builds we expect. Don't test _functions_, test _functionality_.

I know that flies in the face of the concept of unit testing, and some people on the team disagree with me. So do what you want.

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +69,24 @@ def _handle_request_exception(self, ex):
...
+        offset = self.get_query_argument('offset', pagination_constants['DEFAULT_OFFSET'], True)
+        limit = self.get_query_argument('limit', pagination_constants['DEFAULT_LIMIT'], True)
+        return self._validate_arguments(int(offset), int(limit))
+
+    @staticmethod
+    def _validate_arguments(offset: int, limit: int) -> Tuple[int, int]:  # pylint: disable=invalid-sequence-index
+        offset = offset or pagination_constants['DEFAULT_OFFSET']
+        limit = limit or pagination_constants['DEFAULT_LIMIT']
+        # Negative offsets or limits are not allowed
+        non_negative_offset = max(offset, 0)
```
ethomas2: I made this comment on an earlier commit, making it again to make sure you see it since github folded the old comment away with the older commit.

1. offset and limit are user provided strings, which means `int(some_string)` might raise a value error and return a 500 to the user. Catch the error and return a 400 instead.

2. Might as well just rename non_negative_offset and non_negative_limit to offset and limit.

3. If you go with my earlier suggestion (shamelessly stolen from Joey) about allowing offset and limit to be negative, you'll want to get rid of max logic from this method (allow negative numbers to be passed through to the `builds` method.)

app/master/cluster_master.py
nickzuber: Is the negative entries something we want to do now? This might be better left for a separate PR

app/web_framework/cluster_base_handler.py
nickzuber: For (1) what do you think the best way for handling that would be? Use a try/catch and if it throws a value exception use the default instead (not sure if there's a more elegant way to do this in python besides a try/catch)?

And for (2) I think the reason I explicitly named them non_negative_x was because of (3) if we wanted to add that in the future then it's really clear where we disallow negatives. We could totally just rename the variables I don't think it'd matter much if that's what we wanna do

ethomas2: 1.  I'd suggest doing
```python
try:
  a = int(string_variable)
  b = int(other_string_variable)
catch ValueError as ex:
  raise BadRequestError from ex
```

For context, `ClusterBaseHandler` has a map of known exceptions to status codes. When an exception bubbles up to `ClusterBaseHandler` it calls `_handle_request_exception` (a method overwritten from tornado) and translates the exception to the relevant status_code. `BadRequestError` is one of the exceptions in this map.

2 and 3. 
I think the negative offset stuff does belong in this change. I like separating a single PR into multiple PR's when one of the following is true
(a) a single PR is too big
(b) a single PR is taking too long to implement
(c) a single PR feels like it's doing "too many things", ie 2 logically different pieces of functionality in the same commit.

I think none of those criteria apply here, so I don't see why we'd separate it. I'm not necessarily opposed to separating it, it doesn't hurt anything, I just don't see a reason to.

nickzuber: Sounds good to me. 

For clarification, on the negative offset, I'm a little confused on how we want it to behave. Python's negative indexing works because it's start/end but here we have start/amount, so things are a bit different. If we have a `offset=-5` and `limit=20` request for a 100 build result, do we return the builds from 95 to 100 then 0 to 15? Or something like builds 95 to end of builds or limit (in this case 95 to 100)?

app/master/cluster_master.py
josephharrington: I'm fine with leaving negative offsets for later. @ethomas2 I think in your example it should be `offset=-1&limit=1`. I think it only makes sense to support negative offsets. Negative limits (especially combined with negative offsets) feels unintuitive to me.

app/web_framework/cluster_base_handler.py
josephharrington: Just to repeat my above comment, let's only support negative values for `offset` (not `limit`). And I think it's fine to save that for a later change. It seems like most of the conversation/comments/questions on this PR so far have been around the negative offset support, so that's a good sign that it would be good to split out -- get everything merged that everyone agrees on and save the highly-discussed piece for a separate change. This somewhat falls under Evan's point (b) above about splitting things out that are taking too long to implement, but I'd extend that to "too long to *complete* due to implementation or review" including a best effort to try to split out things that you expect to be more contentious or debatable. (And btw, I'll take any disagreement with this comment as proof of my point. 😝)

Also, with regards to your last question about wrap-around -- I don't think it makes sense to support wrap-around. In your example of `offset=-5&limit=20` I think it makes the most sense to only return the last five items. I think that's intuitive.

A clearer example of why wrap-around is not good is if we only had 2 builds total and asked for `offset=0&limit=8` would we return a list of 8 builds which are just the only two builds repeated four times (1, 2, 1, 2, 1, 2, 1, 2)? Bonkers!

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +69,24 @@ def _handle_request_exception(self, ex):
         self.set_status(status_code)
         self.finish()
 
+    # For methods that annotate with `Tuple`, use `disable=invalid-sequence-index`
+    # This bug is fixed in https://github.com/PyCQA/astroid/commit/563031aaf13a44adc8db4f8d0ab8020d550aae00
+    # More information on the issue in https://github.com/PyCQA/pylint/issues/1212
```
josephharrington: Is this fixed in the most recent version of pylint? If so, can you create a new ClusterRunner issue to update pylint (and reference this line so we remember to remove it)?

test/unit/master/test_pagination.py
```diff
@@ -0,0 +1,118 @@
...
+from app.master.build_request import BuildRequest
+from app.master.cluster_master import ClusterMaster
+from app.slave.cluster_slave import SlaveState
+from app.util.exceptions import BadRequestError, ItemNotFoundError
+from app.web_framework.cluster_base_handler import ClusterBaseHandler, pagination_constants
+from test.framework.base_unit_test_case import BaseUnitTestCase
+
+
+@genty
+class TestClusterMaster(BaseUnitTestCase):
```
josephharrington: The test class name should match the name of the file. And on that note, why a new file instead of adding these to the existing test_cluster_master.py? (Edit: see my comment below; I think I understand why you might have split these out.)

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +69,24 @@ def _handle_request_exception(self, ex):
...
+
+    def get_pagination_params(self) -> Tuple[int, int]:  # pylint: disable=invalid-sequence-index
+        offset = self.get_query_argument('offset', pagination_constants['DEFAULT_OFFSET'], True)
+        limit = self.get_query_argument('limit', pagination_constants['DEFAULT_LIMIT'], True)
+        return self._validate_arguments(int(offset), int(limit))
+
+    @staticmethod
+    def _validate_arguments(offset: int, limit: int) -> Tuple[int, int]:  # pylint: disable=invalid-sequence-index
```
josephharrington: I think the name `_validate_arguments` is too vague since this is specific to pagination.

test/unit/master/test_pagination.py
```diff
@@ -0,0 +1,118 @@
...
+            expected_first_build_id: int,
+            expected_last_build_id: int,
+            ):
+        master = ClusterMaster()
+        for build_id in range(1, 501):
+            build_mock = Mock(spec=Build)
+            build_mock.build_id = build_id
+            master._all_builds_by_id[build_id] = build_mock
+
+        offset, limit = ClusterBaseHandler._validate_arguments(offset, limit)
```
josephharrington: This seems out of place. I think a test for ClusterBaseHandler should test logic in ClusterBaseHandler, and a test for ClusterMaster should test logic in ClusterMaster. (Is this why you split out these tests?)

With this call here, this becomes some sort of integration test between ClusterBaseHandler and ClusterMaster, but it ends up not being a realistic integration test since it picks and chooses related methods from each class to call (which may or may not reflect reality).

I think if you wanted to test this logic in ClusterBaseHandler we should add a test class specifically for ClusterBaseHandler that would call _V2BuildsHandler.get directly. Alternatively, a functional test could test the API end-to-end which wouldn't be a bad idea either.

app/web_framework/cluster_base_handler.py
```diff
@@ -11,6 +12,12 @@
 from app.util.session_id import SessionId
 from app.web_framework.api_version_handler import APIVersionHandler
 
+pagination_constants = {
+    'DEFAULT_OFFSET': 0,
+    'DEFAULT_LIMIT': 20,
+    'MAX_LIMIT': 200,
+}
```
josephharrington: I would probably make these separate constants or, if you really want to explicitly group them, an `enum`. A dict feels a bit too free-form.

test/unit/master/test_pagination.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+    def test_pagination_request_with_query_params(
+            self,
+            offset: Optional[int],
+            limit: Optional[int],
+            expected_first_build_id: int,
+            expected_last_build_id: int,
+            ):
+        master = ClusterMaster()
+        for build_id in range(1, 501):
```
josephharrington: Is 501 chosen in relation to the default pagination parameters? If so, maybe express the value in terms of those parameters? Also, are there any edge cases you'd want to test where the actual number of builds is less than MAX_LIMIT? 

An alternative to this pattern is to put the pagination parameters in Configuration (base_config_loader.py) and then set them in test setup to values more convenient for a testing situation. It's not a huge deal but I would guess that we don't need to create 501 build mocks to exercise the logic you are trying to test. :)

app/master/cluster_master.py
ethomas2: I'm fine with leaving negative offsets for a later change, but I just spent a long time thinking about negative offsets though and the exact spec I want, so i'm gonna write it down here. I don't really care if we implement it now, later or never. Fwiw @josephharrington, I do think we can make negative limits a well defined and intuitive concept. Maybe it's not useful, but I think it makes sense.

The mental model I have is this. Lets say you have an array of `N` builds. You can think of any `offset`/`limit` combo as defining a "window" that you place over the array of builds. The return value is just whatever part of the array you can see through that window. If `offset` is negative you mod it by `N`, if it's positive you just leave it alone.

```python
offset = offset if offset > 0 else offset % N  # Don't mod by N if offset is positive so we can handle cases where offset is off the right edge of the array
if limit >= 0:
    start, end = offset, min(offset + limit, N)
else:
    start, end = max(offset + limit, 0), min(offset, N)
return builds[start:end]
```

Some cases to test. (`N` is 100 for all cases). I do agree that the last 2 cases are a little weird, but they're well defined and they fit my model.

| offset        | limit           | return value  | case
|:------------- |:-------------| :-----| :-----|
| 5/-95 | 20      |    [5,24] | positive limit
| -5/95 | -20 | [75,95] | negative limit
| -5/95      | 20 | [95, 99] | limit overflow right
| 5/-95 | -20 | [0,4] | limit overflow left
| 100| 1 | [] | offset off right edge, positive limit
| 101| -1 | [] | offset off right edge. Window does not overlap array
| 100| -2 | [99] | offset off right edge. Window does overlap array

app/master/cluster_master.py
```diff
@@ -62,12 +65,24 @@ def api_representation(self):
...
+        :param limit: The number of builds requested
         """
-        return self._all_builds_by_id.values()
+        num_builds = len(self._all_builds_by_id)
+        offset = offset or 0
+        # Don't return more than the MAX_LIMIT even if we have more builds than that
+        limit = limit if limit is not None else min(num_builds, pagination_constants['MAX_LIMIT'])
+
+        # Reset the offset to 0 if it is out of range
+        starting_index = offset if offset < num_builds else 0
```
ethomas2: I would expect if `offset` is off the right edge of the array and `limit` puts the end even further off the right edge of the array for you to get back the empty array. 
E.g if there are 5 builds, and I get `offset=5,limit=1` I would expect to get back the empty array. Similar to how in python this code evaluates to the empty array.
 ```
arr = [0, 1, 2, 3, 4]
print(arr[5:6])
```

your logic would reset starting_index to 0 though, and return the first element of the array.

File: app/master/cluster_master.py (hunk 6 above)
Comment (josephharrington) [inline app/master/cluster_master.py]: 🙀

File: app/web_framework/cluster_base_handler.py (hunk 9 above)
Comment (nickzuber) [inline app/web_framework/cluster_base_handler.py]: @josephharrington Hah yeah good point, that would get really confusing. I was thinking more like `offset=-1&limit=20` for 10 builds would give builds 9 to 10, then 1 to 9 etc so the wrap wouldn't repeat builds, but yeah like you said it'd get really confusing and I don't think this would be a good idea after I think about it

**Comment by nickzuber** [comment]
@josephharrington My comments on "negative limits" was more regarding that as an illegal input, not really something we'd want to support. Like if someone requested a limit of -20 we'd just default to 0 since negative limit I think we can assume returning no results would make the most sense.

File: test/unit/master/test_pagination.py (hunk 15 above)
Comment (nickzuber) [inline test/unit/master/test_pagination.py]: The 501 here is arbitrary -- I was generating 500 mock builds to test with ids of 1 to 500 inclusively (maybe a comment explaining it would be better?)

File: app/web_framework/cluster_base_handler.py (hunk 10 above)
Comment (nickzuber) [inline app/web_framework/cluster_base_handler.py]: 👍 https://github.com/box/ClusterRunner/issues/395

File: test/unit/master/test_pagination.py (hunk 15 above)
Comment (nickzuber) [inline test/unit/master/test_pagination.py]: Also I don't understand how moving the pagination default values to the config would affect the structure of these particular tests? I do think moving the default pagination values to the config might make more sense, I was wondering where a good place for them to live would be. With them being in the config they'd be more customizable which would be nice.

Are you suggesting these particular tests shouldn't pass any parameters to `builds` but rather only test the default values?

File: test/unit/master/test_pagination.py (hunk 13 above)
Comment (nickzuber) [inline test/unit/master/test_pagination.py]: Ah I see, so each test file is dedicated to testing a certain file in clusterrunner rather than a test file will be dedicated to testing a certain concept in clusterrunner? That's why I was creating new files for these pagination tests, I didn't know that's how we organized things. 

If that's the case, then yeah it'd make more sense to move these `builds` tests to `unit/master/test_cluster_master` and the pagination tests to `unit/web_framework/test_cluster_base_handler`

The reason I used `_validate_arguments` here is because the arguments that get passed into the `builds` endpoint always go through this function first. I think it's fine and probably makes more sense to remove this function from this test and just give "validated" entries to the tests instead.

File: app/web_framework/cluster_base_handler.py (hunk 14 above)
Comment (nickzuber) [inline app/web_framework/cluster_base_handler.py]: Would we rather change this to an enum or move this to the config?

Comment (josephharrington) [inline app/web_framework/cluster_base_handler.py]: I'd personally probably vote config, mostly because there are already established testing patterns/framework around making sure config gets reset for every test. (For example see `Configuration['timings_directory']`.)

File: test/unit/master/test_pagination.py (hunk 15 above)
Comment (josephharrington) [inline test/unit/master/test_pagination.py]: Ah yeah, I meant that if those default values are in config then it's more straightforward to set them to whatever you want during these tests. For example you could set max_limit to 10 (or some low value) that still allowed you to still test the logic without needing to create hundreds of mock objects.

A nice thing about setting config values in tests is that the base test case makes sure they get reset for every test. It kind of works like `patch` that way. Without that automatic reset it can be easy to forget to set a value back to its original value, which could potentially influence other tests that run after yours.

File: test/unit/master/test_pagination.py (hunk 13 above)
Comment (josephharrington) [inline test/unit/master/test_pagination.py]: Yeah, unit+integration tests are a bit different than functional tests in that way. Functional tests are usually oriented around concepts of functionality since by their nature they cannot test specific code classes, but unit+integration should be more oriented around the structure of the code. That makes unit+integration tests a bit easier to organize since it's clearer where to put certain tests and have them share common setup etc., but it also helps us be very conscious of the entry points of our tests (i.e., "every test in file X tests some method in class Y").

File: app/master/cluster_master.py (hunk 16 above)
Comment (josephharrington) [inline app/master/cluster_master.py]: Agreed!

Comment (nickzuber) [inline app/master/cluster_master.py]: Yeah after messing around with it a bit I noticed it seemed really weird to have `offset=10` for something like 5 builds to return all 5. Makes much more sense to return nothing 👍

File: app/master/cluster_master.py [hunk 17]
```diff
@@ -62,12 +65,24 @@ def api_representation(self):
...
+        ending_index = min((starting_index + limit), num_builds)
+
+        requested_builds = islice(self._all_builds_by_id, starting_index, ending_index)
```
ethomas2: Writing this here for visibility. I don't think anyone should actually fix this right now, I just want to make sure this "bug" is known.

I noticed that 
1. `_all_builds_by_id` is an OrderedDict, not an array
2. `islice` returns builds things from an OrderedDict in the order they were added to the dict, not in the numerical order of the keys

This means if somehow builds got added to the OrderedDict out of numerical order, `islice` would return the builds out of order.

For example, say we got 3 requests all at the same time. We created build ids 0,1,2 for them but we added them to the OrderedDict in the order 2,1,0. Then if I made a request to clusterrunner with `offset=0,limit=1` I would get build 2 instead of build 0.

That being said, this behavior is actually impossible right now. The "bug" I described relies on builds being added to the dict in a different order than they are created, but the current code is setup in a way that that can never happen. Builds are created and then immediately added to the dict in `handle_request_for_new_build`. This happens on tornado's single thread, so there's no possibility of threads interleaving operations in a way that causes this bug.

So ... nothing to do right now. Just wanted to write this down. At some point someone should maybe change `_all_builds_by_id` to be an array though, which would fix this potential bug.

app/web_framework/cluster_base_handler.py
nickzuber: So are you saying when we do something like

```python
def setUp(self):
    super().setUp()
    Configuration['timings_directory'] = self._TIMINGS_DIR_SYS_PATH
```

`Configuration['timings_directory']` is changed in this particular class test but it gets reset for all the others?

app/master/cluster_master.py
nickzuber: That's a good catch -- alternatively we could create a range of build ids from the offset/limit and explicitly retrieve those from the OrderedDict. 

Like `offset=5&limit=20` would imply we want builds 6 to 25, so we could just fetch all the builds with an id from 6 to 25 from the OrderedDict (we could then just use a regular dict)

ethomas2: Oh wow. Yeah that's super easy to do, we should just do that now.
`return [self._all_builds_by_id[id] for id in range(starting_index, ending_index)]`

nickzuber: @ethomas2 Agreed 👍 makes much more sense intuitively and is cleaner than islice IMO

app/master/cluster_master.py
```diff
@@ -62,12 +62,34 @@ def api_representation(self):
...
+        limit = max(limit, 0)
+
+        # If limit is set higher than the number of builds/max limit, reduce limit
+        limit = highest_amount if limit > highest_amount else limit
+
+        # Requested offset too high should yield no results
+        if offset > num_builds:
+            return []
+
+        first_build_id = offset + 1
```
ethomas2: Why are we adding 1 here?
Does that mean if I ask for `offset=10,limit=1` i'll actually get the 11th build?
Also, does that mean it's impossible to ever get the first build in the array?

app/master/cluster_master.py
```diff
@@ -62,12 +62,34 @@ def api_representation(self):
...
+        last_build_id = min((offset + limit), num_builds)
+
+        # Add +1 to last_build_id because range is exclusive
+        return [self.get_build(uid) for uid in range(first_build_id, last_build_id + 1)]
```
ethomas2: Also confused why we're adding 1 to last_build_id here. It seems like this would make it so that if I pass in `limit=1`, I actually get 2 builds back

nickzuber: `range`'s params are inclusive and exclusive respectively, so if we want to get builds 1 through 10 we'd do `range(1, 11)` which would produce `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`. Maybe there's a clearer way to do this?

app/master/cluster_master.py
nickzuber: We start our builds with an id of 1 and so on, so with an offset of 0 we'd expect to get builds starting with 1. Maybe I should add a comment here about that?

ethomas2: Ah, so it's because build id's are 1 indexed and offset is 0 indexed. I feel like that might be confusing. Personally if I queried for `offset=10,limit=1` i'd expect to get build_id 10 back. I feel like it might make sense to either start 0 indexing the builds or 1 index the offset. If we 1 indexed the offset, maybe it would make more sense to name it `start`? (I think that might make more sense anyway).

I can't decide if the current behavior is actually confusing, or if i'm just looking for something to complain about. Opinion @josephharrington ?

app/master/cluster_master.py
ethomas2: Ah, my mistake. There's not a clearer way of doing this, I was just bad at reading code.

I thought `last_build_id = first_build_id + limit`
but it's actually `last_build_id = offset + limit`

app/master/cluster_master.py
nickzuber: I think if we know that the builds start at 1 then something like `offset=6` giving builds starting at 7 makes sense. This is why I originally was leaning towards start/amount for names so its a little more obvious where you're starting and ending. 

I'm personally fine with either one and don't have any strong opinions in either way

app/master/cluster_master.py
```diff
@@ -62,12 +65,24 @@ def api_representation(self):
             'slaves': slaves_representation,
         }
 
-    def builds(self):
+    def builds(self, offset: int=None, limit: int=None) -> List['Build']:
```
wjdhollow: Minor. I don't believe you need the quotes around Build for List generics.

app/master/cluster_master.py
```diff
@@ -28,7 +28,7 @@ def __init__(self):
         self._logger = get_logger(__name__)
         self._master_results_path = Configuration['results_directory']
         self._all_slaves_by_url = {}
-        self._all_builds_by_id = OrderedDict()
+        self._all_builds_by_id = dict()
```
josephharrington: I would just write a `{}` instead of `dict()` for consistency with the line above. :p

app/master/cluster_master.py
```diff
@@ -62,12 +62,34 @@ def api_representation(self):
...
+
+        offset = offset if offset is not None else Configuration['pagination_offset']
+        limit = limit if limit is not None else Configuration['pagination_limit']
+
+        # Remove any negative values
+        offset = max(offset, 0)
+        limit = max(limit, 0)
+
+        # If limit is set higher than the number of builds/max limit, reduce limit
+        limit = highest_amount if limit > highest_amount else limit
```
josephharrington: Is this line the same as `limit = min(limit, highest_amount)`?

Also, what do you think about renaming `highest_amount` to `max_limit`?

app/master/cluster_master.py
josephharrington: Why use `get_build()` here instead of getting directly from `_all_builds_by_id`? It seems like code in this method is trying to make sure that every build in the computed range is a valid build id, so if there's a mistake in that logic it is more of a server error than a client error (so we don't need the potential ItemNotFoundError raised from `get_build()`.

Also, isn't this implementation a breaking change for v1 API's /builds endpoint? It seems like it would make that API stop returning the entire list of builds.

app/master/cluster_master.py
josephharrington: Yeah I think it's fine as-is. If we wanted to give builds random strings of characters for an id (or if we wanted to count by 2s or 4s like some dbs do) then we should be able to do that. The paging offset happens to be closely related to the build id in the current indexing scheme but that should be considered a coincidence.

nickzuber: @josephharrington Yeah after I realized this I think I want to switch back to using islice instead of getting builds by their id. The pagination isn't really for getting specific builds from their id but rather getting builds from the collection

ethomas2: I think using `islice` is extra dangerous now that you've changed the dict from an OrderedDict to a regular dict. `islice` returns things in iteration order, but iteration order is not defined for a dict. I expect that `islice` might return different results from machine to machine, or even from hour to hour on the same machine.

I say if you go back to `islice`, at least change the dict back to an OrderedDict. That way at least the iteration order is defined, and we can make guarantees about it. Why did you change it to a normal dict in the first place?

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +64,31 @@ def _handle_request_exception(self, ex):
...
+        limit = self.get_query_argument('limit', Configuration['pagination_limit'], True)
+        try:
+            offset = int(offset)
+            limit = int(limit)
+        except ValueError as ex:
+            raise BadRequestError from ex
+        return self._validate_pagination_params(offset, limit)
+
+    @staticmethod
+    def _validate_pagination_params(offset: int, limit: int) -> Tuple[int, int]:  # pylint: disable=invalid-sequence-index
```
wjdhollow: This doesn't look like it validates anything. In fact, it looks like it does exactly the opposite of validation because it silently fails over to default values. I suggest renaming it to something like _get_pagination_params_or_safe_defaults. Something to indicate that it won't raise an error and that it will return sensible defaults. 

Honestly, if you just combined the two methods, that would would work too.

nickzuber:
@josephharrington Ahh nice catch, yeah we don't want to add any default offset/limit logic in the builds endpoint. If they're not set, we should set them to 0 and `num_builds` respectively so we return everything (for v1). In v2 these will always be set when passed off to the builds method.

app/util/conf/base_config_loader.py
```diff
@@ -105,6 +105,11 @@ def configure_defaults(self, conf):
         conf.set('https_cert_file', None)
         conf.set('https_key_file', None)
 
+        # Default values for pagination requests
+        conf.set('pagination_offset', 0)
+        conf.set('pagination_limit', 20)
+        conf.set('pagination_max_limit', 200)
```
nickzuber: https://github.com/box/ClusterRunner/issues/393#issuecomment-316870581

app/web_framework/cluster_base_handler.py
nickzuber: You're right, this could do with a better name.

I intentionally separated the two methods into Tornado retrieving the values and then sanitizing the values. This way we can test the sanitizing the values function since there isn't a good way to test the former.

app/web_framework/cluster_base_handler.py
```diff
@@ -62,6 +64,31 @@ def _handle_request_exception(self, ex):
...
+        offset = offset if offset is not None else Configuration['pagination_offset']
+        limit = limit if limit is not None else Configuration['pagination_limit']
+
+        # A negative limit will give no results
+        limit = max(limit, 0)
+        offset = max(offset, 0)
+
```
nickzuber: There's a comment buried here somewhere that mentions we might as well repeat the negative logic here since we probably want to do something different with negative offsets in the future.

app/web_framework/cluster_master_application.py
```diff
@@ -263,6 +263,15 @@ def get(self):
         self.write(response)
 
 
+class _V2BuildsHandler(_BuildsHandler):
+    def get(self):
+        offset, limit = self.get_pagination_params()
```
nickzuber: You mean once in the handler and again in the builds?

app/util/conf/base_config_loader.py
```diff
@@ -105,6 +105,11 @@ def configure_defaults(self, conf):
         conf.set('https_cert_file', None)
         conf.set('https_key_file', None)
 
+        # Default values for pagination requests
```
josephharrington: That makes sense, although it has not really been a problem in my experience. @wjdhollow you may want to file an issue if you feel strongly about it -- we'd want to do that all in one change instead of piecemeal.

app/util/conf/base_config_loader.py
wjdhollow: Haha. Nice. That's what I get for not keeping up on Github updates.

PR #397, opened 2017-08-01
Add pagination to v2 endpoints subjobs & atoms

nickzuber:
This adds the same pagination logic as the v2 builds endpoint in https://github.com/box/ClusterRunner/pull/394 but to subjobs and atoms.

I've abstracted some of the logic that was in the builds endpoint to a more generic utility method for taking an offset, limit, and list length and producing the paginated indices. 

A lot of this logic is very similar to https://github.com/box/ClusterRunner/pull/394, so there shouldn't be anything too surprising here

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -194,11 +194,6 @@ class Build(object):
         """
         num_subjobs = len(self._all_subjobs_by_id)
         start, end = get_paginated_indices(offset, limit, num_subjobs)
-
-        # Offset request/starting index is out of bounds, so return no results.
-        if start > num_subjobs:
-            return []
-
         requested_subjobs = islice(self._all_subjobs_by_id, start, end)
         return [self._all_subjobs_by_id[key] for key in requested_subjobs]
 
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 from enum import Enum
+from itertools import islice
 import os
 from queue import Queue, Empty
 import shutil
@@ -8,7 +9,7 @@ from threading import Lock, Thread
 import time
 import uuid
 
-from typing import Dict
+from typing import Dict, List
 
 from app.common.build_artifact import BuildArtifact
 from app.common.metrics import build_state_duration_seconds, ErrorType, internal_errors, serialized_build_time_seconds
@@ -22,6 +23,7 @@ from app.util.counter import Counter
 from app.util.exceptions import ItemNotFoundError
 import app.util.fs
 from app.util.log import get_logger
+from app.util.pagination import get_paginated_indices
 from app.util.single_use_coin import SingleUseCoin
 
 
@@ -61,7 +63,7 @@ class Build(object):
         self._project_type = None
         self._build_completion_lock = Lock()  # protects against more than one thread detecting the build's finish
 
-        self._all_subjobs_by_id = {}
+        self._all_subjobs_by_id = OrderedDict()
         self._unstarted_subjobs = None  # WIP(joey): Move subjob queues to BuildScheduler class.
         self._finished_subjobs = None
         self._failed_atoms = None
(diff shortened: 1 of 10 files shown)


boxcla:

app/web_framework/cluster_master_application.py
```diff
@@ -151,6 +151,16 @@ def get(self, build_id):
         self.write(response)
 
 
+class _V2SubjobsHandler(_SubjobsHandler):
+    def get(self, build_id):
+        offset, limit = self.get_pagination_params()
+        build = self._cluster_master.get_build(int(build_id))
+        response = {
+            'subjobs': [subjob.api_representation() for subjob in build.all_subjobs(offset, limit)]
```
nickzuber: Maybe we should consider a different name other than `all_subjobs`? Maybe like `get_subjobs` since it doesn't necessarily return _all_ the subjobs anymore

app/master/subjob.py
```diff
@@ -85,12 +87,28 @@ def api_representation(self):
...
     @property
-    def atoms(self):
+    def atoms(self) -> List['Atom']:
         """
-        :rtype: list[app.master.atom.Atom]
+        Returns a list of all atoms for this subjob
         """
         return self._atoms
 
+    def get_atoms(self, offset: int=None, limit: int=None) -> List['Atom']:
```
nickzuber: Note that this is different than the property `atoms`

app/master/build.py
```diff
@@ -61,7 +63,7 @@ def __init__(self, build_request):
         self._project_type = None
         self._build_completion_lock = Lock()  # protects against more than one thread detecting the build's finish
 
-        self._all_subjobs_by_id = {}
+        self._all_subjobs_by_id = OrderedDict()
```
nickzuber: If this isn't an OrderedDict, then the subjobs aren't guaranteed to be stored in the order we want or expect. 

```bash
❯ python
Python 3.4.5 (default, Jun  8 2017, 14:05:03) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> d = {}
>>> d['a'] = 1
>>> d['b'] = 1
>>> d['c'] = 1
>>> d
{'c': 1, 'b': 1, 'a': 1}
```

This is a problem when we want to limit the number of subjobs we show. If you have `offset=0` and `limit=10`, it's fair to assume you want to see the first 10 subjobs. If the subjobs aren't ordered, then we'd end up returning _some_ 10 subjobs instead of the first 10.

app/master/subjob.py
```diff
@@ -85,12 +87,28 @@ def api_representation(self):
...
+        Returns a list of atoms for this subjob
+        :param offset: The starting index of the requested build
+        :param limit: The number of builds requested
+        :rtype: list[app.master.atom.Atom]
+        """
+        num_atoms = len(self._atoms)
+        start, end = get_paginated_indices(offset, limit, num_atoms)
+
+        # Offset request/starting index is out of bounds, so return no results.
+        if start > num_atoms:
```
josephharrington: Is this needed? It seems like `self._atoms[start:end]` would do this automatically?

Actually, same question for the similar check before `islice` in `ClusterMaster.builds()` above, and the one in `all_subjobs`.

test/unit/master/test_cluster_master.py
```diff
@@ -311,3 +313,140 @@ def test_builds_with_pagination_request(
...
+            _PAGINATION_OFFSET + 1,
+            _PAGINATION_OFFSET + _PAGINATION_MAX_LIMIT
+        ),
+        too_high_offset=(
+            1000, _PAGINATION_LIMIT,
+            None,
+            None
+        ),
+    )
+    def test_subjobs_with_pagination_request(
```
josephharrington: Did you mean to name this something else? I'm guessing only one of these tests runs.

app/master/subjob.py
nickzuber: For `islice` the out of bounds index will throw an exception, but for a list, partitioning it like that shouldn't cause any problems so we can remove that check here, nice catch

test/unit/master/test_cluster_master.py
nickzuber: Oops yeah, this is meant to be for the atoms

nickzuber:
> Agreed! We could definitely be more consistent in what these methods are called (all_subjobs(), get_atoms(), builds()).

Any opinions? I think `get_` prefix makes the most sense here, but no strong opinions

PR #399, opened 2017-08-03
Fix build cancel when BUILDING

wjdhollow:
This PR addresses an issue where it is possible for slaves to get stuck working on a canceled build. 

This is what currently happens when the state is BUILDING and it is canceled.

1. Slaves are assigned to a build
2. Subjobs are distributed among slaves
3. Subjob completes
4. Slave reports results back to the master
5. Master skips logic to mark subjob complete, reassign subjobs, or free the executor

Step 5 is the most critical part. A slave cannot stop working on a build until all of it's executors are freed. This PR modifies the flow so it looks like this. Steps 1 - 4 are the same.

5. Master marks subjob complete
6. Slave executor is freed
7. When executor count is 0, slave teardown is called
8. Slave returns back to the pool
9. No new slaves are allocated to the build

### DIFF
--- a/app/master/build_scheduler.py
+++ b/app/master/build_scheduler.py
@@ -59,6 +59,8 @@ class BuildScheduler(object):
             return False
         if self._num_executors_allocated >= len(self._build.get_subjobs()):
             return False
+        if self._build.is_canceled:
+            return False
         return True
 
     def allocate_slave(self, slave: Slave) -> bool:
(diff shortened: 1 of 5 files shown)


boxcla:

app/master/build_scheduler.py
```diff
@@ -112,16 +114,17 @@ def execute_next_subjob_or_free_executor(self, slave):
...
+                        slave.start_subjob(subjob)
+                        subjob.mark_in_progress(slave)
 
                 except SlaveError as ex:
                     internal_errors.labels(ErrorType.SubjobWriteFailure).inc()  # pylint: disable=no-member
                     self._logger.warning('Failed to start {} on {}: {}. Requeuing subjob and freeing slave executor...',
                                          subjob, slave, repr(ex))
-                    self._build._unstarted_subjobs.put(subjob)  # todo: This changes subjob execution order. (Issue #226)
-                    # An executor is currently allocated for this subjob in begin_subjob_executions_on_slave.
-                    # Since the slave has been marked for shutdown, we need to free the executor.
```
josephharrington: I'm fine with removing the todo since there is an issue filed for it but should we keep the comment (in some form -- we could remove the reference to begin_subjob_executions_on_slave)?

app/master/build_scheduler.py
```diff
@@ -112,16 +114,17 @@ def execute_next_subjob_or_free_executor(self, slave):
                 subjob = self._build._unstarted_subjobs.get(block=False)
                 self._logger.debug('Sending {} to {}.', subjob, slave)
                 try:
-                    slave.start_subjob(subjob)
-                    subjob.mark_in_progress(slave)
+                    if self._build.is_canceled:
+                        self._free_slave_executor(slave)
```
josephharrington: Could we move this check up, even before acquiring the lock? Maybe like:
```python
if self._build.is_canceled:
    self._free_slave_executor(slave)
    return

try:
    with self._subjob_assignment_lock:
        ...
```
It doesn't really matter, but I feel like we should not even try to get the lock or pull from the unstarted_subjobs queue if the build is already canceled.

I know you don't love multiple return points in a method. 😬 An alternative would be just an if/else around the whole thing.

While we're here, I'm noticing how unnecessarily wide the outer try/except is. The `except Empty` really only applies to that one line `self._build._unstarted_subjobs.get(block=False)`. Not sure if narrowing that try/except would help clean up the flow or not. :woman_shrugging:

app/master/build_scheduler.py
wjdhollow: I removed it because it referenced behavior outside of this function, and the issue was already clear what causes the issue (items being appended to the queue). Comments that reference outside behavior runs the risk of getting stale. It's like trying to look outside of a fish bowl.

But I'll put it back. It's a nice to know and still relevant.

app/master/build_scheduler.py
wjdhollow: Oh, good point. Acquiring locks are expensive.

PR #402, opened 2017-08-09
Migrate builds from master to BuildStore

### LINKED ISSUE #398: Build Result Storage Service
Right now all of the builds information is stored on the master service which is part of the reason why we lose all the build information between restarts. We should abstract this concept of managing builds from the master into its own service. From there, we'd ultimately be able to do things like persist this information in a database.

The `BuildsManager` class could look something like this:

```python
class BuildsManager(object):
    """
    Builds manager service that stores and handles all builds.
    """

    def __init__(self):
        self._all_builds_by_id = OrderedDict()

    def get(self, build_id: int) -> Build:
        """
        Returns a build by id
        :param build_id: The id for the build whose status we are getting
        """
        build = self._all_builds_by_id.get(build_id)
        if build is None:
            raise ItemNotFoundError('Invalid build id: {}.'.format(build_id))

        return build

    def add(self, build: Build):
        """
        Add new build to collection
        """
        self._all_builds_by_id[build.id()] = build

```

All the build creation can be still done on the master, but the actual storing and distributing of those builds would be handled in `BuildsManager`.

Goals
------

The primary goal of this proposal is to move all the builds data off of the master and into an abstracted manager class. This class as it stands wouldn't change any of the current behavior in cluster runner

Notes
------

1. We can eventually pass in a session id to the `get` and `add` methods later down the line when we want to persist builds between restarts. I don't think we'd want to care about the session id in this context until we start moving things to an actual database though, but something to keep in mind. <br />
This could look something like a dictionary of session id's that each map to an ordered dictionary of builds that were created in that session. 

Questions
----------

1. Should the `BuildsManager` class be a singleton? I can't imagine a case where we'd want to have multiple instances of this class.

nickzuber:
Simply move all of the builds from the master service and into a BuildStore class. 

Note that all the builds are still being _created_ by the master, just stored in the BuildStore. 

```python
# build_store.py

@classmethod
def add(cls, build: Build):
    """
    Add new build to collection
    :param build: The build to add to the store
    """
    cls._all_builds_by_id[build.build_id()] = build
```

We'll probably change this later when we add a database/caching layer here, since the build `id`s will be autogenerated. 

Closes #398

### DIFF
--- a/app/master/build_store.py
+++ b/app/master/build_store.py
@@ -6,7 +6,7 @@ from app.master.build import Build
 from app.util.exceptions import ItemNotFoundError
 
 
-class BuildStore():
+class BuildStore:
     """
     Build storage service that stores and handles all builds.
     """
--- a/app/master/build_store.py
+++ b/app/master/build_store.py
@@ -6,7 +6,7 @@ from app.master.build import Build
 from app.util.exceptions import ItemNotFoundError
 
 
-class BuildStore(object):
+class BuildStore():
     """
     Build storage service that stores and handles all builds.
     """
@@ -25,11 +25,12 @@ class BuildStore(object):
         return build
 
     @classmethod
-    def get_range(cls, start: int, end: int) -> List['Build']:
+    def get_range(cls, start: int, end: int) -> List[Build]:
         """
         Returns a list of all builds.
         :param start: The starting index of the requested build
-        :param end: The number of builds requested
+        :param end: 1 + the index of the last requested element, although if this is greater than the total number
+                    of builds available the length of the returned list may be smaller than (end - start)
         """
         requested_builds = islice(cls._all_builds_by_id, start, end)
         return [cls._all_builds_by_id[key] for key in requested_builds]
--- a/app/master/build_store.py
+++ b/app/master/build_store.py
@@ -1,0 +1,50 @@
+from collections import OrderedDict
+from itertools import islice
+from typing import List
+
+from app.master.build import Build
+from app.util.exceptions import ItemNotFoundError
+
+
+class BuildStore(object):
+    """
+    Build storage service that stores and handles all builds.
+    """
+    _all_builds_by_id = OrderedDict()
+
+    @classmethod
+    def get(cls, build_id: int) -> Build:
+        """
+        Returns a build by id
+        :param build_id: The id for the build whose status we are getting
+        """
+        build = cls._all_builds_by_id.get(build_id)
+        if build is None:
+            raise ItemNotFoundError('Invalid build id: {}.'.format(build_id))
+
+        return build
+
+    @classmethod
+    def get_range(cls, start: int, end: int) -> List['Build']:
+        """
+        Returns a list of all builds.
+        :param offset: The starting index of the requested build
+        :param limit: The number of builds requested
+        """
+        requested_builds = islice(cls._all_builds_by_id, start, end)
+        return [cls._all_builds_by_id[key] for key in requested_builds]
+
+    @classmethod
+    def add(cls, build: Build):
+        """
+        Add new build to collection
+        :param build: The build to add to the store
+        """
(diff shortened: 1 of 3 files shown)


boxcla:

app/master/build_store.py
```diff
@@ -0,0 +1,50 @@
...
+        if build is None:
+            raise ItemNotFoundError('Invalid build id: {}.'.format(build_id))
+
+        return build
+
+    @classmethod
+    def get_range(cls, start: int, end: int) -> List['Build']:
+        """
+        Returns a list of all builds.
+        :param offset: The starting index of the requested build
```
ethomas2: The variable names are start and end but the docstrings mention offset and limit

nickzuber: Nice catch, forgot to update these when I changed them

app/master/build_store.py
```diff
@@ -0,0 +1,50 @@
+from collections import OrderedDict
+from itertools import islice
+from typing import List
+
+from app.master.build import Build
+from app.util.exceptions import ItemNotFoundError
+
+
+class BuildStore(object):
```
cmcginty: `class BuildStore:` is fine since all classes inherit from `object` in python 3

app/master/build_store.py
```diff
@@ -0,0 +1,50 @@
...
+    @classmethod
+    def add(cls, build: Build):
+        """
+        Add new build to collection
+        :param build: The build to add to the store
+        """
+        cls._all_builds_by_id[build.build_id()] = build
+
+    @classmethod
+    def size(cls) -> int:
```
cmcginty: Instead of a new method it would be better to fit the standard python data model by implementing [`__len__`](https://docs.python.org/3/reference/datamodel.html#object.__len__)

nickzuber: Eventually I'd like this method to return a tuple of number of cached builds and number of total builds in the database, so I'm not sure if `__len__` would be appropriate here. I'm open to a better name though, not sure if `size` is the best choice here

app/master/build_store.py
```diff
@@ -0,0 +1,50 @@
...
+        :param build_id: The id for the build whose status we are getting
+        """
+        build = cls._all_builds_by_id.get(build_id)
+        if build is None:
+            raise ItemNotFoundError('Invalid build id: {}.'.format(build_id))
+
+        return build
+
+    @classmethod
+    def get_range(cls, start: int, end: int) -> List['Build']:
```
josephharrington: You should be able to remove the quotes from around `'Build'`. Quotes should only be needed for forward references.
https://www.python.org/dev/peps/pep-0484/#forward-references

app/master/build_store.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Returns a list of all builds.
+        :param start: The starting index of the requested build
+        :param end: The number of builds requested
```
josephharrington: This docstring is incorrect. 

It should be something more like:
`:param end: 1 + the index of the last requested element, although if this is greater than the total number of builds available the length of the returned list may be smaller than (end - start)`

app/master/cluster_master.py
```diff
@@ -329,7 +326,7 @@ def get_path_for_build_results_archive(self, build_id: int, is_tar_request: bool
         :param is_tar_request: If true, download the tar.gz archive instead of a zip.
         :return: The path to the archived results file
         """
-        build = self._all_builds_by_id.get(build_id)  # type: Build
+        build = BuildStore.get(build_id)  # type: Build
```
josephharrington: I believe you can remove the typehint comment here now since `BuildStore.get` has its own hint.

app/master/build_store.py
josephharrington: Agreed. You'll see `Class(object):` instead of `Class:` used sporadically throughout the codebase but we should standardize on the latter.

PR #404, opened 2017-08-12
Fix prometheus metric for number of dead slaves

ethomas2:
The prometheus metric for number of dead slaves was counting the total
number of slaves cluster runner has ever known about that are no longer
connected to cluster runner. Even if a slave was purposely disconnected
from cluster runner via graceful shutdown, the prometheus metric would
still count the slave in the total number of dead slaves.

### DIFF
--- a/app/common/metrics.py
+++ b/app/common/metrics.py
@@ -62,12 +62,17 @@ class SlavesCollector:
     def collect(self) -> Iterator[GaugeMetricFamily]:
         active, idle, dead = 0, 0, 0
         for slave in self._get_slaves():
-            if not slave.is_alive(use_cached=True):
-                dead += 1
-            elif slave.current_build_id is not None:
+            if slave.is_alive(use_cached=True) and slave.current_build_id is not None:
                 active += 1
-            else:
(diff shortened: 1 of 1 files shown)


boxcla:

app/common/metrics.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, get_slaves: Callable[[], List['app.master.slave.Slave']]):
     def collect(self) -> Iterator[GaugeMetricFamily]:
         active, idle, dead = 0, 0, 0
         for slave in self._get_slaves():
-            if not slave.is_alive(use_cached=True):
+            if not slave.is_alive(use_cached=True) and not slave.is_shutdown():
```
tjlee0909: Could you add a comment here, stating that the reason why we have this check here, is that if ```slave.is_shutdown``` is True, then that means we have deliberately, and gracefully removed the slave--and that we don't want to categorize deliberate graceful shutdowns as 'dead slaves'.

ethomas2: Done.

PR #405, opened 2017-08-15
Handle build cancellation before atomization and for queued builds

shriganeshs:
This PR handles the build cancellation before atomization and for queued builds.
It basically moves the build atomization and subjob creation code into preparing callback function to avoid explicit checks for canceled build.
Secondly, on build cancel it sets kill_event for all subprocesses for the build.

This PR also makes updates such as 
- post-build tasks will not be executed for canceled build and
- canceled build will not be marked as failed
- for canceled build AtomizerError exception will be attached with more information to make logs readable.

### DIFF
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -11,6 +11,7 @@ import uuid
 
 from typing import Dict, List
 
+from app.master.atomizer import AtomizerError
 from app.common.build_artifact import BuildArtifact
 from app.common.metrics import build_state_duration_seconds, ErrorType, internal_errors, serialized_build_time_seconds
 from app.master.build_fsm import BuildFsm, BuildEvent, BuildState
@@ -149,8 +150,6 @@ class Build(object):
         if not self._preparation_coin.spend():
             raise RuntimeError('prepare() was called more than once on build {}.'.format(self._build_id))
 
-        self._logger.error('SS_DEBUG sleeping now')
-        time.sleep(10)
         self._state_machine.trigger(BuildEvent.START_PREPARE, subjob_calculator=subjob_calculator)
 
     def build_id(self):
@@ -294,7 +293,6 @@ class Build(object):
         Atomize the build. This method is triggered by a state machine transition to the PREPARING state.
         :param event: The Fysom event object
         """
-        self._logger.error('SS_DEBUG _on_enter_preparing_state I am here')
         self._logger.info('Fetching project for build {}.', self._build_id)
         self.project_type.fetch_project()
         self._logger.info('Successfully fetched project for build {}.', self._build_id)
@@ -305,8 +303,16 @@ class Build(object):
 
         subjob_calculator = getattr(event, 'subjob_calculator', None)
         if subjob_calculator is None:
-            raise RuntimeError('Build failed while trying to get subjobs in PREPARING state.')
+            raise RuntimeError('Build failed due to invalid subjob_calculator in PREPARING state.')
+        #try:
         subjobs = subjob_calculator.compute_subjobs_for_build(self._build_id, job_config, self.project_type)
+        # except AtomizerError:
+        #     # Ignore the AtomizerError if the build is canceled (AtomizerError can be intentional when build
+        #     # is in canceled state.
+        #     if self.is_canceled:
+        #         raise
+        #     else:
+        #         raise
 
         self._unstarted_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Move this into BuildScheduler?
         self._finished_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Remove this and just record finished count.
@@ -336,6 +342,10 @@ class Build(object):
         self._state_machine.trigger(BuildEvent.CANCEL)
 
     def _on_enter_canceled_state(self, event):
+        # Set the kill_event to kill the subprocesses for the build
+        self._logger.notice('Set the kill_event for subprocesses for build {}.', self._build_id)
+        self.project_type.kill_subprocesses()
+
         # Deplete the unstarted subjob queue.
         # WIP(joey): Just remove this completely and adjust behavior of other methods based on self._is_canceled().
         # TODO: Handle situation where cancel() is called while subjobs are being added to _unstarted_subjobs
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -303,17 +303,14 @@ class Build(object):
 
         subjob_calculator = getattr(event, 'subjob_calculator', None)
         if subjob_calculator is None:
-            raise RuntimeError('Build failed due to invalid subjob_calculator in PREPARING state.')
-        #try:
-        subjobs = subjob_calculator.compute_subjobs_for_build(self._build_id, job_config, self.project_type)
-        # except AtomizerError:
-        #     # Ignore the AtomizerError if the build is canceled (AtomizerError can be intentional when build
-        #     # is in canceled state.
-        #     if self.is_canceled:
-        #         raise
-        #     else:
-        #         raise
-
+            raise RuntimeError('Build failed as subjob_calculator is not set for this build {}.', self._build_id)
+        try:
+            subjobs = subjob_calculator.compute_subjobs_for_build(self._build_id, job_config, self.project_type)
+        except AtomizerError as ex:
+            if self.is_canceled:
+                # Add info to the AtomizerError error message. If build is canceled, then AtomizerError is expected.
+                ex.args = ex.args + ('This is expected as Build {} is CANCELED.'.format(self._build_id),)
+                raise
         self._unstarted_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Move this into BuildScheduler?
         self._finished_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Remove this and just record finished count.
 
--- a/app/master/build.py
+++ b/app/master/build.py
@@ -77,6 +77,7 @@ class Build(object):
             enter_state_callbacks={
                 BuildState.ERROR: self._on_enter_error_state,
                 BuildState.CANCELED: self._on_enter_canceled_state,
+                BuildState.PREPARING: self._on_enter_preparing_state,
             },
             leave_state_callbacks=leave_state_callbacks
         )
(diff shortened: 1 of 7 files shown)


app/master/build.py
```diff
@@ -309,6 +288,41 @@ def _on_enter_error_state(self, event):
         self._error_message = getattr(event, 'error_msg', default_error_msg)
         self._logger.warning('Build {} failed: {}', self.build_id(), self._error_message)
 
+    def _on_enter_preparing_state(self, event):
+        """
+        Prepare the build by atomization and subjobs creation.
+        This method is triggered by a state machine transition to the PREPARING state.
+        :param event: The Fysom event object
```
wjdhollow: Typehint please.

app/master/build.py
```diff
@@ -309,6 +288,41 @@ def _on_enter_error_state(self, event):
...
+        """
+        self._logger.info('Fetching project for build {}.', self._build_id)
+        self.project_type.fetch_project()
+        self._logger.info('Successfully fetched project for build {}.', self._build_id)
+
+        job_config = self.project_type.job_config()
+        if job_config is None:
+            raise RuntimeError('Build failed while trying to parse clusterrunner.yaml.')
+
+        subjob_calculator = getattr(event, 'subjob_calculator', None)
```
wjdhollow: Sorry, I thought that the build callback allowed you to specify additional kwargs to the function. I didn't realize that it sets a member variable within the Event object instead.

Instead of using `getattr`, how about just instantiating a new instance of SubjobCalculator instead.

```
subjob_calculator = SubjobCalculator()
vs
subjob_calculator = getattr(event, 'subjob_calculator', None)
```

app/master/build.py
```diff
@@ -326,6 +340,10 @@ def cancel(self):
         self._state_machine.trigger(BuildEvent.CANCEL)
 
     def _on_enter_canceled_state(self, event):
+        # Set the kill_event to kill the subprocesses for the build
+        self._logger.notice('Set the kill_event for subprocesses for build {}.', self._build_id)
+        self.project_type.kill_subprocesses()
+
```
wjdhollow: Was there a problem with subprocesses not being killed? 

Is the bottom block for draining the unstarted_subjobs queue still necessary?

app/master/build.py
```diff
@@ -148,30 +150,7 @@ def prepare(self, subjob_calculator):
...
-        self._finished_subjobs = Queue(maxsize=len(subjobs))  # WIP(joey): Remove this and just record finished count.
-
-        for subjob in subjobs:
-            self._all_subjobs_by_id[subjob.subjob_id()] = subjob
-            self._unstarted_subjobs.put(subjob)
-
-        self._timing_file_path = self._project_type.timing_file_path(job_config.name)
-        app.util.fs.create_dir(self._build_results_dir())
-        self._state_machine.trigger(BuildEvent.FINISH_PREPARE)
+        self._state_machine.trigger(BuildEvent.START_PREPARE, subjob_calculator=subjob_calculator)
```
wjdhollow: See my comment on instantiating the subjob_calculator in the callback. If we do that, don't forge to change the method signature for `def prepare()`

test/unit/master/test_build.py
```diff
@@ -273,6 +273,25 @@ def test_cancel_is_a_noop_if_build_is_already_finished(self):
...
+        build._project_type.kill_subprocesses.assert_called_once_with()
+
+    def test_prepare_callback_handles_atomizererror_exception_and_raises_again(self):
+        job_config = self._create_job_config()
+        subjobs = self._create_subjobs(job_config=job_config)
+        mock_subjob_calculator = self._create_mock_subjob_calc(subjobs, compute_subjob_raises_exception=True)
+        build = self._create_test_build(BuildStatus.QUEUED)
+
+        with self.assertRaises(AtomizerError):
+            build.prepare(mock_subjob_calculator)
```
wjdhollow: You should have a test that asserts that cancelling before calling `build.prepare()` does not compute subjobs.

wjdhollow: Is there already a test for queued builds?

app/master/build.py
shriganeshs: `subjob_calculator` comes from `BuildRequestHandler`. Personally I will not prefer it to be part of callback `build._on_enter_preparing_state`. Also, I am not sure but, to move it we may need to make at multiple places.

app/master/build.py
shriganeshs: Previously we were not killing the subprocesses for the build namely atomizer. We need to kill the subprocesses on cancel.
I am not 100% sure about draining of `unstarted_subjobs`. Do you want to say that as subprocesses are killed, it does not matter? My opinion is its a good idea to clean in any case.

test/unit/master/test_build.py
shriganeshs: I will add the test.

app/master/build.py
shriganeshs: Done.

test/unit/master/test_build.py
```diff
@@ -273,6 +273,25 @@ def test_cancel_is_a_noop_if_build_is_already_finished(self):
...
 
+    def test_cancel_calls_kill_subprocesses_to_set_kill_event(self):
+        mock_slave = self._create_mock_slave()
+        build = self._create_test_build(BuildStatus.QUEUED, slaves=[mock_slave])
+        self.patch('app.master.build.util')
+        build.generate_project_type()
+
+        build.cancel()
+
+        build._project_type.kill_subprocesses.assert_called_once_with()
```
ethomas2: This will fail if we ever make `kill_subprocess` accept arguments.

How about
```
self.assertTrue(build._project_type.kill_subprocess.called,
                          'Build did not attempt to kill subprocess')
```

Also, why is the build trying to kill a subprocess anyway if it's in the QUEUED state? It shouldn't have a subprocess unless it's PREPARING, correct?

test/unit/master/test_build.py
```diff
@@ -563,13 +582,16 @@ def _create_mock_slave(self, num_executors=5):
...
-    def _create_mock_subjob_calc(self, subjobs):
+    def _create_mock_subjob_calc(self, subjobs, compute_subjob_raises_exception=False):
         """
         :type subjobs: list[Subjob]
         :rtype: SubjobCalculator
         """
         mock_subjob_calculator = MagicMock(spec_set=SubjobCalculator)
-        mock_subjob_calculator.compute_subjobs_for_build.return_value = subjobs
+        if compute_subjob_raises_exception:
+            mock_subjob_calculator.compute_subjobs_for_build.side_effect = [AtomizerError('Atomizer command failed!')]
```
ethomas2: the rhs doesn't need to be an array. You can just do 
`= AtomizerError('Atomizer command failed!')`
or
`= AtomizerError`

app/master/build.py
wjdhollow: > Personally I will not prefer it to be part of callback build._on_enter_preparing_state

It is already "part" of the callback. The only time that object is used is within that method.

shriganeshs:
@wjdhollow Lets wait for Joey's comments.
I think callback should be lightweight and quick/short. Being said that, I do not mean "instantiating SubjobCalculator in callback" will make it heavyweight or slow. I tried to move the instantiation to callback but it seems its more work.

app/master/build.py
```diff
@@ -309,6 +288,42 @@ def _on_enter_error_state(self, event):
...
+
+        subjob_calculator = getattr(event, 'subjob_calculator', None)
+        if subjob_calculator is None:
+            raise RuntimeError('Build failed as subjob_calculator is not set for this build {}.', self._build_id)
+        try:
+            subjobs = subjob_calculator.compute_subjobs_for_build(self._build_id, job_config, self.project_type)
+        except AtomizerError as ex:
+            if self.is_canceled:
+                # Add info to the AtomizerError error message. If build is canceled, then AtomizerError is expected.
+                ex.args = ex.args + ('This is expected as Build {} is CANCELED.'.format(self._build_id),)
```
josephharrington: Is this exposed anywhere if the build is canceled? It looks like the exception message is only used if the build is not canceled, so maybe this extra message (and the try/except) is not needed.

test/unit/master/test_build.py
```diff
@@ -273,6 +273,26 @@ def test_cancel_is_a_noop_if_build_is_already_finished(self):
...
+        with self.assertRaises(AtomizerError):
+            build.prepare(mock_subjob_calculator)
+
+    def test_cancel_before_prepare_does_not_call_compute_subjobs(self):
+        job_config = self._create_job_config()
+        subjobs = self._create_subjobs(job_config=job_config)
+        mock_subjob_calculator = self._create_mock_subjob_calc(subjobs)
+        build = self._create_test_build(BuildStatus.QUEUED)
+
+        build.cancel()
```
josephharrington: You forgot to call `build.prepare()`.

test/unit/master/test_cluster_master.py
```diff
@@ -216,7 +216,10 @@ def test_handle_result_reported_from_slave_when_build_is_canceled(self):
         build_id = 1
         slave_url = "url"
         build = Build(BuildRequest({}))
+        self.patch('app.master.build.util')
+        build.generate_project_type()
```
josephharrington: Why did you make this change? Just curious.

app/master/build.py
shriganeshs: This is mainly to make logs more readable. So this exception will put stack trace in the logs with 'this is expected' when build is canceled.

app/master/build.py
```diff
@@ -326,6 +341,11 @@ def cancel(self):
         self._state_machine.trigger(BuildEvent.CANCEL)
 
     def _on_enter_canceled_state(self, event):
+        # Set the kill_event to kill the subprocesses for the build, if build cancel is called during PREPARING state
+        if event.src == BuildState.PREPARING:
+            self._logger.notice('Set the kill_event for subprocesses for build {}.', self._build_id)
```
shriganeshs: I agree with `omitting the check for BuildState.PREPARING`. Its simpler, and will work when in future if we add any subprocess to other BuildStates.

app/master/build.py
josephharrington: @shriganeshs Where does this get logged?

test/unit/master/test_build.py
shriganeshs: This one tests `cancel before prepare`, so not calling prepare.

test/unit/master/test_cluster_master.py
shriganeshs: Now as the `cancel callback` is calling `project_type.kill_subprocesses` we need to generate `project_type` object.

test/unit/master/test_build.py
josephharrington: But I don't think this tests anything otherwise. As this test is currently written, `mock_subjob_calculator` has nothing to do with the build object (since it is never passed to prepare()).

It would be impossible for this test to ever fail since `mock_subjob_calculator` is never passed to the build. In my opinion the only valuable thing to test is that `compute_subjobs_for_build` is not called if `build.cancel()` happens before `build.prepare()`.

test/unit/master/test_cluster_master.py
josephharrington: Ok. I ran this test without the added `build.generate_project_type()` call and it still passed, probably because the build is not in the `PREPARING` state. But if you remove the check for that state then this will probably be needed.

shriganeshs: Yes correct and now I have removed that check.

shriganeshs:
Changed the methods in `SubjobCalculator` to just be module-level function. So now there is no need to passing `SubjobCalcular` object to callback event handlers. Also added unit tests around build_request_handler and setting kill event.
As suggested by Joey, merging this PR and I will create a new PR with functional tests for build cancellation.

PR #406, opened 2017-08-21
[WIP] Build Persistence w/ SQLite & SQLAlchemy

### LINKED ISSUE #401: Build Persistence Proposal (SQLite vs. Redis)
Overview
--

This is a proposed design for implementing a storage layer for ClusterRunner so we can persist build history.

### The Problem

As of right now, all of the builds that are created are stored directly on the master service. This causes a few annoyances:

 - We lose all the builds if the master service goes down or we need to restart
 - We are limited to only having a single master service per session

Whenever we initialize a master service, we remove all of the old builds before we start.

```python
if os.path.exists(self._master_results_path):
    fs.async_delete(self._master_results_path)
fs.create_dir(self._master_results_path)
```

We do this because since all the builds are stored on the master, we have no way of getting this information again if the service goes down at any point. We also restart the build ids at 1 so this information would get overwritten anyways. If we move this information off of the master and into some kind of data store, then we wouldn't have to sacrifice this data we have already on disk, and we would still have this information available if we restart the master.

### Possible Solutions

There are a bunch of different ways we can persist this information, but here are a few options that I propose we can go forward with:

 - SQLite Database
 - Redis Data Store

For reference, I ran some [benchmarks to test their performance](https://github.com/box/ClusterRunner/issues/400) in realistic (I hope) situations that they would be used in with ClusterRunner. It's important to note that these benchmarks are not necessarily representative of SQLite and Redis performance in an objective sense. These are two different services, so naturally they would behave differently in the same situations. Also in these particular benchmarks, there are ways to optimize some operations -- however that wouldn't be relevant to the particular implementation we'd be using in ClusterRunner so we don't do that.

SQLite
--

[SQLite](https://www.sqlite.org/about.html) is an easy to use, easy to configure, and lightweight database.

### General Information

SQLite is really easy to set up because the entire database is really just a file. This makes it portable and simple to back up. There's no separate server process involved since the database is in a file -- all operations are just reads and writes to that file. A good quote from the SQLite website is:

> Think of SQLite not as a replacement for Oracle but as a replacement for fopen()

#### Configuration

Since SQLite is essentially just a file, there's no real overhead with set up, configuration, or managing. The only requirement for using SQLite is to be able to read from disk. No additional set up or overhead required in ClusterRunner is nice to have.

#### Efficient Queries with Autoincrement

By default, SQLite [indexes items with a field called `ROWID`](https://sqlite.org/autoinc.html). This field is autogenerated starting at 1 and continuing until the largest possible integer (9223372036854775807). This field autoincrements and is _incredibly_ performant to use in queries. For reference, access queries for builds are [about 4x faster using `ROWID` in SQLite than Redis](https://github.com/box/ClusterRunner/issues/400#rowid). This behavior is already what we want for generating build `id`s so we can get this right out of the box.

The reason queries with `ROWID` are so fast is because, as [the SQLite documentation states](http://sqlite.org/lang_createtable.html#rowid):

> The data for rowid tables is stored as a B-Tree structure containing one entry for each table row, using the rowid value as the key. **This means that retrieving or sorting records by rowid is fast**. Searching for a record with a specific rowid, or for all records with rowids within a specified range is around twice as fast as a similar search made by specifying any other PRIMARY KEY or indexed value.

#### True Concurrency

SQLite is not great with concurrency. Any amount of services can _read_ from the database at any given time, but only one can _write_ at any given time. The writes are queued and performed synchronously. This could come into play if we have multiple master services trying to write builds to the database at the same time, however we should note that we can expect a writer only spends a few milliseconds writing to disk, so this probably won't be an issue in practice for our use case.

#### Maybe Race Conditions with `ROWID`

SQLite only allows one write to the database at a time which means that we hopefully don't have to worry about race conditions. The one issue that I thought of was if we ever had two masters trying to create a build at the same time. This wouldn't be an issue with inserting the build, however there _could_ be some trouble when it comes to getting that build's `id` immediately after it's inserted.

There's a feature that allows us to call `lastrowid` from the SQLite cursor to get the last `id` that was inserted into the table. This is used to get the build's `id` right after we insert it since the `id` comes from the autoincremented `ROWID` within our SQLite table. If more than one master inserted a build before we called `lastrowid`, we could end up in a situation where we think two different builds have the same `id`. 

However, this can be avoided if each master has its own connection to the database. This is because [a SQLite connection cursor is the one who knows the `ROWID` of the last inserted row that _it_ inserted](https://stackoverflow.com/a/6242813/5055063). I think (and correct me if I'm wrong) that one master instance will not be trying to create multiple builds at once. If this is true, then if each master instance try to create a build at the same time, each of their SQLite connections will know the last `ROWID` that _they themselves_ inserted, therefore **not** creating a race condition on getting the correct build `id`.

If for some reason this ever becomes a problem later down the line, we could always make the BuildStore class a singleton and place all incoming build requests into a queue that sequentially handles insertions one at a time. I highly doubt it'd ever come to this though, but if it does we could solve the problem.

### How It Would Look

Whenever we start up ClusterRunner, we'd want to connect to our database in a BuildStore and connect an instance of that BuildStore to our master. Also as a side note, a cool thing about SQLite limit select queries is that they use an `offset, limit` pattern just like we do in our pagination, so no extra logic needed!

#### build_store.py
```python
import sqlite3
from typing import List

from app.master.build import Build
from app.util.exceptions import ItemNotFoundError


class BuildStore(object):
    """
    Build storage service that stores and handles all builds.
    """

    def __init__(self):
        self._sqlite_db = sqlite3.connect('clusterrunner.db')
        self._sqlite_connection = self._sqlite_db.cursor()

        # Check if builds table exists, if not then create it
        if not self._sqlite_connection.execute("SELECT name FROM sqlite_master WHERE type='table' AND name='builds'").fetchone()
            sqlite_connection.execute('CREATE TABLE builds(build TEXT)')

    def get(self, build_id: int) -> Build:
        """
        Returns a build by id
        :param build_id: The id for the build whose status we are getting
        """
        try:
            self._sqlite_connection.execute('SELECT rowid, * FROM builds WHERE rowid=?', (build_id,))
            build_id, build = sqlite_connection.fetchone()
        except:
            raise ItemNotFoundError('Invalid build id: {}.'.format(build_id))

        return build

    def get_range(offset: int=None, limit: int=None) -> List['Build']:
        """
        Returns a list of all builds.
        :param offset: The starting index of the requested build
        :param limit: The number of builds requested
        """
        try:
            if offset and limit:
                self._sqlite_connection.execute('SELECT rowid, * FROM builds limit ?, ?', (offset, limit))
            else:
                self._sqlite_connection.execute('SELECT rowid, * FROM builds')

            results = self._sqlite_connection.fetchall()
        except:
            # Raise selection failure

        return [build for build_id, build in results]

    def add(self):
        """
        Add new build to collection
        """
        # Create build here
        # Builds will have a build_id of `None` when first created and this build_id will be set after insertion
        build = # ...
        try:
            self._sqlite_connection.execute('INSERT INTO builds VALUES(?)', (build,))
            self._sqlite_db.commit()
            build_id = self._sqlite_connection.lastrowid  # Gets the build_id of build we just inserted
            # Finish build creation by assigning its build_id
            # ...
        except:
            # Raise insertion failure

```

#### master_service.py
```python
class ClusterMaster(ClusterService):
    """
    The ClusterRunner Master service: This is the main application class that the web framework/REST API sits on top of.
    """

    API_VERSION = 'v1'

    def __init__(self):
        self._all_slaves_by_url = {}
        self._all_builds_by_id = OrderedDict()
        self._scheduler_pool = BuildSchedulerPool()
        self._build_request_handler = BuildRequestHandler(self._scheduler_pool)
        self._build_request_handler.start()
        self._slave_allocator = SlaveAllocator(self._scheduler_pool)
        self._slave_allocator.start()
        # Assign instance of BuildStore to master service
        self._build_store = BuildStore()

        SlavesCollector.register_slaves_metrics_collector(lambda: self.all_slaves_by_id().values())

```

Redis
--

[Redis](https://redis.io/) is a performant in memory data store that's primarily used for caching information.

### General Information

Since Redis is an in memory data store, this means all of its data is kept in memory. We can write this information to disk, and Redis will do this automatically, after certain intervals of time. Since data is kept in memory, this makes it fast when it comes to setting and getting key values. This also means that the builds are organized in a different way than they would be in SQLite -- in Redis, a key is assigned a value so we'd have a build `id` assigned to a build whereas in SQLite we have a row in a table that has a build `id` and its respective build. Subtle, but still a difference. 

#### Expire Keys

Redis has an expire feature that can expire data after a certain amount of time. This can be pretty useful for us since we'd have to clear out old builds eventually -- with the expire feature, we can help automate some of this process. The one downside to this feature, however, is that [any attempts to manipulate the contents of a build will result in the expire tag to render itself void](https://redis.io/commands/expire). This means we have to be very careful with how we handle the data stored or else we risk removing the expire tag which can be annoying.

#### Performance

Redis is really performant and good at doing insertion and access operations. While looking at [some benchmarks](https://github.com/box/ClusterRunner/issues/400), it's clear to see the power behind Redis and how quickly it can do things. This does, however, come at a cost regarding persistence, which I talk more about later, but in general Redis does a quicker job of storing and looking up information compared to SQLite _depending on what we use for indices_.

#### Configuration Overhead

Redis needs its own server process running for it to work, so this requires installation and set up overhead. Also if we wanted to configure things such as disk persistence timing intervals we'd have to make sure that's set up correctly as well. We'd also need to keep tabs on and maintain the Redis server to make sure it stays up and doesn't have any issues.

#### Deterministic Persistence

Redis actually [isn't great at deterministically persisting data](https://redis.io/topics/persistence). Since it's an in memory data store, all of the builds would naturally be kept in memory. There _is_ a persistence feature where the data store is written to disk, but this isn't super reliable. That's because the intervals which we can write to disk are limiting in the sense of performance. The more frequently we write to disk, the less performant our Redis cache will be. And if we decide to write to disk less often, we run into the risk of losing data stored in memory if our Redis server goes down at any point for any reason.

I'd argue that we can't afford to lose this kind of data. Let's assume user A attempts to create a build. ClusterRunner then creates that build and let's say it's the 100th build in our data store. Therefore, it gets the build `id` of 100. User A now expect the build with an `id` of 100 to have that specific build. Now, let's say something happens and our Redis server goes down before this build was written to disk. Later, user B comes along and creates a build. Now _this_ build is given an `id` of 100 since we lost all the history of the original 100th build. When user A comes to check on her build 100, she get's a totally different build. I think this kind of technical behavior is major and something we can't afford to have happen in our system.

So we have the option of writing to disk more frequently, but even if we shorten that window, there's _still_ an opportunity for this data loss happening (it just makes it less likely). The only way to prevent this is to write to disk every time a new build is created, which could be very expensive. A quote from the Redis website:

> RDB needs to fork() often in order to persist on disk using a child process. Fork() can be time consuming if the dataset is big, and may result in Redis to stop serving clients for some millisecond or even for one second if the dataset is very big and the CPU performance not great.

The main reason we'd want to use Redis over SQLite is for its performance, and doing something like this would make that advantage almost moot.

### How It Would Look

For installing Redis, you can [refer to this guide](https://redis.io/topics/quickstart).

First we would need to spin up a Redis server.

```bash
❯ redis-server /path/to/redis.conf
```

Then we connect to our Redis service in our BuildStore and master services can just interface with that BuildStore to create and get builds.

#### build_store.py
```python
import redis
from typing import List

from app.master.build import Build
from app.util.exceptions import ItemNotFoundError


class BuildStore(object):
    """
    Build storage service that stores and handles all builds.
    """

    def __init__(self):
        # Connect to Redis service
        self._redis_connection = redis.Redis('localhost', db=0)

    def get(self, build_id: int) -> Build:
        """
        Returns a build by id
        :param build_id: The id for the build whose status we are getting
        """
        build = self._redis_connection.get(build_id)
        if build is None:
            raise ItemNotFoundError('Invalid build id: {}.'.format(build_id))

        return build

    def add(self):
        """
        Add new build to collection
        """
        # Create build here
        build = # ...
        self._redis_connection.set(build_id, build)

```

We would also need a way to increment the `build_id` counter correctly and start it off at the last `build_id` each time we restart. This could be something like a lookup for the largest key value in our Redis store when ClusterRunner starts up and starting our counter from there.

This is something we don't really seem to get out of the box and would have to handle ourselves, which isn't a big deal but means more work.

### Other Things to Think About

Regardless of which approach we take, there are some universal things we need to solve and think about.

#### Serialization

A build is stored as a dictionary of information that looks something like this:

```python
{
    'num_subjobs': 0,
    'status': '',
    'details': 'null',
    'failed_atoms': 'null',
    'result': 'null',
    'num_atoms': 'null',
    'id': 1,
    'request_params': {
        'build_project_directory': '',
        'url': '',
        'job_name': '',
        'type': ''
    },
    'state_timestamps': {
        'queued': 1,
        'preparing': 1,
        'building': 'null',
        'error': 'null',
        'finished': 'null',
        'prepared': 'null',
        'canceled': 'null'
    },
    'error_message': 'null'
}
```

Both SQLite and Redis can't store nested dictionaries like this out of the box. One option we have is to serialize the entire dictionary as a string using `json.dumps` and then storing that string, and when we want to get the build back we can use `json.loads` to deserialize. This is kind of expensive and we should avoid doing it if a better exists.

One other idea is to decide what information from this dictionary is important and effectively reduce the amount of things we want to store. If there is any unneeded information here that we don't really care about in the context of persisting builds, leave a comment and we can try to reduce this data structure to something simpler.

### Conclusion and Final Thoughts

With all things considered, I don't think Redis is the right tool for the job here. If we use SQLite `ROWID`s for build `id`s, lookup queries can be as fast, if not faster than Redis lookup times. Writing builds is also very similar in terms of speed when it comes to Redis and SQLite. So when it comes to performance, I don't think there's much of a difference -- if anything, we might prefer SQLite.

Also, I think the point that Redis can't guarentee _all_ information will be written to disk is a huge issue with how we use ClusterRunner. If we're going to persist builds, we need to be able to persist _all_ the builds.

Especially with the set up overhead, SQLite seems like a more practical and easy to use solution. Less maintenance and configuration required with SQLite than Redis helps keep ClusterRunner as a tool that's easy to work with, easy to set up, and most importantly easy to maintain.

At the end of the day, Redis is great for temporary storage but we need a more permanent solution. SQLite seems like a very performant and practical tool for our use case, and I think we should go ahead with that approach.

nickzuber:
## Integrating SQLite w/ SQLAlchemy into ClusterRunner

This is done in an intentionally non intrusive way, changing very little about how cluster runner behaves but instead changing how we store builds within the program. 

All of the builds live in the `BuildStore` (see https://github.com/box/ClusterRunner/pull/402). When a build is created, two things happen:

 - The build is added to a local cache in the `BuildStore` with an `OrderedDict` the same way it was done when they were stored on the master.
 - The build is serialized and stored into the database in its current initial state.

When a build is created, it's initial `build_id` is set to `None` but it's id is then immediately set when it gets stored into the database. The database uses the `build_id` as a build's auto incrementing primary key, so this is automatically generated for us the moment we store it into the database (which is why we don't let CR create the `build_id`). 

When the cluster master service is stopped, the `BuildStore` performs a clean up process where it takes each build that is has locally and stores them into the database, updating any stored information regarding those builds. 

When a request is made to get a build, we first check the `BuildStore`'s local cache for the requested build. If was not found, we attempt to find it in the database. If it's found there, we create a `Build` object using all of the information we stored regarding that build, and return that (we also store this build in our local cache). 

### Connecting the Database 

Each instance of `BuildStore` has a connection to the database using SQLAlchemy. As of right now, there's only one connection (if there were multiple `BuildStore`s, then they would all be using the same connection. This would be easy to change to make it that each `BuildStore` instance has its own unique connection if we wanted to do that in the future. 

## Minor Changes

There are a few minor changes I've made while creating this feature that I want to point out:

#### Subjobs now know when they are completed

For some reason, a subjob doesn't know if it's completed without individually checking if all of the atoms are completed. Even when you call `Subjob.mark_completed`, it doesn't mark the subjob as completed but rather mark all its atoms as completed. I made a minor change to add a `completed` attribute.

This is important to storing builds because we want to be able to organize subjobs into their respective unstarted and completed queues correctly.

```python
class Subjob(object):
    def __init__(self, build_id, subjob_id, project_type, job_config, atoms):
        # ...
        self.completed = False

# ...

def mark_completed(self):
        self._set_atom_state(AtomState.COMPLETED)
        self.completed = True
```

## What's New

A few new files are introduced to the main app regarding how CR interacts with the database. We also moved the `BuildStore` to this directory as well.

```diff
  ├── app
  │   ├── client
  │   ├── common
+ │   ├── database
+ │   │   ├── build_store.py
+ │   │   ├── schema.py
+ │   │   └── session.py
  │   ├── master
- │   │   ├── build_store.py
  │   │   │ ...
  │   ├── slave
  │   │ ...
  │ ...
```

## Points of Concern & Things Left to Do

There are a few areas left that are either incomplete or definitely needs some revision. Especially with the `BuildStore` in general, it feels really verbose at the moment and I'm sure we can get it down and condense it better, but I just wanted to put this up as a WIP.

 - [x] **Updating a build when it completes**
Moved saving logic to Build class so we can simply just save to database when the postbuild tasks finish.<br /><details><summary>_Previous comments_</summary>I had trouble finding exactly where a build totally "completes" because of all the postbuild tasks that get run and how builds are run in general. Is there any point where the master knows a build has completed?</details><br />

 - [x] **Only reconstruct completed builds**
Now, if you try to fetch an incomplete build from from the database, you'll receive and error message telling you that it cannot produce the build because it was never completed.<br />
If you try to fetch a range of builds and one or more of those builds are incomplete, then they will just not appear in your request.<br /><details><summary>_Previous comments_</summary>
Right now, we store builds in _any_ state. This is because if the master service shuts down at any point during a build, we don't wait for the build to complete before we serialize it and store it into the database.<br />
As a first step, we should only care about completely builds and try to guarantee that we can correctly reproduce an old build we stored in the database if it's marked as completed in the database.<br />
So as of right now, if anyone asks for a build that's stored in the database, we try to rebuild and produce it regardless of what state its in. I don't think we can/should guarantee those uncompleted builds' accuracy just yet, so we shouldn't try to reconstruct them at all -- but rather produce some kind of error message.</details><br />

 - [ ] **Guaranteeing a build can be reconstructed totally correctly**
We might be able to have some leeway here since we only want to persist completed builds. I've been playing around with a bunch of different kinds of builds but I feel like I might not have a complete understanding on the kinds of customizations we can add to builds. We should try to make sure that we account for _everything_ that a build can have so we can make sure we know how to reconstruct one correctly and not get any unexpected errors from having made false assumptions about a particular build.<br />
So I would argue that guaranteeing a _complete_ 1:1 correspondence between a reconstructed build and its original form is out of the scope of this PR. Being able to create the same API representation should be what we're aiming for here.<br /><br />

 - [ ] **Range query performance**
Right now if you try to get a range of builds, each one is queried individually. On the scale of hundreds of builds, this performance hit is pretty noticeable. It's important to note that we'll only see this problem if a user actually tries to query hundreds of old builds. Which as of right now with the v1 `build`, can absolutely happen. Since we don't have any limitations on the amount of builds to return when a user hits the builds endpoint, this can be a huge issue once thousands of builds are stored. So, an easy fix would be to migrate to the v2 API _or_ add an upper bound on the amount of builds the v1 API can return.<br />
For reference, it takes approximately **0.94 seconds to load 100 builds** & **1.46 seconds to save 100 builds**.

#### AppVeyor tests failing

AppVeyor tests are failing because when the `tearDownClass` is invoked in tests that create test databases, it complains about how it can't remove the test databases because another process is using them.

#### New imports

~I've imported `sqlite3` and `sqlalchemy` for this branch, is there a proper way to install these packages with CR so that things like CI and other users can have them installed as well? Do I add this to `requirements.txt`?~

Closes #401


boxcla:

test/unit/test_main.py
```diff
@@ -46,6 +46,7 @@ def setUp(self):
 
     def test_master_args_correctly_create_cluster_master(self):
         mock_cluster_master = self.mock_ClusterMaster.return_value  # get the mock for the ClusterMaster instance
+        mock_cluster_master._build_store = _FakeBuildStore()
```
nickzuber: We need to manually add a _build_store because private member attributes aren't created by mock objects.

This is here because we call `master._build_store.clean_up()` when the master service shuts down -- we might be able to remove this later if we decide to move this logic somewhere else in the app

app/subcommands/master_subcommand.py
```diff
@@ -43,4 +43,6 @@ def async_run(self, port, log_level, eventlog_file):
 
         ioloop.start()  # this call blocks until the server is stopped
         ioloop.close(all_fds=True)  # all_fds=True is necessary here to make sure connections don't hang
+        self._logger.notice('About to shut down master service, saving all active builds to database.')
+        cluster_master._build_store.clean_up()  # pylint: disable=protected-access
```
nickzuber: This might not be in the right spot -- is there a better place to call this method when the master shuts down?

app/database/build_store.py
```diff
@@ -0,0 +1,520 @@
...
+    def _update_build(self, build: Build) -> int:
+        """
+        Serialize a Build object and update all of the parts to the database.
+        NOTE: These changes are not committed here. If you want these changes to persist,
+              make sure to commit the session afterwards.
+              We do selectively call commit a few times here but only after we delete rows.
+        :param build_id: The build_id of the build to update in the database.
+        """
+        build_id = build.build_id()
+        (
```
nickzuber: Not sure what we'd prefer to do with something like this formatting wise. Alternatively the method that returns this tuple could return a dict instead or something

app/master/cluster_master.py
```diff
@@ -296,7 +297,7 @@ def handle_result_reported_from_slave(self, slave_url, build_id, subjob_id, payl
         :rtype: str
         """
         self._logger.info('Results received from {} for subjob. (Build {}, Subjob {})', slave_url, build_id, subjob_id)
-        build = BuildStore.get(int(build_id))
+        build = self._build_store.get(int(build_id))
```
cmcginty: Not your code but why is this converted to an `int`. Is the type in the docstring wrong?

app/subcommands/master_subcommand.py
cmcginty: You might be able to use an exit handler to call this automatically.

test/framework/functional/functional_test_cluster.py
```diff
@@ -300,6 +306,12 @@ def kill_master(self):
         :return: The killed master service with return code, stdout, and stderr set.
         :rtype: ClusterController
         """
+        # Delete testing database after we're done
+        if isfile(TEST_DB_NAME):
+            os.remove(TEST_DB_NAME)
+        else:
```
cmcginty: If the tests are deterministic you should not need these conditionals. In case of a pre-mature test failure, you could just catch all errors in this method with at `try` block.

test/unit/database/test_build_store.py
```diff
@@ -0,0 +1,142 @@
...
+TEST_DB_URL = 'sqlite:///test.db'
+
+
+@genty
+class TestBuildStore(BaseUnitTestCase):
+    def setUp(self):
+        super().setUp()
+        self.patch('app.util.fs.create_dir')
+        self.patch('app.util.fs.async_delete')
+        self.patch('os.makedirs')
```
cmcginty: You probably don't want to patch `os` methods directly. Instead patch through the module you are trying to mock that imports `os`.

test/unit/database/test_build_store.py
```diff
@@ -0,0 +1,142 @@
...
+        Configuration['database_url'] = TEST_DB_URL
+        self._build_store = BuildStore()
+
+    def tearDown(self):
+        super().tearDown()
+        self.thread_pool_executor.shutdown()
+
+    def tearDownClass():
+        # Delete testing database after we're done
+        if isfile(TEST_DB_NAME):
```
cmcginty: Same comment as above for this conditional.

app/master/cluster_master.py
```diff
@@ -80,9 +80,10 @@ def get_builds(self, offset: int=None, limit: int=None) -> List['Build']:
         :param offset: The starting index of the requested build
         :param limit: The number of builds requested
         """
-        num_builds = BuildStore.size()
+        builds_in_cache, builds_in_db = self._build_store.size()
```
cmcginty: You might want to split this into separate calls for DB and cache size.

app/master/cluster_master.py
nickzuber: Nice catch, looks like we convert the `build_id` to an int before we ever pass it into this function.

e.g.

```python
self._cluster_master.handle_result_reported_from_slave(
            slave_url, int(build_id), int(subjob_id), file_payload[0])
```

app/database/database_setup.py
```diff
@@ -0,0 +1,124 @@
...
+    _atoms_query = """CREATE TABLE IF NOT EXISTS atoms (
+        uid INTEGER PRIMARY KEY,
+        atom_id INTEGER,
+        build_id INTEGER,
+        subjob_id INTEGER,
+        command_string TEXT,
+        expected_time REAL,
+        actual_time REAL,
+        exit_code INTEGER,
+        state TEXT
```
josephharrington: Is there a reason to do explicit create table statements like this instead of doing a `Base.metadata.create_all` call?

app/database/connection.py
```diff
@@ -0,0 +1,13 @@
...
+
+class Connection():
+    _Session = None
+
+    @classmethod
+    def get(cls, url):
+        if cls._Session is None:
+            engine = create_engine(url, echo=False)
+            cls._Session = sessionmaker(bind=engine)
+        return cls._Session()
```
josephharrington: You might consider a separate initialization method that takes the url parameter, and that can only be called once. Something like
```python
@classmethod
def init(cls, url):
    if cls._Session is None:
        engine = create_engine(url, echo=False)
        cls._Session = sessionmaker(bind=engine)
    else:
        raise Exception('Init can only be called once!')

@classmethod
def get(cls):
    return cls._Session()
```
I like that pattern better because it makes it clearer that the `url` param is only used once, and would also mean you wouldn't have to pass the db url string all around the app.

app/database/database_setup.py
```diff
@@ -0,0 +1,124 @@
...
+        sqlite_cursor.execute(cls._failed_subjob_atom_pairs_query)
+        sqlite_cursor.execute(cls._build_requests_query)
+        sqlite_cursor.execute(cls._build_fsm_query)
+        sqlite_cursor.execute(cls._subjobs_query)
+        sqlite_cursor.execute(cls._atoms_query)
+        sqlite_connection.commit()
+        sqlite_connection.close()
+
+    @classmethod
+    def reset(cls):
```
josephharrington: Is it useful to have this? Why not just delete the db file?

app/database/database_setup.py
```diff
@@ -0,0 +1,124 @@
+import sqlite3
+
+from app.util.conf.configuration import Configuration
+
+
+class FailedSQLiteTableSetup(Exception):
+    pass
+
+
+class DatabaseSetup:
```
josephharrington: This does not seem like something that should be a class. I'm not sure we need this at all since sqlalchemy can do creates, but if it is needed I'd rather the methods in this class just be functions inside the module (not wrapped in a class).

app/database/build_store.py
```diff
@@ -0,0 +1,537 @@
...
+        build_id = build_schema.build_id
+
+        # Build Meta
+        build_meta = BuildMetaSchema(
+            build_id=build_id,
+            artifacts_tar_file=build._artifacts_tar_file,
+            artifacts_zip_file=build._artifacts_zip_file,
+            error_message=build._error_message,
+            postbuild_tasks_are_finished=bool(build._postbuild_tasks_are_finished),
+            timing_file_path=build._timing_file_path,
```
josephharrington: Some of these things (especially the paths) seem like they could be computed from the build id and don't need to be saved.

app/database/schema.py
```diff
@@ -0,0 +1,102 @@
...
+
+class BuildStateSchema(Base):
+    """Describes the state of a Build"""
+    __tablename__ = 'builds'
+
+    build_id = Column(Integer, primary_key=True, autoincrement=True)
+    completed = Column(Boolean)
+
+
+class BuildMetaSchema(Base):
```
josephharrington: Why are BuildStateSchema/BuildMetaSchema/BuildArtifactSchema/etc. separate? There are several tables whose rows have a with a one-to-one correspondence with a build -- I'm interested in why you chose to split them apart.

I'm still working through reviewing this and adding comments, but here's some interesting reading in the meantime 😉:
https://martinfowler.com/bliki/AnemicDomainModel.html

app/database/schema.py
```diff
@@ -0,0 +1,102 @@
...
+    uid = Column(Integer, primary_key=True, autoincrement=True)
+    build_id = Column(Integer)
+    failed_artifact_directory = Column(String)
+
+
+class FailedSubjobAtomPairsSchema(Base):
+    """Describes the list of _failed_subjob_atom_pairs (List[int, int]) from BuildArtifact"""
+    __tablename__ = 'failed_subjobs_atom_pairs'
+
+    uid = Column(Integer, primary_key=True, autoincrement=True)
```
josephharrington: In general I'd go with `id` over `uid`. (Did you avoid `id` because it's a keyword?)

app/database/schema.py
```diff
@@ -0,0 +1,102 @@
...
+
+    build_id = Column(Integer, primary_key=True)
+    state = Column(String)
+    queued = Column(Float)
+    finished = Column(Float)
+    prepared = Column(Float)
+    preparing = Column(Float)
+    error = Column(Float)
+    canceled = Column(Float)
+    building = Column(Float)
```
josephharrington: These are the timestamps, right? I'd append `_ts` to the column names.

app/subcommands/master_subcommand.py
josephharrington: Yeah, we'd want something like this implemented in its own teardown callback so that it's executed explicitly as part of graceful shutdown (and not affected, for example, by ioloop raising an exception). Search for occurrences of `add_teardown_callback`.

app/util/conf/base_config_loader.py
```diff
@@ -162,6 +166,11 @@ def _get_config_file_whitelisted_keys(self):
             'default_http_timeout',
             'https_cert_file',
             'https_key_file',
+            'pagination_offset',
+            'pagination_limit',
+            'pagination_max_limit',
```
josephharrington: Do the pagination params need to be whitelisted?

app/database/database_setup.py
nickzuber: Hmm that'd do the trick

app/util/conf/base_config_loader.py
nickzuber: Oh I was under the impression we'd want to be able to customize them in the config. I can remove them from this list

app/database/database_setup.py
nickzuber: Yeah 😅 I made this class before we started using SQLAlchemy and didn't realize you could create tables with it, we can definitely do away with this entire class

app/database/schema.py
nickzuber: Actually yeah that's exactly why, does that not matter in this context? I didn't want to accidentally override some built-in method

app/database/schema.py
nickzuber: I originally split them apart when I was breaking down all the attributes of a build -- I'm sure there's tons of room for improvement on how these tables are structured (for example, condensing tables like these into one). Originally it was just a little more clear where the values were coming from and where they belong

app/database/schema.py
josephharrington: Yeah we've used `id` as a db field in other services and it hasn't caused any issues. Usages of it would nearly always be namespaced in an instance or class reference anyway so it wouldn't override the builtin.

app/database/connection.py
nickzuber: Actually after thinking about it more, I don't think we should disallow the ability to create new sessions. I do like and support the idea of separating the `get` into an `init` or `create` for actually creating the session, but we should be able to create a new session if we explicitly want to

nickzuber:
> **Updating a build when it completes**
I had trouble finding exactly where a build totally "completes" because of all the postbuild tasks that get run and how builds are run in general. Is there any point where the master knows a build has completed?

So it looks like the best way to figure out when a build has completely finished (including postbuild tasks) is to handle it within the build itself. We could move the build serialization logic out of the BuildStore and into the Build, basically adding a `Build.save` method which would serialize and store the build in its current state into the database.

How does this sound? This would mean that both the BuildStore and the Build will be dealing with the database. I personally think this is fine and kind of makes sense (it makes sense for the build to own its serialization process), but I'm not particularly crazy about the idea of the Build also talking to the database. It shouldn't be a big deal but I'd like the database logic to be isolated in the BuildStore -- it doesn't seem like we can get away with that though in the end. 

Thoughts?

app/database/connection.py
josephharrington: I'm not sure what you meant by this comment. When were we talking about disallowing the ability to create new sessions?

josephharrington:
@nickzuber I'm about to go through the serializing stuff you added in detail, but your comment about having the Build own its own serialization is along the lines of what I was generally going to suggest.

> ...I'm not particularly crazy about the idea of the Build also talking to the database
I'm not sure what you mean by "talking to the database". If you mean that the Build class is interacting with model objects then I think that's fine. It's better for objects to know how they themselves are stored and loaded than for some other single class to know how every object type in the app is stored and loaded.

If we had done this from scratch, we probably would have made the Build object and its db model representation one and the same, so saving to the db would be automatic and integrated with the Build class itself. This is what I was getting at by posting the link to the "Anemic Domain Model" post. I prefer that approach (making the Build class be its own db model class) in general but that's probably too heavy to do in a single change -- incremental is better.

I'll probably give briefer feedback on your current BuildStore code, since I'm assuming that you'll pursue this idea of moving some of that into the Build class.

tjlee0909:
For relatively simple persistence use-cases (like the one we are introducing now into ClusterRunner), I'm all for just having the entity know about its persistence backing (just having all of the DB-model stuff in the Build class).

If we were in WebApp-land, then I totally understand the need for intermediate classes (and even entirely separate services, such as Credence), but we aren't.

That said, one of the appeals of the BuildStore abstraction was that we could swap it with a trivial in-memory dictionary implementation (what we have right now, except behind the BuildStore abstraction) -- this way we could test incrementally by first making sure that the abstraction holds before actually swapping the concrete implementation.

@nickzuber Could you elaborate a little bit more about why we need to break the abstraction due to it being difficult to figure out whether or not a build is actually completed?

nickzuber:
@tjlee0909 Is the master service ever notified when a build completes and which build it is? Basically the master service doesn't know _when_ to tell the BuildStore to update a build after it's been completed since the master service doesn't know when a build finishes.

I think giving the Build a `save` method that does this for us when it completes its postbuild tasks makes the most sense. In the BuildStore we can just call `build.save()` instead of updating the build internally in the BuildStore, so not much changes there.

@josephharrington
I'm about to push a new commit that implements this idea so you guys can take a look.

Also

>That said, one of the appeals of the BuildStore abstraction was that we could swap it with a trivial in-memory dictionary implementation (what we have right now, except behind the BuildStore abstraction) -- this way we could test incrementally by first making sure that the abstraction holds before actually swapping the concrete implementation.

Are you talking about https://github.com/box/ClusterRunner/pull/402 here?

app/database/schema.py
```diff
@@ -0,0 +1,73 @@
...
+    completed = Column(Boolean)
+
+
+class AtomsSchema(Base):
+    """Describes the subjobs associated with a Build"""
+    __tablename__ = 'atoms'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    atom_id = Column(Integer)
+    build_id = Column(Integer)
```
josephharrington: If these are going to be separate tables, the build_id column should be indexed since that's what you're pretty much always querying by.

Also, consider adding foreign key relationships between the related tables -- they can simplify the query code, make saves easier, etc.
http://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html
http://docs.sqlalchemy.org/en/latest/orm/cascades.html

app/database/build_store.py
```diff
@@ -0,0 +1,307 @@
...
+        q_build_atoms = session.query(AtomsSchema).filter(AtomsSchema.build_id == build_id).all()
+        q_build_subjobs = session.query(SubjobsSchema).filter(SubjobsSchema.build_id == build_id).all()
+
+        # If a query returns None, then we know the build wasn't found in the database
+        if not q_build:
+            return None
+
+        # Build wasn't completed when we stored it
+        if not allow_incompleted_builds and not bool(int(q_build.completed)):
+            raise IncompleteBuild('Cannot load build (id: {}) because it was never completed.'.format(build_id))
```
josephharrington: Again, I would rather have this behavior:
- During shutdown any builds currently in progress are failed with a message like "Master shut down while build was executing."
- Those builds enter the error state. They're loaded just like any other finished build after the master restarts.
- We don't add any code right now around supporting or preserving in-progress builds.

In my mind, attempting to preserve the state of in-progress builds when the master restarts adds a bunch of complexity with no benefit. Maybe in the future we'll decide to support resuming builds after the master restarts, but the changes needed for that are much more extensive than even what you have here. I'm also not convinced that we'll ever do that -- it would be a complicated feature and the benefit we'd get out of it is not very significant.

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
...
+        q_failed_artifact_directories = session.query(FailedArtifactDirectoriesSchema).filter(FailedArtifactDirectoriesSchema.build_id == self._build_id).all()
+        q_failed_subjob_atom_pairs = session.query(FailedSubjobAtomPairsSchema).filter(FailedSubjobAtomPairsSchema.build_id == self._build_id).all()
+        q_build_atoms = session.query(AtomsSchema).filter(AtomsSchema.build_id == self._build_id).all()
+        q_build_subjobs = session.query(SubjobsSchema).filter(SubjobsSchema.build_id == self._build_id).all()
+
+        # If this wasn't found, it's safe to assume that the build doesn't exist within the database
+        if q_build is None:
+            raise ItemNotFoundError('Unable to find build (id: {}) in database.'.format(self._build_id))
+
+        q_build.completed = self._status() == BuildState.FINISHED
```
josephharrington: I don't think the db should have any computed fields. The db should simply be a copy of the Build instance state.

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
...
+        q_build.error_message = self._error_message
+        q_build.postbuild_tasks_are_finished = self._postbuild_tasks_are_finished
+        q_build.setup_failures = self.setup_failures
+        q_build.timing_file_path = self._timing_file_path
+
+        build_artifact_dir = None
+        if self._build_artifact is not None:
+            build_artifact_dir = self._build_artifact.build_artifact_dir
+
+        q_build.build_artifact_dir = build_artifact_dir
```
josephharrington: If this is a nullable field can you just do
```python
if self._build_artifact:
    q_build.build_artifact_dir = self._build_artifact.build_artifact_dir
```

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
...
+            # Clear all old directories
+            session.query(FailedArtifactDirectoriesSchema) \
+                .filter(FailedArtifactDirectoriesSchema.build_id == self._build_id) \
+                .delete()
+
+            # Commit changes so we don't delete the newly added rows later
+            session.commit()
+
+            # Add all the updated versions of the directories
+            for directory in self._build_artifact._get_failed_artifact_directories():  # pylint: disable=protected-access
```
josephharrington: Private access :( -- how do you feel about letting the BuildArtifact class (and other table-related classes) store/load its own state?

On a higher level, how do you feel about not storing any internals of the BuildArtifact class at all? Its constructor only takes in a directory on disk, and it seems like all of its internal state can be computed based on that. (Maybe it would take some refactoring to get that working, but that's not surprising for this change. I'd rather do minor refactoring of the code than add a bunch of extra columns to the db.)

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
...
+        if self._build_artifact is not None:
+            build_artifact_dir = self._build_artifact.build_artifact_dir
+
+        q_build.build_artifact_dir = build_artifact_dir
+
+        if self._build_artifact is not None:
+            # Clear all old directories
+            session.query(FailedArtifactDirectoriesSchema) \
+                .filter(FailedArtifactDirectoriesSchema.build_id == self._build_id) \
+                .delete()
```
josephharrington: Is this needed?
1. Failed artifacts shouldn't change once they're set, right?
2. Even beyond that, it looks like the BuildArtifact class already populates this from what is on disk. It seems like we can just rely on what's on disk instead of trying to store it in the db.

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
...
+        q_build.preparing_ts = fsm_timestamps['preparing']
+        q_build.error_ts = fsm_timestamps['error']
+        q_build.canceled_ts = fsm_timestamps['canceled']
+        q_build.building_ts = fsm_timestamps['building']
+
+        # Subjobs
+        # Clear all old Subjobs and Atoms
+        session.query(SubjobsSchema) \
+            .filter(SubjobsSchema.build_id == self._build_id) \
+            .delete()
```
josephharrington: Did you save atoms and subjobs this way (delete+create) because its simpler than updating any rows already there?

app/database/build_store.py
```diff
@@ -0,0 +1,307 @@
...
+                    exit_code=atom.exit_code,
+                    state=atom.state,
+                )
+                session.add(atom_schema)
+
+        session.commit()
+        return build_id
+
+    @classmethod
+    def _reconstruct_build(cls, build_id, allow_incompleted_builds=False) -> Build:
```
josephharrington: What do you think of having a `Build.load_from_db` classmethod instead?

app/database/build_store.py
```diff
@@ -0,0 +1,307 @@
...
+        build._finished_subjobs = Queue(maxsize=len(subjobs))
+        for _, subjob in subjobs.items():
+            if subjob.completed:
+                build._finished_subjobs.put(subjob)
+            else:
+                build._unstarted_subjobs.put(subjob)
+
+        # Spend preparation coin if this build is already completed
+        if q_build.completed:
+            build._preparation_coin.spend()
```
josephharrington: Overall this reconstruction seems complex and is messing around with the internals of a Build object that nothing besides the Build class itself should really be messing with. There seems to be a lot here around recreating the state of in-progress builds, which is added complexity that I think we should avoid on this first pass (if not forever).

Refactoring the Build class is already very hard because of its complexity (and so it _should_ be refactored to make it simpler). I don't want to make it harder to refactor, but I think the amount of internals touched by this persistence logic will make it much harder.

Really, the main requirement is to allow users to get the state of previous builds after the master is restarted. To me, that means the only data the database needs to contain is what is exposed by the API. We don't need to recreate a Build object exactly as it existed before the restart -- we just need to make sure the data that users are interested in is available.

app/database/build_store.py
```diff
@@ -0,0 +1,439 @@
...
+from app.master.build import Build
+from app.master.build_request import BuildRequest
+from app.master.build_fsm import BuildState
+from app.master.subjob import Subjob
+from app.util.exceptions import ItemNotFoundError
+from app.util.log import get_logger
+
+
+# pylint: disable=protected-access
+class BuildStore:
```
nickzuber: I thought the point of the BuildStore was to move the builds off of the master? If the master has an instance of the BuildStore, does that not defeat the purpose? I like the idea of the BuildStore being in charge of the storing the builds and it being a separate service, rather than something that the master itself has an instance of. Like if we wanted to use the BuildStore in other places in the app it'd be nice to be able to import it and use it, rather than importing the master and then trying to access its instance of the BuildStore.

Also this is how it was proposed in https://github.com/box/ClusterRunner/pull/402, what changed your mind on its design?

app/database/build_store.py
```diff
@@ -0,0 +1,439 @@
...
+
+    @classmethod
+    def get(cls, build_id: int, allow_incompleted_builds=False) -> Optional[Build]:
+        """
+        Returns a build by id.
+        :param build_id: The id for the build whose status we are getting
+        """
+        build = cls._cached_builds_by_id.get(build_id)
+        if build is None:
+            cls._logger.info('Requested build (id: {}) was not found in cache, checking database.'.format(build_id))
```
nickzuber: What's "debug level"?

app/database/build_store.py
```diff
@@ -0,0 +1,439 @@
...
+            build = cls._reconstruct_build(build_id, allow_incompleted_builds=allow_incompleted_builds)
+            if build is not None:
+                cls._cached_builds_by_id[build_id] = build
+                cls._logger.notice('Build (id: {}) was added to cache.'.format(build_id))
```
nickzuber: Some of these logging messages can get really noising, I wouldn't be opposed to removing them 👍

app/database/build_store.py
```diff
@@ -0,0 +1,439 @@
...
+        """
+        Add new build to collection.
+        :param build: The build to add to the store.
+        """
+        build_id = cls._store_build(build)
+        build._build_id = build_id
+        cls._cached_builds_by_id[build_id] = build
+
+    @classmethod
+    def save(cls, build: Build):
```
nickzuber: Nothing -- this isn't used anywhere. I've removed it since this commit, my bad!

app/database/build_store.py
```diff
@@ -0,0 +1,307 @@
...
+        """
+        Save current state of all cached builds.
+        """
+        session = Connection.get()
+        cls._logger.notice('Saving all active builds to database...')
+        for build_id in cls._cached_builds_by_id:
+            build = cls._cached_builds_by_id[build_id]
+            build.save()
+        session.commit()
+        cls._logger.notice('...done')
```
nickzuber: These logging messages were more for my sake of making sure everything is working okay -- more than happy to clean them up

app/database/build_store.py
nickzuber: Oh good catch, I don't think we'd ever use this method anywhere -- we could remove it

app/master/build.py
nickzuber: I did it this way because I wasn't sure if we're guaranteed to _always_ have the same amount of subjobs and/or atoms for a build. Maybe this assumption is wrong -- in which case we should just update the subjobs/atoms columns instead of delete+create.

Is my assumption wrong?

app/master/build.py
nickzuber: Same reason as https://github.com/box/ClusterRunner/pull/406#discussion_r134908476

Yeah there's some things related to builds that I'm not very familiar with, so I'm happy to take suggestions on which things we can do away with persisting (or maybe some things we need to add).

app/master/build.py
nickzuber: Sounds good, we can always query the completed builds by their status rather than just a "completed" column

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
         """
         return os.path.join(Configuration['build_symlink_directory'], str(uuid.uuid4()))
 
+    def save(self):
+        """Serialize the Build object and update all of the parts to the database."""
+        session = Connection.get()
+
+        q_build = session.query(BuildSchema).filter(BuildSchema.build_id == self._build_id).first()
```
nickzuber: I have no strong opinion about the naming convention here, I'd prefer for the sqlalchemy models to be clearly named though so it's obvious what they are. Open to suggestions!

app/database/schema.py
nickzuber: > the build_id column should be indexed

What do you mean by indexed? More than one subjob and atom will have the same build_id, which is why I didn't use that as its primary key (unless this isn't what you're getting at)

app/database/build_store.py
nickzuber: I point this issue out in the "Points of Concern & Things Left to Do" section of this PR description, I think we need to redefine our goal here with what we consider a reconstructed build

app/master/build.py
```diff
@@ -564,6 +569,121 @@ def _generate_unique_symlink_path_for_build_repo(self):
         """
         return os.path.join(Configuration['build_symlink_directory'], str(uuid.uuid4()))
 
+    def save(self):
```
nickzuber: `save` assumes that this build already exists in the database and we're just updating its values.

`_store_build` is creating this build and storing it for the first time.

They both work slightly differently -- `save` queries the database using SQLAlchemy for the model and `_store_build` creates the model from the build. Maybe there's a way to condense these?

app/master/build.py
nickzuber: As long as things like ```self._failed_artifact_directories``` and ```self._failed_subjob_atom_pairs``` can be loaded correctly again from doing something like

```python
# build.py

timing_data = self._read_subjob_timings_from_results()
self._create_build_artifact(timing_data)
```

We may have to jump through a few more hoops to initialize it but I agree -- I'd much rather refactor some code than have to store more things in the database

app/database/build_store.py
```diff
@@ -0,0 +1,439 @@
...
+        Returns a list of all builds.
+        :param start: The starting index of the requested build.
+        :param end: 1 + the index of the last requested element, although if this is greater than the total number
+                    of builds available the length of the returned list may be smaller than (end - start).
+        """
+        builds = []
+        # Add 1 to start & end because we're create build_id's, not indices
+        for build_id in range(start + 1, end + 1):
+            try:
+                builds.append(cls.get(build_id, allow_incompleted_builds=allow_incompleted_builds))
```
nickzuber: Originally I thought we could just do a range query, but then I realized for each build we'd still need to get all its subjobs and atoms, which would require additional queries.

So even with a range query on builds tables, we'd still need to individually query all the atoms and subjobs for each build -- so in the end we don't save much time querying. Is there a canonical way to solve this problem? I feel like there must be, I'll try to look around to see if I can find anything

app/database/build_store.py
```diff
@@ -0,0 +1,307 @@
...
+    def clean_up(cls):
+        """
+        Save current state of all cached builds.
+        """
+        session = Connection.get()
+        cls._logger.notice('Saving all active builds to database...')
+        for build_id in cls._cached_builds_by_id:
+            build = cls._cached_builds_by_id[build_id]
+            build.save()
+        session.commit()
```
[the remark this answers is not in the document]
nickzuber: Much better 👍 cool use of generators too

app/database/build_store.py
nickzuber: > During shutdown any builds currently in progress are failed with a message like "Master shut down while build was executing."

Could you point to an example of failing a build with an error message?

app/database/build_store.py
nickzuber: That could work, I think it makes more sense for the Build class to be responsible for serializing and deserializing a Build object, so I support this

app/database/build_store.py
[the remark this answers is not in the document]
josephharrington: Sqlalchemy lets you define a schema such that queries for related tables are automatically loaded without needing to query explicitly. I kind of hinted at it in other comments, but the relevant docs are:
http://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html
http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.lazy

Using relationships would allow you to only do an explicit query for the BuildSchema object, and the related queries for subjobs/atoms would happen automatically (immediately or as needed, depending on the `lazy` setting).

app/database/build_store.py
josephharrington: https://github.com/box/ClusterRunner/blob/master@%7B2017-08-24%7D/app/master/build_request_handler.py#L102

Just ignore my #WIP notes. I left a mess. :)

app/database/schema.py
josephharrington: http://www.sqlitetutorial.net/sqlite-index/
https://sqlite.org/queryplanner.html
http://docs.sqlalchemy.org/en/latest/core/constraints.html#indexes

Primary key columns get indexed automatically. But if you query on a column without an index then every row in the table will be examined as part of the query (a "full table scan").

app/master/build.py
josephharrington: You're right that the number of subjobs and atoms for a build can vary across builds, but once subjobs/atoms have been generated for a given build they won't change.

So the more complex but (theoretically) less IO-intensive way would be to 
1. query for all the subjobs saved to the db
2. iterate through each and update with the current state (match up Subjob N with SubjobSchema N and copy the state over)
3. commit the changes, sqlalchemy will determine if the state of a subjob/atom row has actually changed and if anything needs to be written

Basically the idea is to update a row for a subjob or atom if it already exists, instead of deleting it and recreating it. I imagine this could drastically reduce the amount of time it takes to save a build in the case when nothing has changed about the subjobs/atoms since the last time the build was saved.

app/master/build.py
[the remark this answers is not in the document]
josephharrington: 👍 Maybe I am missing some caveat, but it seems like they should at least be able to share some code around setting the model fields.

josephharrington:
Closing for now, but we should reference this if we revisit build persistence.

PR #407, opened 2017-08-23
Add functional tests for build cancel

shriganeshs:
This PR adds functional test for build cancellation
- while atomizing (QUEUED state)
- while building (BUILDING state)

### DIFF
--- a/test/functional/job_configs.py
+++ b/test/functional/job_configs.py
@@ -174,6 +174,46 @@ BasicSleepingJob:
     expected_num_atoms=5,
 )
 
+# This is a very basic job where each atom (total 90) just sleeps for 1 second.
+JOB_WITH_SLEEPS_90SEC = FunctionalTestJobConfig(
+    config={
+        'posix': """
+SleepingJob90Sec:
+    commands:
+        - sleep 1
+    atomizers:
+        - TOKEN: seq 0 90 | xargs -I {} echo "This is atom {}"
+
+""",
+        'nt': """
+SleepingJob90Sec:
+    commands:
+        - timeout 1 > NUL
+    atomizers:
+        - TOKEN: FOR /l %n in (0,1,90) DO @echo This is atom %n
+""",
+    },
+    expected_to_fail=False,
+    expected_num_subjobs=90,
+    expected_num_atoms=90,
+)
+
+# This is a very basic job where each atom (total 90) just sleeps for 1 second.
+JOB_WITH_SLEEPING_ATOMIZER_90SEC = FunctionalTestJobConfig(
+    config={
+        'posix': """
+SleepingAtomizerJob90Sec:
+    commands:
+        - echo $ATOM
+    atomizers:
+        - ATOM: echo "$$" > /tmp/atomizer_pid.txt; for x in $(seq 1 90); do sleep 1 && echo $x; done
+""",
+    },
+    expected_to_fail=False,
+    expected_num_subjobs=90,
+    expected_num_atoms=90,
+)
+
 FAILING_SETUP_JOB = FunctionalTestJobConfig(
     config={
         'posix': """
--- a/test/functional/master/test_build_cancellation.py
+++ b/test/functional/master/test_build_cancellation.py
@@ -1,0 +1,126 @@
+import os
+import subprocess
+import tempfile
+import time
+from unittest import skipIf
+import yaml
+
+from app.master.build_fsm import BuildState
+from app.util import poll
+from app.util.process_utils import is_windows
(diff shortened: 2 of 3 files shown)


boxcla:

test/functional/master/test_build_cancellation.py
```diff
@@ -0,0 +1,66 @@
...
+            'config': yaml.safe_load(JOB_WITH_SLEEPING_ATOMIZER_90SEC.config[os.name])['SleepingAtomizerJob90Sec'],
+            'project_directory': project_dir.name,
+        })
+        build_id = build_resp['build_id']
+        self.assertTrue(master.block_until_build_has_status(build_id, ['QUEUED'], timeout=30),
+                        'The build should be in queued state within the timeout.')
+        self.assert_build_status_contains_expected_data(build_id, {'status': 'QUEUED'})
+        # Sleep for 5 sec to let CR-master start the atomization process
+        time.sleep(5)
+        master.cancel_build(build_id)
```
shriganeshs: I tried to programmatically get the `pid` of `atomizer` subprocess, to check if its being killed within the given timeout after build is canceled. But I think it needs more effort. I will add that as a TODO.

test/functional/job_configs.py
```diff
@@ -174,6 +174,47 @@ def __init__(self, config, expected_to_fail, expected_num_subjobs, expected_num_
...
+
+# This is a very basic job where each atom (total 90) just sleeps for 1 second.
+JOB_WITH_SLEEPING_ATOMIZER_90SEC = FunctionalTestJobConfig(
+    config={
+        'posix': """
+SleepingAtomizerJob90Sec:
+    commands:
+        - echo $ATOM
+    atomizers:
+        - ATOM: for x in $(seq 1 90); do sleep 1 && echo $x; done
```
josephharrington: No config for windows?

test/functional/master/test_build_cancellation.py
josephharrington: I was thinking that an alternate way to do this test (which would not require the `sleep(5)`) would be to make the atomizer something like:

`- ATOM: touch /tmp/atomizer-is-executing; for x in $(seq 1 90); do sleep 1 && echo $x; done`
Then the test could block until that /tmp/atomizer-is-executing file exists.

You could also do something like `echo "$$" > /tmp/atomizer-pid`, then have the test validate the atomizer process is killed, but I'm not sure if there's a good way to do that on windows.

My main concern with sleeps is that they can cause flaky test failures if the test runs slowly for some reason, and they slow down the test if the test would otherwise run quickly.

test/functional/master/test_build_cancellation.py
```diff
@@ -0,0 +1,66 @@
...
+        # Start a slave even though its not needed as build is canceled during atomization
+        slaves = self.cluster.start_slaves(1, num_executors_per_slave=1, start_port=43001)
+        project_dir = tempfile.TemporaryDirectory()
+        build_resp = master.post_new_build({
+            'type': 'directory',
+            'config': yaml.safe_load(JOB_WITH_SLEEPING_ATOMIZER_90SEC.config[os.name])['SleepingAtomizerJob90Sec'],
+            'project_directory': project_dir.name,
+        })
+        build_id = build_resp['build_id']
+        self.assertTrue(master.block_until_build_has_status(build_id, ['QUEUED'], timeout=30),
```
nickzuber: Doesn't really change anything but should we be using `BuildState` here instead of typing out the string like `QUEUED`?

test/functional/master/test_build_cancellation.py
shriganeshs: Done.

test/functional/master/test_build_cancellation.py
shriganeshs: Done

test/functional/master/test_build_cancellation.py
```diff
@@ -0,0 +1,66 @@
...
+import time
+from unittest import skipIf
+import yaml
+
+from app.util.process_utils import is_windows
+from test.framework.functional.base_functional_test_case import BaseFunctionalTestCase
+from test.functional.job_configs import JOB_WITH_SLEEPS_90SEC, JOB_WITH_SLEEPING_ATOMIZER_90SEC
+
+
+@skipIf(is_windows(), 'Fails on AppVeyor; see issue #345')
```
shriganeshs: I intentionally kept the "skip" as the test is very similar to the one I copied from. But lets discuss this. Same comment for the not having `job config` for Windows as I was planning to skip tests on Windows.

PR #409, opened 2017-09-21
Fix SSL cacerts not found when running frozen CR

### LINKED ISSUE #408: SSL cacerts cannot be found when running frozen ClusterRunner package
Following exception is raised when trying to run slave with frozen CR package.

```
[2017-09-21 12:18:27.978] 25979 WARNING SlaveTornadoThr decorators      Call to Network.post raised SSLError("[Errno 20] Not a directory"). Retrying in 12.8 seconds.
[2017-09-21 12:18:40.795] 25979 ERROR   SlaveTornadoThr tornado.applica Exception in callback functools.partial(<function wrap.<locals>.wrapped at 0x7f642dbb32f0>)
  File "/home/sshintre/ClusterRunnerBuild/venv-cr/lib/python3.4/site-packages/requests/packages/urllib3/util/ssl_.py", line 115, in ssl_wrap_socket
NotADirectoryError: [Errno 20] Not a directory

During handling of the above exception, another exception occurred:

  File "/home/sshintre/ClusterRunnerBuild/venv-cr/lib/python3.4/site-packages/requests/adapters.py", line 327, in send
  File "/home/sshintre/ClusterRunnerBuild/venv-cr/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py", line 493, in urlopen
  File "/home/sshintre/ClusterRunnerBuild/venv-cr/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py", line 291, in _make_request
  File "/usr/lib64/python3.4/http/client.py", line 1137, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib64/python3.4/http/client.py", line 1182, in _send_request
    self.endheaders(body)
  File "/usr/lib64/python3.4/http/client.py", line 1133, in endheaders
    self._send_output(message_body)
  File "/usr/lib64/python3.4/http/client.py", line 963, in _send_output
    self.send(msg)
  File "/usr/lib64/python3.4/http/client.py", line 898, in send
    self.connect()
  File "/home/sshintre/ClusterRunnerBuild/venv-cr/lib/python3.4/site-packages/requests/packages/urllib3/connection.py", line 190, in connect
  File "/home/sshintre/ClusterRunnerBuild/venv-cr/lib/python3.4/site-packages/requests/packages/urllib3/util/ssl_.py", line 119, in ssl_wrap_socket
requests.packages.urllib3.exceptions.SSLError: [Errno 20] Not a directory

```

shriganeshs:
SSL certificates cannot be found when running frozen CR package.
resolves #408

### DIFF
--- a/app/util/conf/base_config_loader.py
+++ b/app/util/conf/base_config_loader.py
@@ -1,5 +1,5 @@
-from os.path import dirname, join, realpath, expanduser, isdir
-from os import chmod
+from os.path import dirname, join, realpath, expanduser, isdir, isfile
+from os import chmod, environ
 import platform
 import sys
 import shutil
@@ -125,6 +125,8 @@ class BaseConfigLoader(object):
         # Set protocol scheme (by default it is set to 'http')
         if Configuration['https_cert_file'] and Configuration['https_key_file']:
             conf.set('protocol_scheme', 'https')
+            # Workaround to fix the SSL cacerts not found issue if running CR frozen package.
+            self._override_cacerts()
 
     def load_from_config_file(self, config, config_filename):
         """
@@ -134,6 +136,21 @@ class BaseConfigLoader(object):
         self._load_section_from_config_file(config, config_filename, BASE_CONFIG_FILE_SECTION)
         if self.CONFIG_FILE_SECTION:
             self._load_section_from_config_file(config, config_filename, self.CONFIG_FILE_SECTION)
+
+    def _override_cacerts(self):
+        """
+        Override path for requests certs if this is a frozen executable run.
+        SSL cacerts cannot be found when running frozen CR package. Workaround the issue by bundling the cacert.pem
+        file into frozen package and at startup update the environment to point to these cacerts.
+        """
+        if getattr(sys, 'frozen', False):
+            root_directory = dirname(sys.executable)
+            cacert_file = '{}/cacert.pem'.format(root_directory)
+            if isfile(cacert_file):
+                environ['REQUESTS_CA_BUNDLE'] = cacert_file
+            else:
+                print('Missing SSL cacerts file {} in the frozen ClusterRunner package.'.format(cacert_file))
+                sys.exit(1)
 
     def _get_config_file_whitelisted_keys(self):
         """
--- a/app/util/conf/base_config_loader.py
+++ b/app/util/conf/base_config_loader.py
@@ -1,5 +1,6 @@
+import errno
 from os.path import dirname, join, realpath, expanduser, isdir, isfile
-from os import chmod, environ
+from os import chmod, environ, strerror
 import platform
 import sys
 import shutil
@@ -149,8 +150,7 @@ class BaseConfigLoader(object):
             if isfile(cacert_file):
                 environ['REQUESTS_CA_BUNDLE'] = cacert_file
             else:
-                print('Missing SSL cacerts file {} in the frozen ClusterRunner package.'.format(cacert_file))
-                sys.exit(1)
+                raise FileNotFoundError(errno.ENOENT, strerror(errno.ENOENT), cacert_file)
 
     def _get_config_file_whitelisted_keys(self):
         """
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,8 @@
 # Install with "pip install -r requirements.txt"
 
+# Downgrade the setuptools version to workaround the issue replacing the setuptools dependency
+# IOError METADATA no such file. Check https://github.com/pypa/setuptools/issues/951 for more details.
+setuptools==33.1.1
 configobj==5.0.6
 fysom==2.1.2
 Logbook==0.7.0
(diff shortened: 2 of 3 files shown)


boxcla:

shriganeshs:
Not sure why travis tests are failing with following error during `init-dev`. There are no changes related to requirements in this PR.
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/six-1.10.0.dist-info/METADATA'
```

app/util/conf/base_config_loader.py
```diff
@@ -135,6 +137,20 @@ def load_from_config_file(self, config, config_filename):
         if self.CONFIG_FILE_SECTION:
             self._load_section_from_config_file(config, config_filename, self.CONFIG_FILE_SECTION)
 
+    def _override_cacerts(self, root_directory):
```
josephharrington: Do you think it might be better to add this logic in main.py instead? It's not really related to config loading, so I'm not sure it belongs in BaseConfigLoader.

shriganeshs: How about putting it into `BaseConfigLoader.configure_postload` section?
This change is only required if we are using HTTPS protocol.

shriganeshs: Here 
```
        if Configuration['https_cert_file'] and Configuration['https_key_file']:
            conf.set('protocol_scheme', 'https')
            _override_cacerts(root_directory)
```

josephharrington: That sounds good.
\

requirements.txt
```diff
@@ -1,5 +1,6 @@
 # Install with "pip install -r requirements.txt"
 
+setuptools==33.1.1
```
josephharrington: Please keep this alphabetized. Also consider adding a comment on this setuptools line referencing the issue.

app/util/conf/base_config_loader.py
```diff
@@ -135,6 +137,21 @@ def load_from_config_file(self, config, config_filename):
...
+        file into frozen package and at startup update the environment to point to these cacerts.
+        """
+        if getattr(sys, 'frozen', False):
+            root_directory = dirname(sys.executable)
+            cacert_file = '{}/cacert.pem'.format(root_directory)
+            if isfile(cacert_file):
+                environ['REQUESTS_CA_BUNDLE'] = cacert_file
+            else:
+                print('Missing SSL cacerts file {} in the frozen ClusterRunner package.'.format(cacert_file))
+                sys.exit(1)
```
josephharrington: Seems like this should be an exception instead of printing and calling sys.exit. This class should only be responsible for parsing config. Any prints or calls to sys.exit should probably be in main.py -- what do you think?

requirements.txt
shriganeshs: I was under the impression that the packages in the list are installed in sequence and to avoid the issue (https://github.com/pypa/setuptools/issues/951) setuptools needs to be at the top. But if thats not the case I will alphabetized.
I will add the comment. According to the issue (FileNotFoundError) as described in https://github.com/pypa/setuptools/issues/951, I need to avoid the setuptools version 35.x.x and onwards. Downgrading of setuptools is a workaround for the issue.

shriganeshs: Alphabetization not possible here. I verified it.

josephharrington: Ah good find -- I didn't realize the order mattered.

app/util/conf/base_config_loader.py
```diff
@@ -135,6 +138,20 @@ def load_from_config_file(self, config, config_filename):
...
+        file into frozen package and at startup update the environment to point to these cacerts.
+        """
+        if getattr(sys, 'frozen', False):
+            root_directory = dirname(sys.executable)
+            cacert_file = '{}/cacert.pem'.format(root_directory)
+            if isfile(cacert_file):
+                environ['REQUESTS_CA_BUNDLE'] = cacert_file
+            else:
+                raise FileNotFoundError(errno.ENOENT, strerror(errno.ENOENT), cacert_file)
+
```
shriganeshs: This can be done with `os.stat(filename)` which will throw FileNotFound exception. But I think raising an exception manually is cleaner approach.

shriganeshs: What do you think about whether to catch this exception in main and print and exit(1) or let the stack trace get printed and continue?

So in main:_initialize_configuration
```
try:
        conf_loader.configure_postload(config)
except FileNotFoundError:
        print "Error etc."
        sys.exit(1)
```

josephharrington: Eh, I think it's fine to print the stack trace. It's not really a user error.

josephharrington:
Do you think this needs unit tests?

shriganeshs:
I do not see the need of unit tests, as this is a workaround for cx_freeze issue. Its not a part of ClusterRunner logic and also its possible that in future it may get fixed in newer versions of cx_Freeze. This is my opinion, otherwise unit tests for `_override_cacerts` are straight-forward.

josephharrington:
Sure, I can dig it.

PR #410, opened 2017-10-04
Set pipefail option in subprocesses

### LINKED ISSUE #321: Set pipefail option in shell environment
This is a suggestion from @pwsouth to make the results of shell executions less surprising.
https://sipb.mit.edu/doc/safe-shell/

I'm not sure what the equivalent is for windows.

ethomas2:
Fixes #321

### DIFF
--- a/app/util/process_utils.py
+++ b/app/util/process_utils.py
@@ -76,6 +76,21 @@ def Popen_with_delayed_expansion(cmd, *args, **kwargs):
         else:
             cmd_with_deplayed_expansion.extend(cmd)
         cmd = cmd_with_deplayed_expansion
+    else:
+        # Ordinarily, if you pipe commands in bash (eg "cmd1 | cmd2 | cmd3")
+        # and the rightmost command succeeds, the whole command will succeed
+        # even if one of the earlier commands fail. This is potentially
+        # problematic. See https://github.com/box/ClusterRunner/issues/321 for
+        # an example. To prevent this, set bash's pipefail option
+        should_set_pipefail = (
+            kwargs.get('shell') is True and
+            kwargs.get('executabe') in [None, '/bin/bash'] and
+            os.path.exists('/bin/bash') and
+            isinstance(cmd, str)
+        )
+        if should_set_pipefail:
+            kwargs['executable'] = '/bin/bash'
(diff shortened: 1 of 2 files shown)


boxcla:

app/util/process_utils.py
```diff
@@ -76,6 +76,9 @@ def Popen_with_delayed_expansion(cmd, *args, **kwargs):
         else:
             cmd_with_deplayed_expansion.extend(cmd)
         cmd = cmd_with_deplayed_expansion
+    else:
+        if kwargs.get('shell') == True:
+            cmd = 'set -o pipefail; ' + cmd
```
cmcginty: Can you add a short comment to explain why this is needed since it might not be that obvious.

test/unit/util/test_process_utils.py
```diff
@@ -40,7 +40,9 @@ def test_Popen_with_deplayed_expansion(self, input_cmd, os_name, expected_final_
         Popen_with_delayed_expansion(input_cmd)
 
         # Assert
-        mock_subprocess_popen.assert_called_once_with(expected_final_cmd)
+        self.assertEqual(len(mock_subprocess_popen.call_args_list), 1)
+        (cmd,), _ = mock_subprocess_popen.call_args
+        self.assertEqual(cmd, expected_final_cmd)
```
josephharrington: This doesn't test the code you added.

app/util/process_utils.py
```diff
@@ -76,6 +76,21 @@ def Popen_with_delayed_expansion(cmd, *args, **kwargs):
...
+        # an example. To prevent this, set bash's pipefail option
+        should_set_pipefail = (
+            kwargs.get('shell') is True and
+            kwargs.get('executabe') in [None, '/bin/bash'] and
+            os.path.exists('/bin/bash') and
+            isinstance(cmd, str)
+        )
+        if should_set_pipefail:
+            kwargs['executable'] = '/bin/bash'
+            cmd = 'set -o pipefail; ' + cmd
```
josephharrington: Either this logic fits better somewhere else or this method name/docstring should have probably been updated since you added code unrelated to delayed expansion.

PR #412, opened 2018-02-09
Handle build cancellation at slave level

shriganeshs:


boxcla:

josephharrington:
@shriganeshs Do you mind adding a commit message/PR description to this? It's good to record the historical context for changes.

app/master/build_scheduler_pool.py
```diff
@@ -20,12 +21,16 @@ def get(self, build):
         :type build: Build
         :rtype: BuildScheduler
         """
+        if isinstance(build, Build):
```
josephharrington: I think changes to this class might not be necessary if my suggestions in my previous comments are doable. If they are necessary though, my personal preference would be to add a `build_id` parameter to this method, then the validation logic here would check that exactly one of `build` or `build_id` is specified.

app/master/build.py
```diff
@@ -205,7 +206,10 @@ def _parse_payload_for_atom_exit_code(self, subjob_id):
             )
             atom_exit_code_file_sys_path = os.path.join(artifact_dir, BuildArtifact.EXIT_CODE_FILE)
             with open(atom_exit_code_file_sys_path, 'r') as atom_exit_code_file:
-                subjob.atoms[atom_id].exit_code = int(atom_exit_code_file.read())
+                exit_code = int(atom_exit_code_file.read())
+                subjob.atoms[atom_id].exit_code = exit_code
+                if subjob.is_canceled == False and (exit_code == -15 or exit_code == -9): # SIGTERM or SIGKILL
```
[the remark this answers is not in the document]
shriganeshs: It may happen that the subjob just complemented and before sending results it got canceled.

app/master/build_scheduler_pool.py
shriganeshs: Yes. I forgot to point out this. Its a hack which we used and we can remove some of the changes. I will update it and get back to you.
Also I tested the POST on "cancel" endpoint and it works.

shriganeshs:
I will address all the comments when I schedule this for sprint.

josephharrington:
Hey @shriganeshs, do you still plan to finish this or should we close it?

shriganeshs:
This PR is a hackathon project. So basically treat it as a PoC. Closing as we will not be merging this.

PR #414, opened 2018-03-10
Stop raising prometheus error metric on setup build failure

ethomas2:
We used to increment a prometheus error counter whenever build setup
failed. We did this because a failure in build setup used to kill the
slave process. We fixed that in #377. Incrementing the prometheus
counter is now unnecessary noise.


boxcla:

ethomas2:
Thanks @raythang for pointing this out

PR #420, opened 2018-03-30
Reconfigure HTTP request buckets

mayurva:
This change reconfigures the HTTP buckets such that we get a better insight into requests taking longer than 10 seconds


boxcla:

PR #421, opened 2018-03-30
Add rpm packaging using fpm

cmcginty:


boxcla:

conf/clusterrunner-master
```diff
@@ -0,0 +1,99 @@
+#!/bin/sh
```
cmcginty: This file is from a sysv template project on github. I only modified the defines at the top.

app/util/autoversioning.py
```diff
@@ -19,72 +14,18 @@ def get_version():
...
-
-def restore_original_package_version_file():
-    """
-    Restore the backed up version of package_version.py. This is just a convenience method to help
-    us remember not to commit changes to the package version file.
-    """
-    if not _try_rename(_VERSION_FILE_BACKUP_PATH, _VERSION_FILE_PATH):
-        # Ensure the temp file is removed if there was no backup copy.
-        os.remove(_VERSION_FILE_PATH)
+    return _calculate_source_version() or _get_frozen_package_version()
```
cmcginty: I feel like this is still overly complicated. The reason right now is so that the developer can see the CR version change in their local repo and have a "-mod" value append. Maybe this is not necessary. Then all you need do is:
1) call _calculate_source_verison() in setup.py when you are building the pex file
2) all other code calls _get_frozen_packge_version() which reads the valid version from pex
3) if #2 fails because you're in a local repo, then the version string could fallback to default value or a timestamp.

The current version generation is in the reverse order from above, so that the developer will always see the latest calculated version in their repo.

app/util/autoversioning.py
```diff
@@ -1,90 +1,32 @@
...
-
-def restore_original_package_version_file():
-    """
-    Restore the backed up version of package_version.py. This is just a convenience method to help
-    us remember not to commit changes to the package version file.
-    """
-    if not _try_rename(_VERSION_FILE_BACKUP_PATH, _VERSION_FILE_PATH):
-        # Ensure the temp file is removed if there was no backup copy.
-        os.remove(_VERSION_FILE_PATH)
+    return _calculate_source_version() or _get_frozen_package_version() or 'UNK_VERSION'
```
cmcginty: I feel like this is still overly complicated. The reason right now is so that the developer can see the CR version change in their local repo and have a "-mod" value append. Maybe this is not necessary. Then all you need do is:

call _calculate_source_verison() in setup.py when you are building the pex file
all other code calls _get_frozen_packge_version() which reads the valid version from pex
if #2 fails because you're in a local repo, then the version string could fallback to default value or a timestamp.
The current version generation is in the reverse order from above, so that the developer will always see the latest calculated version in their repo.

app/util/autoversioning.py
```diff
@@ -1,90 +1,32 @@
...
-
-def restore_original_package_version_file():
-    """
-    Restore the backed up version of package_version.py. This is just a convenience method to help
-    us remember not to commit changes to the package version file.
-    """
-    if not _try_rename(_VERSION_FILE_BACKUP_PATH, _VERSION_FILE_PATH):
-        # Ensure the temp file is removed if there was no backup copy.
-        os.remove(_VERSION_FILE_PATH)
+    return _calculate_source_version() or _get_frozen_package_version() or '0.0'
```
cmcginty: I feel like this is still overly complicated. The reason right now is so that the developer can see the CR version change in their local repo and have a "-mod" value append. Maybe this is not necessary. Then all you need do is:

call _calculate_source_verison() in setup.py when you are building the pex file
all other code calls _get_frozen_packge_version() which reads the valid version from pex
if #2 fails because you're in a local repo, then the version string could fallback to default value or a timestamp.
The current version generation is in the reverse order from above, so that the developer will always see the latest calculated version in their repo.

conf/clusterrunner-master
mayurva: this is cool. Can we have a similar init.d script for slaves too?

cmcginty: Right. I was planning to add in a separate PR later. It's basically a copy of the current script so wasn't going to add much value to this one. I'll note it in the jira issue I have for deploying the slaves.

app/util/autoversioning.py
```diff
@@ -144,11 +82,13 @@ def _repo_has_uncommited_changes():
...
     """
-    Check if the current HEAD is in the first-parent chain of origin/master. The first-parent chain of origin/master
-    consists of all the "trunk" commits. All other commits are either on merged branches or haven't been merged at all.
+    Check if the current HEAD is in the first-parent chain of origin/master. The first-parent chain
+    of origin/master consists of all the "trunk" commits. All other commits are either on merged
+    branches or haven't been merged at all.
 
     :type commit_hash: str
     :rtype: bool
+    :raises CalledProcessError: if git not a git repo or is a shallow clone
```
cmcginty: Should say `if app is not in a git repo`

conf/clusterrunner-master
```diff
@@ -0,0 +1,99 @@
...
+### BEGIN INIT INFO
```
ethomas2: I usually add `set -euxo pipefail` at the top of any non-trivial bash script to make bash behave more sanely
https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

conf/clusterrunner-master
```diff
@@ -0,0 +1,99 @@
...
+    [ -f "$pid_file" ] && ps -p `get_pid` > /dev/null 2>&1
+}
+
+case "$1" in
+    start)
+    if is_running; then
+        echo "Already started"
+    else
+        echo "Starting $name"
+        cd "$dir"
```
ethomas2: Generally not a fan of a process changing the working directory it started off with. How about executing all of this code in a subshell?
```
echo "Starting $name"
(
  cd "$dir"
  if [-z "$user" ]; then
  ...
)
```

ethomas2: hmmm, then again, maybe this is a terrible idea. If the subshell exits 1 you'd probably want the whole script to exit 1. There might be some other caveats I haven't thought of. Lets discuss this when you get back from PTO

conf/clusterrunner-master
```diff
@@ -0,0 +1,99 @@
...
+# Description:       Enable clusterrunner service provided by daemon.
+### END INIT INFO
+
+dir="/var/lib/clusterrunner"
+cmd="nohup clusterrunner master"
+user="jenkins"
+
+name=`basename $0`
+pid_file="/var/run/$name.pid"
+stdout_log="/var/log/$name.log"
```
ethomas2: Why are we separating stdout_log and stderr_log?

conf/clusterrunner-master
```diff
@@ -0,0 +1,99 @@
...
+                break
+            fi
+
+            echo -n "."
+            sleep 1
+        done
+        echo
+
+        if is_running; then
+            echo "Not stopped; may still be shutting down or shutdown may have failed"
```
ethomas2: How do we feel about making this red with ansi color codes? Superfluous and unnecessary? Or useful and cool?

app/util/autoversioning.py
ethomas2: ayyyyy. Hi @nickzuber !

app/util/autoversioning.py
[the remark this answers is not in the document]
cmcginty: No reason. I kind of when in a circle to try and make this be more obviously a bad version number, but it breaks setup.py inside of travis. So I changed back to a version string. I'll update this back to `0.0.0`

conf/clusterrunner-master
cmcginty: Did you see my original comment. I didn't write this file. It's from https://github.com/fhd/init-script-template.

If we want to make change we probably should try to do them upstream.

ethomas2: Ah, didn't realize that this file was copy pasted. What do you mean upstream? The only thing upstream from this file is systemd right? It's not like we can change that. And even if we could I don't think any bug that would be avoided by `set -euxo pipefail` could be properly handled by anything other than the bash script itself (downstream or upstream).

I _can_ see the aversion to modifying a copy-pasted file. Afaikt this file looks bug free right now and `set -euxo pipefail` won't really do anything. My main worry is that we'll make modifications to this file, introduce some bug and that bug will be exacerbated by bash's shitty default behavior. My secondary worry is that there's already a bug in this file that I haven't spotted that will be exacerbated by bash's shitty default behavior.

PR #422, opened 2018-04-05
Add heartbeat functionality between master and slaves

mayurva:


boxcla:

app/web_framework/cluster_master_application.py
```diff
@@ -418,3 +422,16 @@ def post(self):
...
+    def get(self):
+        response = {
+            'slaves': [slave.api_representation() for slave in self._cluster_master.all_slaves_by_id().values()]
+        }
+        self.write(response)
+
+    @authenticated
+    def post(self, slave_id):
+        self._cluster_master.receive_heartbeat_from_slave(slave_id)
+        self.write({'message': 'heartbeat API POST'})
```
josephharrington: I think empty response is fine here. Did you have an intent for the message?

app/slave/cluster_slave.py
```diff
@@ -56,6 +58,29 @@ def __init__(self, port, host, num_executors=10):
...
+        state_url = self._master_api.url('slave', self._slave_id, 'heartbeat')
+        try:
+            self._network.post_with_digest(state_url, request_params={'slave': {'heartbeat': True}},
+                                           secret=Secret.get())
+            self._heartbeat_count = 0
+        except (requests.ConnectionError, requests.Timeout):
+            self._heartbeat_count += 1
+            if self._heartbeat_count >= self._heartbeat_count_threshold:
+                self._logger.warning('Master is not responding to heartbeats')
+                self._heartbeat_job.remove()
```
josephharrington: Do you plan to do more here? It seems like this is a good point to shut down the slave service. This is related to my other comment about running in a SafeThread.

app/master/cluster_master.py
```diff
@@ -55,6 +62,22 @@ def __init__(self):
 
         SlavesCollector.register_slaves_metrics_collector(lambda: self.all_slaves_by_id().values())
 
+    def configure_heartbeat(self):
+        self._heartbeat_job = self._scheduler.add_job(self.heartbeat, 'interval', seconds=self._heartbeat_frequency)
+        return self._scheduler
+
+    def heartbeat(self):
```
josephharrington: It seems like the word "heartbeat" refers to two different things. On the slave it's an API call to the master. On the master it's a periodic cleanup of the slave list. Should these have different names?

app/master/cluster_master.py
```diff
@@ -334,3 +357,9 @@ def get_path_for_build_results_archive(self, build_id: int, is_tar_request: bool
             raise ItemNotReadyError('Build artifact file is not yet ready. Try again later.')
 
         return archive_file
+
+    def receive_heartbeat_from_slave(self, slave_id):
+        self._thread_pool_executor.submit(self._async_receive_heartbeat_from_slave(int(slave_id)))
```
josephharrington: I don't think this needs to be async, does it? It's only setting a single value. It should only be necessary to spawn a thread if the work triggered by an API call is going to take a significant amount of time.

app/master/cluster_master.py
```diff
@@ -55,6 +62,22 @@ def __init__(self):
...
 
+    def configure_heartbeat(self):
+        self._heartbeat_job = self._scheduler.add_job(self.heartbeat, 'interval', seconds=self._heartbeat_frequency)
+        return self._scheduler
+
+    def heartbeat(self):
+        t = datetime.datetime.now()
+        slaves_to_disconnect = []
+        for slave in self._all_slaves_by_url.values():
+            if slave.is_alive() and not slave.is_responsive(t, self._heartbeat_frequency):
```
josephharrington: Maybe it makes more sense to expose `slave.last_heartbeat_time()` and do the comparison here. It feels a bit strange to ask a `Slave` instance if it's responsive while also telling it the entire context for which it will be considered responsive.

app/master/cluster_master.py
```diff
@@ -55,6 +62,22 @@ def __init__(self):
...
+                slaves_to_disconnect.append(slave)
+
+        for slave in slaves_to_disconnect:
+            self._disconnect_slave(slave)
+            self._logger.warning('Slave {} marked offline as it is not sending heartbeats.'.format(
```
josephharrington: This should probably be more than a warning. I think this should be logged as an error and/or increase a Prometheus error metric.

requirements.txt
```diff
@@ -3,6 +3,7 @@
 # Downgrade the setuptools version to workaround the issue replacing the setuptools dependency
 # IOError METADATA no such file. Check https://github.com/pypa/setuptools/issues/951 for more details.
 setuptools==33.1.1
+apscheduler==3.5.1
```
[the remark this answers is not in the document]
mayurva: yeah. Me and @cmcginty briefly discussed about `apscheduler` being a overkill, but it seemed pretty straight forward to set up and use. I was hoping to use their native `TornadoScheduler`, but the current tornado version we use doesn't support it. I'll check out the `sched` module. Also good point on using safe thread. I'll verify the behavior on uncaught exceptions.

app/master/cluster_master.py
mayurva: ah. Interesting point. Thanks!

requirements.txt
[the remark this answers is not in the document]
josephharrington: As for apscheduler being easy to set up and use -- I totally agree. The code itself that you had around it is pretty clean. I'm less concerned about that than just the size and functionality of the library that we won't (and might never) leverage.

I don't think using `sched` would be much more complex. Something like:
```python
def __init__(self):
    self._hb_scheduler = sched.scheduler()
    SafeThread(target=self._start_beatin, name='HeartbeatLoop', daemon=True).start()  # this could also go in a public start() method similar to BuildRequestHandler

def _start_beatin(self):
    self._hb_scheduler.enter(0, 0, self._do_heartbeat)
    self._hb_scheduler.run()

def heartbeat_loop():
    # network request calls, error handling, metrics recording here
    self._hb_scheduler.enter(self._hb_freq, 0, self._do_heartbeat)  # schedule next beat. (if you wanted to be super precise you could subtract from the frequency the amount of time this method took to execute, but that shouldn't matter much for the intervals we're talking about.)
```

app/web_framework/cluster_master_application.py
```diff
@@ -54,9 +54,11 @@ def __init__(self, cluster_master):
                 RouteNode(r'queue', _QueueHandler),
                 RouteNode(r'slave', _SlavesHandler, 'slaves').add_children([
                     RouteNode(r'(\d+)', _SlaveHandler, 'slave').add_children([
-                        RouteNode(r'shutdown', _SlaveShutdownHandler, 'shutdown')
+                        RouteNode(r'shutdown', _SlaveShutdownHandler, 'shutdown'),
+                        RouteNode(r'heartbeat', _SlavesHeartbeatHandler)
```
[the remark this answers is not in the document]
mayurva: So.. I actually wanted to add it to v2 only, but seems like slave only uses v1 APIs to talk to master. I was unsure if there was a specific reason for that. I just decided to be consistent with the existing code.

I think we could use the `/slave/<id>` endpoint, especially since we are updating the same data structure. My initial thoughts were to maintain separate data structure to track heartbeat, but that seemed overly complicated when I got to the implementation.

app/web_framework/cluster_master_application.py
```diff
@@ -418,3 +422,16 @@ def post(self):
             [int(slave_id) for slave_id in self.decoded_body.get('slaves')]
 
         self._cluster_master.set_shutdown_mode_on_slaves(slaves_to_shutdown)
+
+
+class _SlavesHeartbeatHandler(_ClusterMasterBaseAPIHandler):
+    def get(self):
```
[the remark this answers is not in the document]
mayurva: The idea was to add endpoint for heartbeat info for all the slaves, but I think it can be cleaned up since it is now exactly same as GET on `/slave` endpoint

app/slave/cluster_slave.py
mayurva: On second thoughts, since slave is just going to sit idle after this point, it probably makes sense to just kill it. We can change the behavior in future.

app/master/cluster_master.py
```diff
@@ -35,6 +38,10 @@ def __init__(self):
         self._build_request_handler.start()
         self._slave_allocator = SlaveAllocator(self._scheduler_pool)
         self._slave_allocator.start()
+
+        self._heartbeat_frequency = Configuration['heartbeat_frequency'] * Configuration['heartbeat_count_threshold']
```
ethomas2: Nit: `heartbeat_frequency` should be named `heartbeat_period` or `heartbeat_interval`. A frequency has units 'something per second', a period has units of seconds. I expect heartbeat_frequency to have units 'beats per second'. When I multiplied the frequency by a time interval, I would expect to get the number of beats that should occur in that time interval. That's not what this variable represents.

app/master/slave.py
```diff
@@ -236,6 +238,14 @@ def _expected_session_header(self):
 
         return headers
 
+    def set_heartbeat(self, time):
+        self._heartbeat = time
+
+    def is_responsive(self, time, heartbeat_frequency):
```
ethomas2: 1. Can we rename `time` to `current_time` or `now` (that's both clearer, and it avoids shadowing the time module, which is a common python module)
2. Rename self._heartbeat to self._last_checkin
3. rename heartbeat_frequency to heartbeat_period
4. This method can be rewritten as
```python
return (current_time - self._last_checkin) < heartbeat_period
```
5. Given that this method is effectively a one-liner, it might be cleaner to just expose self._last_checkin to the caller, and let the caller execute this one line itself
(the argument against this point is that the name is_responsive describes what this one line does better than the inlined code does. I'm not sure if I agree with that or not. If you want to make that argument i'll give it some thought and see if I agree)

app/slave/cluster_slave.py
```diff
@@ -56,6 +58,29 @@ def __init__(self, port, host, num_executors=10):
...
+        self._heartbeat_frequency = Configuration['heartbeat_frequency']
+        self.scheduler = BackgroundScheduler()
+        self._heartbeat_job = None
+        self._logger.info('Heartbeat will run every {} seconds'.format(self._heartbeat_frequency))
+
+    def configure_heartbeat(self):
+        self._heartbeat_job = self.scheduler.add_job(self.heartbeat, 'interval', seconds=self._heartbeat_frequency)
+        return self.scheduler
+
+    def heartbeat(self):
```
ethomas2: Rename to send_heartbeat. Same comment about function names being verb phrases

app/slave/cluster_slave.py
```diff
@@ -56,6 +58,29 @@ def __init__(self, port, host, num_executors=10):
...
+        self._heartbeat_count = 0
+        self._heartbeat_count_threshold = Configuration['heartbeat_count_threshold']
+        self._heartbeat_frequency = Configuration['heartbeat_frequency']
+        self.scheduler = BackgroundScheduler()
+        self._heartbeat_job = None
+        self._logger.info('Heartbeat will run every {} seconds'.format(self._heartbeat_frequency))
+
+    def configure_heartbeat(self):
+        self._heartbeat_job = self.scheduler.add_job(self.heartbeat, 'interval', seconds=self._heartbeat_frequency)
+        return self.scheduler
```
ethomas2: Same comment about not needing to return the scheduler, and about inlining this into the constructor

app/master/cluster_master.py
```diff
@@ -35,6 +38,10 @@ def __init__(self):
...
+        self._scheduler = BackgroundScheduler()
+        self._heartbeat_job = None
```
ethomas2: In general, I don't like explicitly assigning None to variables. Sometimes that can't be avoided, but it's rare.

In this example, inlining configure_heartbeat would do the trick

app/slave/cluster_slave.py
```diff
@@ -56,6 +58,29 @@ def __init__(self, port, host, num_executors=10):
...
+
+    def heartbeat(self):
+        state_url = self._master_api.url('slave', self._slave_id, 'heartbeat')
+        try:
+            self._network.post_with_digest(state_url, request_params={'slave': {'heartbeat': True}},
```
ethomas2: This is unfortunate. I'd really like to see this be
`yield self._network.post_with_digest(...)`

Unfortunately post_with_digest doesn't return a future. And rewriting it to return a future would be pretty involved. Definitely doesn't belong in this pr. Can you add a TODO to

1. Rewrite network.post_with_digest to return a future/be a coroutine
2. Run this code on tornado's event loop instead of on it's own thread

app/slave/cluster_slave.py
ethomas2: If that's the case then you should add a TODO here explaining the intention. Right now it looks incomplete (because it is)

ethomas2: Also, how/why is the request to reconnect to the master different than the heartbeat request? Why can't you just keep sending requests to the master forever, regardless of if the master is responding (i.e. never call self._heartbeat_job.remove()).

When the master comes back online and the first successful heartbeat is received the master connects that slave.

You may have already explained this in your design doc. If you did, sorry for not noticing this until now.

ethomas2:
Also, put type hints on all your functions pls. Most functions in CR don't have type hints, because we just started adding them recently (probably why you didn't realize you should be doing it), but every new function in CR should have type hints

Examples
```python
mbp-IT002597:ClusterRunner ethomas$ ack "def.*\(.*:.*\)"
app/client/cluster_api_client.py
210:    def connect_slave(self, slave_url: str, num_executors: int=10) -> int:
227:    def get_slave_status(self, slave_id: int) -> dict:
237:    def block_until_slave_offline(self, slave_id: int, timeout: int=None) -> bool:

app/common/console_output.py
14:    def from_plaintext(cls, path: str) -> 'ConsoleOutput':
22:    def from_zipfile(cls, zip_path: str, path_in_archive: str) -> 'ConsoleOutput':
34:    def __init__(self, file: BinaryIO):
41:    def segment(self, max_lines: int=50, offset_line: Optional[int]=None) -> ConsoleOut
```

app/slave/cluster_slave.py
mayurva: Interesting point. The TODO is to call connect_to_master for reconnection instead of just simple post to heartbeat when master is dead. The idea is that master's identity might have changed as it probably restarted or something. I guess in future heartbeat/connect_to_master can be combined and behave differently based on current state of connection, but we are probably not there yet. In any case, I'll clarify this in a TODO comment.

test/unit/master/test_cluster_master.py
```diff
@@ -213,6 +213,38 @@ def test_updating_slave_to_nonexistent_state_should_raise_bad_request_error(self
         with self.assertRaises(BadRequestError):
             master.handle_slave_state_update(slave, 'NONEXISTENT_STATE')
 
+    def test_async_receive_heartbeat_from_slave_calls_set_heartbeat_for_the_slave(self):
+        master = ClusterMaster()
+        slave_url = "url"
+        slave_id = master.connect_slave(slave_url, 1)
+        slave = Mock()
+        master.get_slave = MagicMock(return_value=slave)
+        master._async_receive_heartbeat_from_slave(slave_id['slave_id'])
```
ethomas2: Not a fan of calling a private method in a test. It's an indication that you're testing an implementation detail, which in turn is an indication that this test might be brittle. Brittle tests are a huge pet peeve of mine, and something that this team has suffered heavily from in the past.

The easiest way to fix this test is to call `master.receive_heartbeat_from_slave` directly, patch the slave constructor, and assert that when you call `master.receive_heartbeat_from_slave` the return value of the slave constructor (i.e. the slave) gets `set_heartbeat` called. 

An improvement upon that test is instead of calling `master.receive_heartbeat_from_slave`, send a POST to the heartbeat endpoint and assert that slave.set_heartbeat is called tornado's [AsyncHTTPTestCase](http://www.tornadoweb.org/en/stable/testing.html#tornado.testing.AsyncHTTPTestCase) will be helpful here. Look at `get_app()` and `fetch`

See the internal box code that I slack'd to you

test/unit/master/test_cluster_master.py
```diff
@@ -213,6 +213,38 @@ def test_updating_slave_to_nonexistent_state_should_raise_bad_request_error(self
...
+        slave_dead=(False,True,),
+        slave_responsive=(True,True,),
+    )
+    def test_heartbeat_disconnects_unresponsive_slave(self, slave_alive, slave_responsive):
+        master = ClusterMaster()
+        slave_url = "url"
+        slave = master.connect_slave(slave_url, 1)
+        slave = master.get_slave(int(slave['slave_id']))
+
+        slave.is_alive = MagicMock(return_value=slave_alive)
```
ethomas2: Same comment about clobbering a method from a real object.

Instead of clobbering `slave.is_alive`, use patch to patch out the `Slave()` constructor, that way you have a reference to a mock slave object, and set the attributes on that. 

Instead of clobbering master._disconnect_slave and asserting that that's called, patch the Slave constructor and, assert `slave.mark_dead()` is called

app/web_framework/cluster_master_application.py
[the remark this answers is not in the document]
mayurva: It seems like the `/slave/<id>` endpoint currently only handles slave state change functionality. Heartbeat is not really a state change. So, I think we either keep a separate `heartbeat` end point as it is right now, or send heartbeat to `/slave/<id>` with current state of the slave, and add logic in master to do nothing if there is no state change. What do you guys think?

Regardless, I think `_last_heartbeat_time` should be updated for both of those requests.

requirements.txt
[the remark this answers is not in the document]
mayurva: uhh.. I hate how GitHub handles comments, but I reimplemented the heartbeat using sched. I still not fully satisfied, but I think it's better than before

app/master/slave.py
```diff
@@ -236,6 +238,12 @@ def _expected_session_header(self):
 
         return headers
 
+    def set_last_heartbeat_time(self):
```
shriganeshs: Minor (can ignore): Say `update` instead of `set`. IMO `set` usually sets the value to whatever is passed in as an argument.

app/master/cluster_master.py
```diff
@@ -53,8 +58,36 @@ def __init__(self):
...
+        # Configure heartbeat tracking
+        self._unresponsive_slaves_cleanup_interval = Configuration['unresponsive_slaves_cleanup_interval']
+        self._hb_scheduler = sched.scheduler()
+
         SlavesCollector.register_slaves_metrics_collector(lambda: self.all_slaves_by_id().values())
 
+    def start_heartbeat_tracker_thread(self):
+        self._logger.info('Heartbeat tracker will run every {} seconds'.format(
+            self._unresponsive_slaves_cleanup_interval))
+        SafeThread(target=self._start_heartbeat_tracker, name='HeartbeatTrackerThread', daemon=True).start()
```
ethomas2: I'm not sure I agree with this use of `SafeThread` (or with most uses of `SafeThread`). I know @josephharrington explicitly said to use it here, but I disagree. `SafeThread` will kill the whole process if an error bubbles to the top of it. My understanding of the history of SafeThread is that it was introduced in a time before we had alerts in cluster runner. The only way of knowing if an error happened was to kill the whole process. 

But now we have alerts (at least on the master). If an error bubbles up to here, I wouldn't want to kill the whole process, i'd just want to log an error, trigger an alert and keep chugging along. Especially considering that this thread isn't handling anything super duper business critical. It's not as if failures in this thread will make CR unable to run builds or anything like that. Crashing the master seems like an over reaction.

app/slave/cluster_slave.py
```diff
@@ -56,6 +58,41 @@ def __init__(self, port, host, num_executors=10):
...
 
+        # Configure heartbeat
+        self._heartbeat_failure_count = 0
+        self._heartbeat_failure_threshold = Configuration['heartbeat_failure_threshold']
+        self._heartbeat_interval = Configuration['heartbeat_interval']
+        self._hb_scheduler = sched.scheduler()
+
+    def start_heartbeat_thread(self):
+        self._logger.info('Heartbeat will run every {} seconds'.format(self._heartbeat_interval))
+        SafeThread(target=self._start_heartbeat, name='HeartbeatThread', daemon=True).start()
```
ethomas2: Sort've the same comment about SafeThread here, but i'm less confident that this is bad than the use of SafeThread in the master. We should probably talk about this.

1. Crashing a slave is much less devastating than crashing the master
2. We don't have alerts on the slave ... so maybe crashing the slave is the best we can do? Maybe we _shouldn't_ even make alerts on the slave. Maybe the proper response for a slave erring is just crashing the slave and relying on the master/kubernetes to spin up a new one. (hehe, kubernetes. I can dream)

app/slave/cluster_slave.py
```diff
@@ -56,6 +58,41 @@ def __init__(self, port, host, num_executors=10):
...
+            self._heartbeat_failure_count = 0
+        except (requests.ConnectionError, requests.Timeout):
+            self._heartbeat_failure_count += 1
+            if self._heartbeat_failure_count >= self._heartbeat_failure_threshold:
+                self._logger.error('Master is not responding to heartbeats')
+
+                # TODO: Right now the slave simply dies when it does not hear back from master. The next step would
+                # be to try to reconnect to master at this point. In future the heartbeat and connect_to_master
+                # methods can combined into one. This combined method will behave differently based on current state.
+                self.kill()
```
ethomas2: Why even kill it? Why not just let it stay up forever? Indefinitely spamming the master with requests.

I suppose this is gonna change really soon anyway, so nbd.

app/master/cluster_master.py
josephharrington: To clarify, I agree with Evan that this should definitely not be a SafeThread. The main exception to that that I can think of is if we ever extend this to be a generic periodic task thread using an event loop (for example to support generic periodic tasks like doing git gc on repos, purging old build data, etc.) Even if we do make it a SafeThread in that case, we'd still probably want a try/except around each individual task execution. (Individual tasks should be allowed to fail without killing the entire periodic execution mechanism.)

test/unit/master/test_cluster_master.py
```diff
@@ -213,6 +215,29 @@ def test_updating_slave_to_nonexistent_state_should_raise_bad_request_error(self
...
+    def test_heartbeat_disconnects_unresponsive_slave(self, slave_alive, seconds_since_last_heartbeat):
+        master = ClusterMaster()
+        slave_url = "url"
+        slave = master.connect_slave(slave_url, 1)
+        slave = master.get_slave(int(slave['slave_id']))
+        last_heartbeat_time = datetime.now() - timedelta(seconds=seconds_since_last_heartbeat)
+
+        slave.is_alive = MagicMock(return_value=slave_alive)
+        slave.get_last_heartbeat_time = MagicMock(return_value=last_heartbeat_time)
+        slave.mark_dead = Mock()
```
josephharrington: Do you need a real slave instance here? It's a bit surprising to see directly replacing methods on real objest with mocks via assignment and can lead to surprising results and stale tests (for example if someone renamed the mark_dead method but did not rename it here, your assertion about it never being called would pass trivially.) I believe the `patch` functionality does checks to guard against this.

More specifically, you should be able to patch out the `Slave` constructor so that you can inject a full mock Slave object that also asserts method calls match the real object api.

test/unit/master/test_cluster_master.py
```diff
@@ -213,6 +215,29 @@ def test_updating_slave_to_nonexistent_state_should_raise_bad_request_error(self
...
+
+        master._disconnect_non_heartbeating_slaves()
```
josephharrington: I think there was a comment about this previously, but ideally we don't test via private methods. There are exceptions, but most of the time it's a code smell.

test/unit/master/test_cluster_master.py
```diff
@@ -213,6 +215,29 @@ def test_updating_slave_to_nonexistent_state_should_raise_bad_request_error(self
...
+        slave_unresponsive=(True,1000,),
+        slave_dead=(False,60,),
+        slave_responsive=(True,60,),
+    )
+    def test_heartbeat_disconnects_unresponsive_slave(self, slave_alive, seconds_since_last_heartbeat):
+        master = ClusterMaster()
+        slave_url = "url"
+        slave = master.connect_slave(slave_url, 1)
+        slave = master.get_slave(int(slave['slave_id']))
+        last_heartbeat_time = datetime.now() - timedelta(seconds=seconds_since_last_heartbeat)
```
josephharrington: Don't use real datetime.now in tests. Mock it out so that we have deterministic test inputs.

test/unit/slave/test_cluster_slave.py
```diff
@@ -237,6 +244,34 @@ def test_execute_subjob_passes_base_executor_index_to_executor(self):
...
+        unresponsive_master=(False,1,),
+        unresponsive_master_retry=(False,2,),
+    )
+    def test_heartbeat_sends_post_to_master(self, is_master_responsive, heartbeat_failure_threshold):
+        expected_slave_heartbeat_url = 'http://{}/v1/slave/1/heartbeat'.format(self._FAKE_MASTER_URL)
+        Configuration['heartbeat_failure_threshold'] = heartbeat_failure_threshold
+
+        slave = self._create_cluster_slave()
+        slave.connect_to_master(self._FAKE_MASTER_URL)
+        slave.kill = Mock()
```
josephharrington: Same comment as above. Also, I'd rather us not mock out methods on the class under test. Assert on the behavior of the method, not the method itself being called.

test/unit/slave/test_cluster_slave.py
```diff
@@ -237,6 +244,34 @@ def test_execute_subjob_passes_base_executor_index_to_executor(self):
...
+        if not is_master_responsive:
+            self.mock_network.post_with_digest.side_effect = requests.ConnectionError
+
+        slave._run_heartbeat()
```
josephharrington: Same comment as above about testing private methods. If you can't find a way around that, maybe at least have a supporting test that verifies `start_heartbeat_thread` sets up `slave._run_heartbeat` to run periodically. (Currently it looks like you have no tests that hit the public methods you added to the service classes.)

I didn't try writing this so it may be harder than I think, but my first instinct would be to 
- patch out SafeThread so that it executes its target synchronously
- set configuration so that the Nth post call will fail
- set configuration (or stub calls to datetime.now) so that there is no delay between scheduler calls
- assert that calling `start_heartbeat_thread` ends up sending N-1 post calls to the master

Anyway, just a suggestion. This is more of a test robustness issue than an actual issue with the feature code.

app/web_framework/cluster_master_application.py
```diff
@@ -418,3 +421,10 @@ def post(self):
             [int(slave_id) for slave_id in self.decoded_body.get('slaves')]
 
         self._cluster_master.set_shutdown_mode_on_slaves(slaves_to_shutdown)
+
+
+class _SlavesHeartbeatHandler(_ClusterMasterBaseAPIHandler):
+    @authenticated
+    def post(self, slave_id):
+        slave = self._cluster_master.get_slave(int(slave_id))
+        slave.set_last_heartbeat_time()
```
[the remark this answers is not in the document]
josephharrington: Also, you may want to add some unit tests for the Slave class. I think I saw you mocked out everything on the Slave instance so that logic is not tested via your current tests?

app/web_framework/cluster_master_application.py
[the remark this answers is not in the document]
mayurva: yeah.. I thought about that. I'll not make any changes now, but definitely keep that in mind if we decide to revisit this.

app/web_framework/cluster_master_application.py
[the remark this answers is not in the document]
mayurva: um.. yeah.. the reason I removed the tests from the Slave class was because I moved all the logic out of there. The only methods left in there are a getter and a setter. Not sure if they really need unit tests.

app/slave/cluster_slave.py
ethomas2: I would say it's _acceptable_ for this to be a SafeThread, but I wouldn't go so far as to say it _should_ be a SafeThread. Given that we're not assuming ephemeral slaves, the ideal for "maximum robustness" is that the master and slave should be able to tolerate arbitrarily long periods of time without being able to communicate with one another. When communication becomes possible again, they should seamlessly reconnect. What happens if there's a network partition between the master and slaves that lasts an hour? What happens if the master's vm dies and it takes us 30 minutes to bring it back up? I think the standard for perfection in those cases is that the slave seamlessly reconnects itself back to the master once the issue is resolved, not that the slave kills itself.

I say it's acceptable for the slave to kill itself is because in practice if the slave can't communicate with the master for more than ~5 minutes then something has probably gone seriously wrong and we probably will need an engineer to intervene anyway. But I do think intentionally killing the slave is intentionally giving up on perfection.

app/slave/cluster_slave.py
ethomas2: I was just talking about the comment @mayurva left

```TODO: Right now the slave simply dies when it does not hear back from master.
The next step would be to try to reconnect to master at this point. In future
the heartbeat and connect_to_master methods can combined into one. This
combined method will behave differently based on current state.```

I think the plan is for this to happen soon

File: app/web_framework/cluster_master_application.py (hunk 27 above)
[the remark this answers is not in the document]
Comment (josephharrington) [reply app/web_framework/cluster_master_application.py]: @mayurva Sure, at some point it starts to get a little into testing philosophy. I usually find myself trying to ignore the complexity of the actual implementation and write tests for class behavior. I can see an argument for not wanting to test single line methods (on the other hand those are some of the easiest tests to write!)

My personal view is that even single lines can have bugs in them or someone can refactor them incorrectly. If someone comes along to one of my single-line methods and adds an if statement, all of a sudden they now have two tests to write -- one for their new functionality and one for the original functionality. Basically my view is that if the application depends on the code then they should be covered by *some* test -- whether it's a unit/integration/functional test.

Anyway, just my two cents. Don't consider my comments blocking if you feel strongly a different way. Happy to chat more in person if you'd rather.

File: app/slave/cluster_slave.py (hunk 20 above)
Comment (josephharrington) [reply app/slave/cluster_slave.py]: > the master and slave should be able to tolerate arbitrarily long periods of time without being able to communicate with one another
> ...
> I think the standard for perfection in those cases is that the slave seamlessly reconnects itself back to the master once the issue is resolved

Let's discuss this offline. It sounds like we have very different thoughts on what the behavior should be in the case of network failure.

File: test/unit/slave/test_cluster_slave.py [hunk 28]
```diff
@@ -237,6 +245,33 @@ def test_execute_subjob_passes_base_executor_index_to_executor(self):
...
+
+        if is_master_responsive:
+            self.mock_network.post_with_digest.assert_called_once_with(
+                expected_slave_heartbeat_url,request_params={'slave': {'heartbeat': True}}, secret=ANY)
+        else:
+            if heartbeat_failure_threshold == 1:
+                self.assertEqual(self._mock_sys.exit.call_count, 1,
+                                                 'slave dies when it decides that master is dead')
```
cmcginty: can you fix the indent here.

test/unit/slave/test_cluster_slave.py
```diff
@@ -237,6 +245,33 @@ def test_execute_subjob_passes_base_executor_index_to_executor(self):
 
         executor.execute_subjob.assert_called_with(1, 2, [], 12)
 
+    @genty_dataset(
+        responsive_master=(True,1,),
```
cmcginty: You only need the trailing comma if the tuple only has one value. For example:

```
responsive_master=(True,)    # required trailing comma
responsive_master=(True,1)   # two args, only 1 comma needed
```

app/master/slave.py
```diff
@@ -26,6 +27,7 @@ def __init__(self, slave_url, num_executors, slave_session_id=None):
         self._num_executors_in_use = Counter()
         self._network = Network(min_connection_poolsize=num_executors)
         self.current_build_id = None
+        self._last_heartbeat_time = datetime.now()
```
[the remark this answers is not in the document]
mayurva: Umm yeah. I could do that. I think the way it is currently written makes it clear that the variable is being initialized. Calling the function might obscure it a bit

test/unit/master/test_cluster_master.py
```diff
@@ -213,6 +219,40 @@ def test_updating_slave_to_nonexistent_state_should_raise_bad_request_error(self
...
+    @genty_dataset (
+        slave_unresponsive=(True,1000,),
+        slave_dead=(False,60,),
+        slave_responsive=(True,60,),
+    )
+    def test_heartbeat_disconnects_unresponsive_slave(self, slave_alive, seconds_since_last_heartbeat):
+        last_heartbeat_time = self._mock_current_datetime - timedelta(seconds=seconds_since_last_heartbeat)
+        master = ClusterMaster()
+
+        mock_slave = Mock()
```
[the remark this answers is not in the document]
mayurva: Me and Evan ran into a lot of issues in mocking this particular class. If I do specify `spec` I get the following error. It is because of the way we use public member id. I proper fix may be to change `id` to a property, but not sure if that should be a part of this PR:

```
AttributeError: Mock object has no attribute 'id'
```

PR #423, opened 2018-04-05
Add Dockerfile that builds and packages ClusterRunner

cmcginty:


boxcla:

app/util/autoversioning.py
```diff
@@ -51,7 +51,7 @@ def _calculate_source_version():
         hash_extension = '' if head_commit_is_on_trunk else '-{}'.format(head_commit_hash[:7])
         mod_extension = '' if not _repo_has_uncommited_changes() else '-mod'
         return '{}.{}{}{}'.format(_MAJOR_MINOR_VERSION, commit_count, hash_extension, mod_extension)
-    except subprocess.CalledProcessError:
+    except (subprocess.CalledProcessError, FileNotFoundError):
```
josephharrington: Haha, fun. This is a good safeguard for packaging. Is that what the assert in setup.py guards against?

Now that I'm thinking about it, I wonder at what point the clusterrunner service would fail if git was missing. Probably not on startup. 😬

Dockerfile
```diff
@@ -0,0 +1,24 @@
+#
+# Docker workflow for building and packaging ClusterRunner to an RPM.
+#
+
+# STAGE 1: Python-Manylinux (RHEL5) base with Python 3.4 enabled to create linux_x86_64 pex.
```
josephharrington: Is Manylinux a common base image? Just curious what all the requirements are for a base image if/when we ever update to a newer version of Python.

Also, are these images pretty trustworthy? Obvs we should make sure there aren't any ways to mess with our precious clusterrunner service. :gollum_emoji:

cmcginty: I'll add some comment around this. It is an official support image for building pythonw wheel.

https://www.python.org/dev/peps/pep-0513/#docker-image

app/util/autoversioning.py
cmcginty: Yeah, I hit this bug and the rpm was building with version 0.0. I made a change however to remove the assert in setup.py

requirements.txt
```diff
@@ -4,7 +4,7 @@ configobj==5.0.6
 fysom==2.1.2
 Logbook==0.7.0
 prometheus_client==0.0.19
-psutil==2.1.2
+psutil==2.2.0
```
cmcginty: v2.1.2 has a build bug in the c extension for older linux kernels.

setup.py
```diff
@@ -6,6 +6,7 @@
 from app.util import autoversioning
 
 version = autoversioning.get_version()
+assert version != '0.0'
```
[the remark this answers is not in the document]
cmcginty: I'm going to move this check to the makefile.

Dockerfile
cmcginty: The current image has CPython 2.7, 3.3, 3.4, 3.5 and 3.6

Makefile
```diff
@@ -129,7 +139,20 @@ endif
 		$(CR_CONF)=$(RPM_CR_CONF) \
 		conf/clusterrunner-master=/etc/init.d/
 
+.PHONY: docker-rpm
+docker-rpm: TAG=productivity/clusterrunner
```
ethomas2: will TAG ever be anything other than productivity/clusterrunner?

cmcginty: No, I can declare this in the target, which is probably clearer solution.

PR #425, opened 2018-04-16
WIP: Add slave registry to store connected slaves

shriganeshs:
Add slave registry to store connected slaves at master side.
This PR adds a SlaveRegistry singleton class to keep all slaves by url (and id) dictionary. It provides methods `add_slave` and `remove_slave` so that (in future) `shutdown` will delete the slave from the dictionary. This will help with autoscaling as it adds ability to delete scale-in slaves (i.e. shutdown/dead) by removing these slaves from the SlaveRegistry.


boxcla:

PR #426, opened 2018-04-17
Install clusterrunner-slave service with rpm

cmcginty:


boxcla:

conf/clusterrunner-slave
```diff
@@ -0,0 +1,99 @@
...
+    if is_running; then
+        echo "Already started"
+    else
+        echo "Starting $name"
+        cd "$dir"
+        if [ -z "$user" ]; then
+            sudo $cmd >> "$stdout_log" 2>> "$stderr_log" &
+        else
+            sudo -u "$user" $cmd >> "$stdout_log" 2>> "$stderr_log" &
+        fi
```
shriganeshs: Forgot to comment on this before.
Currently if slave hostname is not properly configured (missing fqdn), even though CR-slave connects to master, later when master tries to send subjob to slave, it fails and slave goes offline.
Currently Terraform provisioners pretty much guaranty that the hostname will be configured correctly, but that may not be the case with autoscaling.
We need to figure out how can we make sure that slave hostname is configured properly before stating slave service.

cmcginty: Maybe this helps, but the services will not start by default. That means that if you want the service to wait till there if a FQDN set, then make sure we don't do anything after installing the yum package. Once the FQDN is defined, then start the service and mark it as "enabled" on startup.

PR #427, opened 2018-04-17
Remove slave from registry when shutdown-ed

shriganeshs:
Add remove functionality to the existing /shutdown endpoint.
Add slave registry to store connected slaves at master side.
This PR adds a `SlaveRegistry` singleton class to store and maintain all slaves by url (and id) dictionary. It provides methods `add_slave` and `remove_slave`. `Shutdown` will delete the slave from this dictionary. This will help with autoscaling as it adds ability to remove slaves during auto scale-in operation (i.e. shutdown/dead) by removing these slaves from the SlaveRegistry.


boxcla:

app/master/cluster_master.py
```diff
@@ -173,11 +139,10 @@ def connect_slave(self, slave_url, num_executors, slave_session_id=None):
         # todo: Validate arg types for this and other methods called via API.
         # If a slave had previously been connected, and is now being reconnected, the cleanest way to resolve this
         # bookkeeping is for the master to forget about the previous slave instance and start with a fresh instance.
-        if slave_url in self._all_slaves_by_url:
-            old_slave = self._all_slaves_by_url.get(slave_url)
+        try:
```
ethomas2: I find this less readable and less robust than it was previously. 

Less readable because with this large try catch it's not clear what line we're expecting to throw the ItemNotFoundError.
Less robust because if a line in [111, 125] unexpectedly throws an ItemNotFoundError, we'll silently drop the unexpected error.

I realize you were sort've forced into writing it like this because `get_slave` throws an error and you just copy pasted that code, but I don't like like that `get_slave` throws an error in the first place.

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+class SlaveRegistry(Singleton):
+    """
+    SlaveRegistry class is a singleton class which stores and maintains list of connected slaves.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.all_slaves_by_url = {}
+        self.all_slaves_by_id = {}
+        self._slave_dict_lock = Lock()
```
ethomas2: I don't think this lock is necessary. If i'm not mistaken SlaveRegistry methods are only being called by ClusterMaster, which is only being called by tornado's thread. You can verify (or falsify) this claim by printing [threading.get_ident()](https://docs.python.org/3/library/threading.html#threading.get_ident) in slave registry methods. See if the identifier ever changes.

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+
+    def add_slave(self, slave):
```
ethomas2: Use python3 type annotations instead of docstrings throughout this file

`def add_slave(self, slave: Slave)`

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+        """
+        Look for a slave in the registry and if not found raise "ItemNotFoundError" exception.
+
+        :param slave_id: The id of the slave to return
+        :type slave_id: int
+        :param slave_url: The url of the slave to return
+        :type slave_url: str
+        :return: The instance of the slave
+        :rtype: Slave | None
+        """
```
ethomas2: The rtype of this function is just `Slave`, not `Slave | None`

I think it _should_ be `Slave | None` as I argued above, but currently it's not.

app/web_framework/cluster_master_application.py
```diff
@@ -28,6 +29,7 @@ def __init__(self, cluster_master):
         default_params = {
             'cluster_master': cluster_master,
         }
+        self._slave_registry = SlaveRegistry.singleton()
```
cmcginty: Where is this used?

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+        Add slave in SlaveRegistry.
+
+        :type slave: Slave
+        """
+        with self._slave_dict_lock:
+            self.all_slaves_by_url[slave.url] = slave
+            self.all_slaves_by_id[slave.id] = slave
+
+    def remove_slave(self, slave=None, slave_url=None):
```
cmcginty: Having `slave` in the name is redundant, so just call it `remove()`

app/master/slave.py
```diff
@@ -1,14 +1,17 @@
...
 from app.master.build import Build
 from app.master.subjob import Subjob
 from app.util import analytics, log
 from app.util.counter import Counter
+from app.util.exceptions import ItemNotFoundError
 from app.util.network import Network, RequestFailedError
 from app.util.secret import Secret
 from app.util.session_id import SessionId
 from app.util.url_builder import UrlBuilder
+from app.util.singleton import Singleton
```
cmcginty: alphabetize :neckbeard:

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
     def get_last_heartbeat_time(self) -> datetime:
         return self._last_heartbeat_time
 
+    def _update_slave_registry(self):
```
cmcginty: `_remove_slave_from_registry()`

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+        super().__init__()
+        self.all_slaves_by_url = {}
+        self.all_slaves_by_id = {}
+        self._slave_dict_lock = Lock()
+
+    def add_slave(self, slave):
+        """
+        Add slave in SlaveRegistry.
+
+        :type slave: Slave
```
cmcginty: annotation

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+        with self._slave_dict_lock:
+            self.all_slaves_by_url[slave.url] = slave
+            self.all_slaves_by_id[slave.id] = slave
+
+    def remove_slave(self, slave=None, slave_url=None):
+        """
+        Remove slave from the both url and id dictionary.
+
+        :type slave: Slave
+        :type slave_url: str
```
cmcginty: annotation

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+            except (ItemNotFoundError, KeyError):
+                # Ignore if slave does not exists in SlaveRegistry
+                pass
+
+    def get_slave(self, slave_id=None, slave_url=None):
+        """
+        Look for a slave in the registry and if not found raise "ItemNotFoundError" exception.
+
+        :param slave_id: The id of the slave to return
+        :type slave_id: int
```
cmcginty: annotation

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+        if slave_id is not None:
+            error_msg = 'Requested slave (slave_id={}) does not exist.'.format(slave_id)
+        else:
+            error_msg = 'Requested slave (slave_url={}) does not exist.'.format(slave_url)
+        raise ItemNotFoundError(error_msg)
+
+    def get_all_slaves_by_id(self):
+        return self.all_slaves_by_id
+
+    def get_all_slaves_by_url(self):
```
cmcginty: same as above

app/master/slave.py
cmcginty: Make sure to document this assumption if you remove the lock.

app/master/slave.py
```diff
@@ -244,6 +250,87 @@ def update_last_heartbeat_time(self):
...
+
+
+class SlaveRegistry(Singleton):
+    """
+    SlaveRegistry class is a singleton class which stores and maintains list of connected slaves.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.all_slaves_by_url = {}
```
cmcginty: add `# type:` annotations

app/web_framework/cluster_master_application.py
```diff
@@ -417,7 +420,7 @@ class _SlavesShutdownHandler(_ClusterMasterBaseAPIHandler):
     @authenticated
     def post(self):
         shutdown_all = self.decoded_body.get('shutdown_all')
-        slaves_to_shutdown = self._cluster_master.all_slaves_by_id().keys() if shutdown_all else\
+        slaves_to_shutdown = SlaveRegistry.singleton().get_all_slaves_by_id().keys() if shutdown_all else\
```
cmcginty: Use `( )` instead of `\`. But in this case I would split into a regular `if / else` clause.

app/master/cluster_master.py
shriganeshs: Done.

app/master/slave.py
shriganeshs: I do not agree. `remove_slave` is called from `Slave.mark_slave_idle`.

app/master/slave.py
shriganeshs: Done.

app/master/slave.py
shriganeshs: What if its either Slave or ItemNotFound Exception? I am doing `Union[Slave, ItemNotFoundError]`. Is that correct?

app/web_framework/cluster_master_application.py
shriganeshs: I was using it before. Now removed.

app/master/slave.py
shriganeshs: I understand your point, but I think, as its a SlaveRegistry, `remove_slave` is more readable.

app/master/slave.py
shriganeshs: Done.

app/master/slave.py
shriganeshs: Done.

app/master/slave.py
cmcginty: There is a lot of code that is now just `slave_reg.remove_slave`. All of the items in the slave registry are slaves which is that's why I think its redundant to name it that way. If you follow this convention you should have `*_slave` virtually in every method name.

app/master/slave.py
cmcginty: Add type annotations. Also if you are not going address getter changes then make this attribute private.

shriganeshs: Is there any example of this?

shriganeshs: `self.all_slaves_by_url = types.DictType.__new__(types.DictType, (), {})` or
`self.all_slaves_by_url = dict()`?

cmcginty: https://www.python.org/dev/peps/pep-0484/

Something like:

```
self.all_slaves_by_url = {}  # type: Dict[String,ClusterRunnerSlave]
```

test/unit/master/test_cluster_master.py
```diff
@@ -47,6 +48,8 @@ def setUp(self):
         self._thread_pool_executor_cls.return_value.submit.side_effect = \
             self.thread_pool_executor.submit
 
+        SlaveRegistry.reset_singleton()
```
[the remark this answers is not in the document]
shriganeshs: Done in https://github.com/box/ClusterRunner/pull/443

PR #428, opened 2018-04-23
Add Dockerized release target for publishing RPM to Artifactory

cmcginty:
- Add Jenkinsfile with build/release job
- Remove sha256 hash from Dockerfile for portability
- Fix "first time" eval errors in docker-rpm target
- Move common defines to top of Makefile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # STAGE 1: Official PEP 513 Python Manylinux (RHEL5) base with Python 3.4 enabled to create
 #          linux_x86_64 pex.
 #          NOTE: To update image run: "docker pull quay.io/pypa/manylinux1_x86_64:latest"
-FROM quay.io/pypa/manylinux1_x86_64@sha256:1fc8ee3bc9d668222a519d84d8c3b3e2bbc3eabe4595a1661d7b037a038d2e87 AS stage1
+FROM quay.io/pypa/manylinux1_x86_64:latest AS stage1
 ENV PATH="/opt/python/cp34-cp34m/bin:${PATH}"
 
 WORKDIR /ClusterRunner
@@ -18,7 +18,7 @@ RUN make dist/clusterrunner
 
 # STAGE 2: CentOS 7 base w/ fpm to package pex into an rpm.
 #          NOTE: To update image run: "docker pull cdrx/fpm-centos:7"
-FROM cdrx/fpm-centos@sha256:fd314b5b8fdc78f714ba6d3e83d45ae58dd8dfff7f21320e55ec854587b01c2f AS stage2
+FROM cdrx/fpm-centos:7 AS stage2
 WORKDIR /root
 COPY . .
 COPY --from=stage1 /ClusterRunner/dist/clusterrunner ./dist/
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,6 @@
 
 # STAGE 1: Official PEP 513 Python Manylinux (RHEL5) base with Python 3.4 enabled to create
 #          linux_x86_64 pex.
-#          NOTE: To update image run: "docker pull quay.io/pypa/manylinux1_x86_64:latest"
 FROM quay.io/pypa/manylinux1_x86_64:latest AS stage1
 ENV PATH="/opt/python/cp34-cp34m/bin:${PATH}"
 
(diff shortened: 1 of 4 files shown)


boxcla:

Dockerfile
```diff
@@ -5,7 +5,7 @@
 # STAGE 1: Official PEP 513 Python Manylinux (RHEL5) base with Python 3.4 enabled to create
 #          linux_x86_64 pex.
 #          NOTE: To update image run: "docker pull quay.io/pypa/manylinux1_x86_64:latest"
-FROM quay.io/pypa/manylinux1_x86_64@sha256:1fc8ee3bc9d668222a519d84d8c3b3e2bbc3eabe4595a1661d7b037a038d2e87 AS stage1
+FROM quay.io/pypa/manylinux1_x86_64:latest AS stage1
```
ethomas2: Why are you using latest instead of pinning it to a version?

Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN make dist/clusterrunner
 
 # STAGE 2: CentOS 7 base w/ fpm to package pex into an rpm.
 #          NOTE: To update image run: "docker pull cdrx/fpm-centos:7"
-FROM cdrx/fpm-centos@sha256:fd314b5b8fdc78f714ba6d3e83d45ae58dd8dfff7f21320e55ec854587b01c2f AS stage2
+FROM cdrx/fpm-centos:7 AS stage2
```
ethomas2: Is this pinned to a version?

Makefile
```diff
@@ -147,17 +177,38 @@ rpm: $(CR_BIN) $(PKG_INFO)
 .PHONY: docker-rpm
 docker-rpm:
 	$(call print_msg, Running ClusterRunner Docker RPM builder... )
-	$(eval TAG := productivity/clusterrunner)
+	docker build -t $(DOCKER_TAG) -f Dockerfile .
```
ethomas2: Is the -f necessary if this is the only Dockerfile in the repo?

Makefile
```diff
@@ -147,17 +177,38 @@ rpm: $(CR_BIN) $(PKG_INFO)
...
-	@# Docker "run" must be called before the next steps.
-	$(eval CONTAINER_ID := $(shell docker ps -alq))
-	docker cp $(CONTAINER_ID):$(RPM_PATH) dist/
+	@# Docker cp does not support globing, so the path to the RPM file must be
+	@# detected with a query. The order of the commands are important and they
+	@# must all be run in the same "shell" for the variables to be available to
+	@# the final "docker cp" command.
+	DOCKER_RPM_PATH=$$(docker run $(DOCKER_TAG) sh -c "ls /root/$(DIST_DIR)/*.rpm") && \
+	CONTAINER_ID=$$(docker ps -alq) && \
+	docker cp $$CONTAINER_ID:$$DOCKER_RPM_PATH $(DIST_DIR)
```
ethomas2: `docker rm` the container after you grab the rpm out of it

Makefile
```diff
@@ -147,17 +177,38 @@ rpm: $(CR_BIN) $(PKG_INFO)
...
+.PHONY: release
+release: $(PY_PKG_INFO) $$($$RPM_PATH)
+	curl -u $(ARTIFACTORY) -X PUT $(ARTIFACTORY_URL)/$(RPM_VERSION)-$(RPM_RELEASE)/$(RPM_FILE) -T $(RPM_PATH)
+
+# Run the "release" target in the Docker container. Technically this is not
+# required, but guarantees that a release can be made without a working Python
+# environment.
+.PHONY: docker-release
+docker-release: docker-rpm
+	docker run -e ARTIFACTORY=$(ARTIFACTORY) $(DOCKER_TAG) /usr/bin/make release
```
ethomas2: Add a `--rm` option to `docker run`

Dockerfile
cmcginty: No, Pinning doesn't work unfortunately because the remotes don't keep old versions. I guess to implement pinning we need to have an docker image cache.

I guess I can remove the comment above this change.

Dockerfile
cmcginty: See my other comment.

Makefile
cmcginty: No its not, but I like being extra explicit in Makefiles since reader might not be familiar with Docker

Makefile
cmcginty: 👍

Makefile
cmcginty: 👍

PR #431, opened 2018-04-26
Support 'None' values in configfile

mayurva:
If the value for any configuration is set to the string 'None', the
config loader will set it to noneType


cmcginty:
Is this to get around enable/disable of certfile?

mayurva:
Closing this as we can set the config to empty string and it will be ignored.

PR #432, opened 2018-04-26
Kill slave process on non 200 heartbeat response

mayurva:
If heartbeat response is not HTTP 200, it means that the master is up,
but does not recognize the slave. In such cases the slave should
immediately try to reconnect to master. This event should also be
considered as heartbeat failure and increment the heartbeat failure
count. If the slave gets more failures than the threshold, the process
dies

### DIFF
--- a/app/slave/cluster_slave.py
+++ b/app/slave/cluster_slave.py
@@ -74,24 +74,27 @@ class ClusterSlave(ClusterService):
 
     def _run_heartbeat(self):
         try:
-            self._send_heartbeat_to_master()
-            self._heartbeat_failure_count = 0
+            response = self._send_heartbeat_to_master()
+
+            # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
+            if not response.ok:
+                self._logger.error('Master did not respond with HTTP 200')
+                self.kill()
+            else:
+                self._heartbeat_failure_count = 0
         except (requests.ConnectionError, requests.Timeout):
             self._heartbeat_failure_count += 1
             if self._heartbeat_failure_count >= self._heartbeat_failure_threshold:
-                self._logger.error('Master is not responding to heartbeats')
-
-                # TODO: Right now the slave simply dies when it does not hear back from master. The next step would
-                # be to try to reconnect to master at this point. In future the heartbeat and connect_to_master
-                # methods can combined into one. This combined method will behave differently based on current state.
+                self._logger.error('Master has not responded to last {} heartbeats.'.format(
+                    self._heartbeat_failure_threshold))
                 self.kill()
 
         self._hb_scheduler.enter(self._heartbeat_interval, 0, self._run_heartbeat)
 
     def _send_heartbeat_to_master(self):
         heartbeat_url = self._master_api.url('slave', self._slave_id, 'heartbeat')
-        self._network.post_with_digest(heartbeat_url, request_params={'slave': {'heartbeat': True}},
-                                       secret=Secret.get())
+        return self._network.post_with_digest(heartbeat_url, request_params={'slave': {'heartbeat': True}},
+                                              secret=Secret.get())
 
     def api_representation(self):
         """
--- a/app/slave/cluster_slave.py
+++ b/app/slave/cluster_slave.py
@@ -78,15 +78,19 @@ class ClusterSlave(ClusterService):
 
             # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
             if not response.ok:
-                self._logger.error('Master did not respond with HTTP 200')
+                self._logger.error('Received HTTP {} from master.'.format(response.status_code))
+                self._logger.error('The slave process is shutting down.')
                 self.kill()
             else:
                 self._heartbeat_failure_count = 0
         except (requests.ConnectionError, requests.Timeout):
             self._heartbeat_failure_count += 1
+            self._logger.warning('Master has not responded to last {} heartbeats.'.format(
+                self._heartbeat_failure_count))
             if self._heartbeat_failure_count >= self._heartbeat_failure_threshold:
-                self._logger.error('Master has not responded to last {} heartbeats.'.format(
+                self._logger.error('Heartbeat failure threshold of {} reached.'.format(
                     self._heartbeat_failure_threshold))
+                self._logger.error('The slave process is shutting down.')
                 self.kill()
 
         self._hb_scheduler.enter(self._heartbeat_interval, 0, self._run_heartbeat)
--- a/test/unit/slave/test_cluster_slave.py
+++ b/test/unit/slave/test_cluster_slave.py
@@ -280,6 +280,7 @@ class TestClusterSlave(BaseUnitTestCase):
 
         mock_response = MagicMock(spec=requests.models.Response, create=True)
         mock_response.ok = False
+        mock_response.status_code = http.client.NOT_FOUND
         self.mock_network.post_with_digest.return_value = mock_response
 
         slave._run_heartbeat()
--- a/test/unit/slave/test_cluster_slave.py
+++ b/test/unit/slave/test_cluster_slave.py
@@ -150,6 +150,7 @@ class TestClusterSlave(BaseUnitTestCase):
         def _get_success_mock_response():
             mock_response = MagicMock(spec=requests.models.Response, create=True)
             mock_response.status_code = http.client.OK
+            mock_response.ok = True
             return mock_response
 
         def fake_network_post(url, *args, **kwargs):
@@ -256,6 +257,7 @@ class TestClusterSlave(BaseUnitTestCase):
 
         slave = self._create_cluster_slave()
         slave.connect_to_master(self._FAKE_MASTER_URL)
+
         if not is_master_responsive:
             self.mock_network.post_with_digest.side_effect = requests.ConnectionError
 
@@ -271,6 +273,21 @@ class TestClusterSlave(BaseUnitTestCase):
             else:
                 self.assertEqual(self._mock_sys.exit.call_count, 0,
                                  'slave keeps running when heartbeat failure threshold is not reached')
+
+    def test_slave_dies_on_http_404_from_master(self):
+        slave = self._create_cluster_slave()
+        slave.connect_to_master(self._FAKE_MASTER_URL)
+
+        mock_response = MagicMock(spec=requests.models.Response, create=True)
+        mock_response.ok = False
+        self.mock_network.post_with_digest.return_value = mock_response
+
+        slave._run_heartbeat()
+
+        self.mock_network.post_with_digest.assert_called_once_with(
+                        ANY,request_params={'slave': {'heartbeat': True}}, secret=ANY)
+        self.assertEqual(self._mock_sys.exit.call_count, 1,
+                         'slave dies when it receives HTTP 404 from master')
 
     def _create_cluster_slave(self, **kwargs):
         """


boxcla:

app/slave/cluster_slave.py
```diff
@@ -74,24 +75,36 @@ def _start_heartbeat(self):
...
-            self._send_heartbeat_to_master()
-            self._heartbeat_failure_count = 0
+            response = self._send_heartbeat_to_master()
+
+            # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
+            # The slave tries to reconnect to master at this point.
+            # This is also counted as a heartbeat failure and the slave takes an action accordingly
+            if response.status_code != http.client.OK:
+                self._act_on_heartbeat_failure()
+                self._logger.info('Master responded with {}. Attempting to reconnect.'.format(response.status_code))
```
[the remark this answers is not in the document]
mayurva: This is on purpose. The slave may decide to die inside `_act_on_heartbeat_failure` because the heartbeat threshold is reached. Say, a case where master is returning 500s. We only want to reconnect only if that is not the case.

app/slave/cluster_slave.py
```diff
@@ -74,24 +75,36 @@ def _start_heartbeat(self):
...
         try:
-            self._send_heartbeat_to_master()
-            self._heartbeat_failure_count = 0
+            response = self._send_heartbeat_to_master()
+
+            # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
+            # The slave tries to reconnect to master at this point.
+            # This is also counted as a heartbeat failure and the slave takes an action accordingly
+            if response.status_code != http.client.OK:
+                self._act_on_heartbeat_failure()
```
ethomas2: This method name is vague. If this method is doing so many things that "act" is the best descriptor, that's probably an indication that it's doing too many things.

app/slave/cluster_slave.py
```diff
@@ -74,24 +75,36 @@ def _start_heartbeat(self):
...
     def _run_heartbeat(self):
         try:
-            self._send_heartbeat_to_master()
-            self._heartbeat_failure_count = 0
+            response = self._send_heartbeat_to_master()
+
+            # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
+            # The slave tries to reconnect to master at this point.
+            # This is also counted as a heartbeat failure and the slave takes an action accordingly
+            if response.status_code != http.client.OK:
```
ethomas2: `!=200` might be a little strict
`if response.status_code > 299`

app/slave/cluster_slave.py
```diff
@@ -74,24 +75,36 @@ def _start_heartbeat(self):
...
 
+    def _act_on_heartbeat_failure(self):
+        self._heartbeat_failure_count += 1
+        if self._heartbeat_failure_count >= self._heartbeat_failure_threshold:
+            self._logger.error('Master is not responding to heartbeats')
+
+            # TODO: Right now the slave simply dies when it does not hear back from master. The next step would
+            # be to try to reconnect to master at this point. In future the heartbeat and connect_to_master
+            # methods can combined into one. This combined method will behave differently based on current state.
+            self.kill()
```
ethomas2: If this change is connecting slaves to the master, then I think it's time to get rid of this `self.kill()`

as per this comment
https://github.com/box/ClusterRunner/pull/422#discussion_r180558189

ethomas2: Which further implies that it's time to get rid of this method completely

app/slave/cluster_slave.py
```diff
@@ -74,24 +75,36 @@ def _start_heartbeat(self):
...
-            self._heartbeat_failure_count = 0
+            response = self._send_heartbeat_to_master()
+
+            # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
+            # The slave tries to reconnect to master at this point.
+            # This is also counted as a heartbeat failure and the slave takes an action accordingly
+            if response.status_code != http.client.OK:
+                self._act_on_heartbeat_failure()
+                self._logger.info('Master responded with {}. Attempting to reconnect.'.format(response.status_code))
+                self.connect_to_master(self._master_url)
```
ethomas2: 1. This connect_to_master method adds 2 callbacks that will trigger when someone kills the slave. If this gets called a lot of times I think multiple instances of those callbacks will get collected and called multiple times.

2. I don't think we should be calling connect_to_master at all here. I talked with @cmcginty offline about this and it seems like there might've been some confusion. My understanding was always that there would be no difference between sending a heartbeat and sending a request to connect to the master. The slave would periodically send a heartbeat and if the master didn't know about that slave it would connect it, if the master did know about that slave then it would be a noop. So you never have to explicitly connect to the master, the heartbeat request takes care of that for you

app/slave/cluster_slave.py
mayurva: yeah.. It's a bad name. I'll think about a better name or possibly splitting it into multiple methods.

ethomas2: I don't mean to imply that you can _always_ find a better name btw. Sometimes you you just have to make one method do 5 things. I don't think that's the case here though

app/slave/cluster_slave.py
josephharrington: `response.ok`?

ethomas2:
I think I like the slave-suicide idea. Very clean. Can you confirm that all of the following assumptions i'm making are true?

1. When the slave dies or suicides gracefully it sends a request to the master telling the master that it died, and the master clears whatever state it had for the slave
2. When the slave dies ungracefully (ie it gets kill -9'd or the vm dies) the master will clear it's state about the slave after it misses enough heartbeat intervals from the slave
3. When a slave restarts it gets a new slave id from the master and a new slave object is allocated in the master's memory
4. When the master sends a subjob to a slave, it sends the slave id in the request to the slave. If the slave id in the request does not match what the slave thinks its id is, the slave kills itself.

If we go with this that means we're building in a core assumption into cluster runner that something external to cluster runner will be there to restart dead slaves. Are we ok with building that assumption into an open source product? I'm not even sure we can make that assumption at box right now. We can't use systemd to restart dead slaves because we have too many sl6 slave nodes. Eventually i'd like to run cluster runner on kubernetes, but that's a long way off. @josephharrington do you have an idea for what would be responsible for restarting slave nodes? 

If we can figure out what's responsible for the restarting, and those 4 assumptions I made are correct, then I think slave suicide is a really clean idea.

josephharrington:
@ethomas2 

> When the slave dies or suicides gracefully it sends a request to the master telling the master that it died, and the master clears whatever state it had for the slave

I think right now if the slave "reconnects" (connects with the same url as some previous slave) then it will overwrite any existing state on the master side. The master assumes a slave that just connected is idle and ready to be assigned to a build.

> When the slave dies ungracefully (ie it gets kill -9'd or the vm dies) the master will clear it's state about the slave after it misses enough heartbeat intervals from the slave

@mayurva can answer this.

> When a slave restarts it gets a new slave id from the master and a new slave object is allocated in the master's memory

This happens when a slave connects, yes.

> When the master sends a subjob to a slave, it sends the slave id in the request to the slave. If the slave id in the request does not match what the slave thinks its id is, the slave kills itself.

It's not the slave id, it's the session id in the header of requests. If the master sends a request to a slave with an unexpected session id, the slave rejects the request with a 412. Looking at the code, for some calls from the master this might fail silently since return code is not checked. 😞

--

As far as what would restart the slave service if it died, I imagined it would restart itself (via a Popen and child process detach). If that's too tricky then maybe an alternative would be adding a loop in the `SlaveSubcommand` level where we could reinstantiate the service objects and restart the web server.

This is all just spitballing. My main point was that we need *some* way to make sure the slave service state is reset when it reconnects to the master, however that happens.

mayurva:
@ethomas2 and @josephharrington I have addressed most of the comments. I'll try to answer some of your questions here:

The idea of this pull request is that if the slave gets a 404 response from master, there is no need to send any more heartbeats since the master is responding, but does not identify the slave. Sounds like attempting to reconnect to master was a bit of a premature optimization. It's probably better to let the slave die at this point. As per the changes from #426 slave process is now a daemon configured through `init.d`, so slave process will restart will come up with a clean state. The master will connect the slave as a new slave, and will have a clean state too.

mayurva:
@ethomas2 answering some of your questions inline here:

> The real reason I don't want the slave to restart itself is just because it "feels weird".

if I understand this correctly, you mean instead of slave process deciding to restart itself, it should be `initd` or `systemd` that restarts it. This is what will happen now. The slave decides to kill itself when it is in an unexpected state, and `init.d` will restart it.

> When the slave dies ungracefully (ie it gets kill -9'd or the vm dies) the master will clear it's state about the slave after it misses enough heartbeat intervals from the slave

Right now the only action the master takes is set `is_alive=False` for the slave, which means master will not send any more subjobs to the slave. If the slave restarts and reconnects, master will treat it as a brand new slave, which I think should be fine operationally.

josephharrington:
@ethomas2 

> I originally thought to add the slave id in the request to avoid that. I just now realized that the build id will already be in the request. Will the slave reject the subjob in step 3 because it doesn't recognize the build id? As long as the slave rejects the subjob (and the master doesn't fail the build) i'm happy.

The session id is designed to handle this scenario. A new session id is generated when the slave starts up and sent to the master when the slave connects. The master will include that session id when it sends requests to the slave. If the session id doesn't match what the slave expects, it will reject the request.

> It would have strange interactions with init.d/systemd/other watchdog processes.

That's a good point. Any thoughts on the other idea of altering `SlaveSubcommand` so that we can reinstantiate and do a "soft reset" without actually restarting the process? Important to note that this is to handle reconnecting and has a different goal than managing process lifecycle like systemd would. I think the alternative is to manually reset all the state from within the existing `ClusterSlave` instance which feels error-prone to me. 

> @cmcginty just made some rpm that writes the pid of the slave process to a file

Interesting, I missed that one. The master and slave services already [write their pid](https://github.com/box/ClusterRunner/blob/master/app/subcommands/slave_subcommand.py#L48) to a file on startup.

josephharrington:
@mayurva I'm also fine with avoiding trying to do reconnects for now if the plan is to let the system handle the process. I think that's cleaner. 

Re: @ethomas2's comment:
> If we go with this that means we're building in a core assumption into cluster runner that something external to cluster runner will be there to restart dead slaves. Are we ok with building that assumption into an open source product?

I think that's fine. It would be a more advanced setup step, but optional. Setting up something external to restart slave services could be an optional thing that service owners could do to increase robustness of their cluster.

ethomas2:
@mayurva 
> As per the changes from #426 slave process is now a daemon configured through init.d, so slave process will restart will come up with a clean state.
It's a daemon, but I don't think init.d is configured to restart the slave process when it dies. That's easy to do with systemd, but i'm not sure about init.d (maybe you're more familiar with init.d than I am). I was talking with @cmcginty about this he said worst case scenario we can modify the init.d script ourselves to monitor the process, but I don't _think_ that functionality exists now.

@josephharrington 
> Any thoughts on the other idea of altering SlaveSubcommand so that we can reinstantiate and do a "soft reset" without actually restarting the process?
I like that better than Popen but less than relying on init.d. After talking with @cmcginty it seems that we _can_ leverage init.d, even if we're not right now.

app/slave/cluster_slave.py
```diff
@@ -74,24 +74,27 @@ def _start_heartbeat(self):
 
     def _run_heartbeat(self):
         try:
-            self._send_heartbeat_to_master()
-            self._heartbeat_failure_count = 0
+            response = self._send_heartbeat_to_master()
+
+            # A non HTTP 200 response indicates that the master is up, but does not recognize the slave.
+            if not response.ok:
+                self._logger.error('Master did not respond with HTTP 200')
```
josephharrington: Do you want to log what the actual response code was here? Might be useful for debugging.

PR #435, opened 2018-04-27
Fix autoversion bugs in release steps

cmcginty:
- Correctly catch except from "import pkg_resources"
- Do not filter source from Docker to prevent false-positive "dirty"
  check


boxcla:

PR #436, opened 2018-04-27
Update "git diff" test to be compatible with "docker build"

cmcginty:


boxcla:

PR #437, opened 2018-04-27
Fix autoversioning lint and test cases

cmcginty:


boxcla:

PR #439, opened 2018-05-01
Modify Jenkinsfile to perform a signed rpm release

cmcginty:
- Extract the PKG-INFO file from docker image
- Add release-signed make target for running Box release script

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,17 +1,25 @@
 pipeline {
-    agent { label 'rosie' }
+    agent none
 
     stages {
         stage('Build') {
+            agent { label 'rosie' }
             steps {
-                sh 'make docker-rpm'
+                sh 'make clean docker-rpm'
+
+                // Copy the RPM and PKG-INFO file to the next stage.
+                stash name: 'rpm', includes: 'dist/*.rpm'
+                stash name: 'pkg-info', includes: 'clusterrunner.egg-info/PKG-INFO'
             }
         }
         stage('Release') {
+            agent { label 'pe-builder' }
             steps {
-                withCredentials([usernameColonPassword(credentialsId: 'artifactory', variable: 'ARTIFACTORY')]) {
-                    sh 'make docker-release'
-                }
+                // Write the RPM and PKG-INFO file from the previous stage.
+                unstash 'rpm'
+                unstash 'pkg-info'
+
+                sh 'make release-signed'
             }
         }
     }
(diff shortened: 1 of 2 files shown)


boxcla:

Jenkinsfile
```diff
@@ -1,17 +1,21 @@
 pipeline {
-    agent { label 'rosie' }
+    agent none
```
ethomas2: I see agent is being overridden below, but can you put a default agent here anyway? So someone doesn't add a new stage and accidentally run on some random node

Jenkinsfile
```diff
@@ -1,17 +1,21 @@
...
 
     stages {
         stage('Build') {
+            agent { label 'rosie' }
             steps {
-                sh 'make docker-rpm'
+                sh 'make clean docker-rpm'
+                stash name: 'rpm', includes: 'dist/*.rpm'
```
ethomas2: What do stash and unstash do? If they're relatively uncommon commands then add a comment please. If they're common and I'm just not familiar with jenkinsfiles then just leave it alone

Makefile
```diff
@@ -175,18 +175,22 @@ rpm: $(CR_BIN) $(PY_PKG_INFO)
...
-	mkdir -p $(DIST_DIR)
+	mkdir -p $(DIST_DIR) $(PY_EGG_INFO_DIR)
 	@# Docker cp does not support globing, so the path to the RPM file must be
 	@# detected with a query. The order of the commands are important and they
 	@# must all be run in the same "shell" for the variables to be available to
 	@# the final "docker cp" command.
 	DOCKER_RPM_PATH=$$(docker run $(DOCKER_TAG) sh -c "ls /root/$(DIST_DIR)/*.rpm") && \
 	CONTAINER_ID=$$(docker ps -alq) && \
 	docker cp $$CONTAINER_ID:$$DOCKER_RPM_PATH $(DIST_DIR) && \
+	docker cp $$CONTAINER_ID:/root/$(PY_PKG_INFO) $(PY_EGG_INFO_DIR) && \
```
ethomas2: Are we copying python eggs out of the container? If so why?

Makefile
```diff
@@ -208,6 +212,16 @@ release: $(PY_PKG_INFO) $$($$RPM_PATH)
 docker-release: docker-rpm
 	docker run --rm -e ARTIFACTORY=$(ARTIFACTORY) $(DOCKER_TAG) /usr/bin/make release
 
+# See "release" target. The major difference is that this target uses the Box custom
+# rpm-to-artifactory tooling that is made available for signing and uploading rpms.
+.PHONY: release-signed
+release-signed: $(PY_PKG_INFO) $$($$RPM_PATH)
```
ethomas2: Can we get rid of release and docker-release now?

cmcginty: Also it is more useful for users outside of Box.

PR #440, opened 2018-05-03
Fix bug with slave reconnection

shriganeshs:
When slave is reconnected the old-slave entry should be removed
from slave registry, before adding new entry for the same slave.


boxcla:

PR #441, opened 2018-05-04
Optimize heartbeat scheduling loop

mayurva:
sched.enter needs to be called only once for scheduling the heartbeat loop

### DIFF
--- a/app/master/cluster_master.py
+++ b/app/master/cluster_master.py
@@ -70,7 +70,7 @@ class ClusterMaster(ClusterService):
         Thread(target=self._start_heartbeat_tracker, name='HeartbeatTrackerThread', daemon=True).start()
 
     def _start_heartbeat_tracker(self):
-        self._hb_scheduler.enter(0, 0, self._disconnect_non_heartbeating_slaves)
+        self._disconnect_non_heartbeating_slaves()
         self._hb_scheduler.run()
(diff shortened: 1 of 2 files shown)


boxcla:

app/master/cluster_master.py
```diff
@@ -70,12 +70,13 @@ def start_heartbeat_tracker_thread(self):
         Thread(target=self._start_heartbeat_tracker, name='HeartbeatTrackerThread', daemon=True).start()
 
     def _start_heartbeat_tracker(self):
-        self._hb_scheduler.enter(0, 0, self._disconnect_non_heartbeating_slaves)
+        self._hb_scheduler.enter(self._unresponsive_slaves_cleanup_interval, 0,
```
cmcginty: Just call `_disconnect_non_heartbeating_slaves()` to init the heartbeat tests

mayurva: Oh I see. Makes sense, but we want to wait for `_unresponsive_slaves_cleanup_interval` seconds after the master starts up. That way heartbeat doesn't run before the slaves get a chance to connect. Is there another way to achieve that?

cmcginty: Why wait? I don't think its worth the extra complexity in the code.

PR #453, opened 2021-12-01
Error handling

DavidSchnepper:


app/util/fs.py
```diff
@@ -121,6 +124,11 @@ def tar_directories(target_dirs_to_archive_paths, tarfile_path):
         for dir_path, archive_name in target_dirs_to_archive_paths.items():
             target_dir = os.path.normpath(dir_path)
             tar.add(target_dir, arcname=archive_name)
+    # Verify that the tarfile is readable
+    with tarfile.open(tarfile_path) as tar:
+        for member in tar.getmembers():
+            with tar.extractfile(member.name) as target:
```
rsennewald, line 130: I'm not very familiar with why, but this fails when this function is called with an exception like the following: 
`line 130, in tar_directories
    with tar.extractfile(member.name) as target:
AttributeError: __exit__`
I can also confirm that files on the filesystem are correct tarfiles in the case of where we failed to receive a valid tar on the master, so I don't think this will be helpful and we can remove it.

rsennewald (approved):

